PR #102, opened 2016-05-31
Support deleting a project

frsyuki:
* Add projects.deleted_at and projects.deleted_name columns.

* Deleting a project actually doesn't delete the project from the DB so
  that past sessions and attempts can still be remained.

* Deleted project is accessible by id but not by name or index. This is
  applicable both for internal API and REST API.

* Deleting a project move projects.name to projects.deleted_name so that
  a new project can reuse the same name.

* Starting a new session of a deleted project is rejected.

* Added a new "delete" command.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseProjectStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseProjectStoreManager.java
@@ -193,17 +193,17 @@ public class DatabaseProjectStoreManager
         }
 
         @Override
-        public <T> T obsoleteProject(int projId, ProjectObsoleteAction<T> func)
+        public <T> T deleteProject(int projId, ProjectObsoleteAction<T> func)
             throws ResourceNotFoundException
         {
             return transaction((handle, dao, ts) -> {
                 StoredProject proj = requiredResource(
-                        dao.getProjectByIdWithLock(siteId, projId),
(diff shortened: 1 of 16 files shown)


digdag-core/src/main/java/io/digdag/core/database/DatabaseProjectStoreManager.java
```diff
@@ -693,11 +735,18 @@ public StoredProjectMapper(ConfigMapper cfm)
         public StoredProject map(int index, ResultSet r, StatementContext ctx)
                 throws SQLException
         {
+            String name = r.getString("name");
+            Optional<Instant> deletedAt = Optional.absent();
+            if (r.wasNull()) {
```
danielnorberg: In my model I would use `name` to show "the input users gave when the user starts the session.". I.e., the `name` is permanent and `key` is ephemeral. `name` is the project name that the user specified. `key` is assigned the value of `name` when the project is pushed.

To use a filesystem analogue, `key` is like a directory entry pointing to an inode. When the directory entry is deleted/removed/unlinked, the `key` is removed/released and can be reused to refer to another inode (project).

It feels like we're a bit into bike shedding territory here though. Both of our approaches are mostly isomorphic and I feel it's more important to merge and release this feature than to achieve perfect consensus.

danielnorberg:
👍

PR #108, opened 2016-06-07
Experimental plugin mechanism

frsyuki:
How it works:

* "system plugins" are loaded when digdag starts.
  * system plugins are configured by digdag's config file.
* "dynamic plugins" are loaded when a task of a workflow runs.
  * dynamic plugins are configured in workflow definition.
* `plugin.enabled = false` disables dynamic plugins.
* A plugin provides operator factories.

Dynamic plugins are configured in workflow definition as following:

```
_export:
  plugin:
    dependencies:
    - org.eclipse.aether:aether-transport-http:1.1.0
```

("org.eclipse.aether:aether-transport-http:1.1.0" is just an example because there're no released plugins yet)


frsyuki:
@danielnorberg PTAL

digdag-core/src/main/java/io/digdag/core/plugin/RemotePluginLoader.java
```diff
@@ -0,0 +1,190 @@
...
+            URL url;
+            try {
+                url = artifactResult.getArtifact().getFile().toPath().toUri().toURL();
+            }
+            catch (MalformedURLException ex) {
+                throw Throwables.propagate(ex);
+            }
+            urls.add(url);
+        }
+        return new PluginClassLoader(urls.build(), RemotePluginLoader.class.getClassLoader(),
```
danielnorberg: We should load plugins using a class loader hierarchy that does not include the classpath of digdag itself in order to prevent problems with dependency conflicts etc. E.g. use the bootstrap classloader as parent instead of the system classloader. If necessary, load plugin interface classes from the system classloader.

danielnorberg: Here's an example of a plugin mechanism doing this: https://github.com/spotify/helios/blob/master/helios-service-registration/src/main/java/com/spotify/helios/serviceregistration/ServiceRegistrarLoader.java#L62

Note though that it uses the extension classloader which is slated for deprecation so the bootstrap classloader should be used instead.

frsyuki: `io.digdag.core.plugin.PluginClassLoader` works as following:

* If a class is in shared packages (`io.digdag.spi`), lookup from the parent (system classloader) first, then lookup from the child (classpath of the plugin).
* otherwise, opposite order: lookup from the child (classpath of the plugin), then lookup from the parent (system classloader)

(this idea is taken from Presto's [PluginClassLoader](https://github.com/prestodb/presto/blob/c7bf63a2138c2e61f0edb307cefbf6b35fba25c2/presto-main/src/main/java/com/facebook/presto/server/PluginManager.java#L295-L299), by the way).

Dependency conflicts don't happen with this way because all classes excepting io.digdag.spi are loaded first from the plugin's classpath.

Difference is that if a plugin includes missing classes (which is probably not a big deal), my idea tires to load them from the parent classloader, and your idea throws NoClassDefFoundError immediately.

danielnorberg: Sounds good.

I would prefer to be strict and throwing `NoClassDefFoundError` instead of loading non `io.digdag.spi` classes using parent/system classloader. I think it _is_ a big deal. The reason being that a plugin might otherwise unknowingly gain an indirect dependency on a class in the digdag parent/system classpath. That dependency might then break (best case with a `NoClassDefFoundError`, worst case with silently broken behavior) at some point in the future due to internal changes in digdag even though the plugin itself has not changed.

PR #116, opened 2016-06-14
td_wait & td_wait_table operators

danielnorberg:


digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
```diff
@@ -0,0 +1,132 @@
...
+        public TaskResult runTask()
+        {
+            Config params = request.getConfig().mergeDefault(
+                    request.getConfig().getNestedOrGetEmpty("td"));
+
+            int pollInterval = getPollInterval(params);
+
+            String query = templateEngine.templateCommand(workspacePath, params, "query", UTF_8);
+
+            int defaultRows = query.trim().isEmpty() ? 0 : 1;
```
frsyuki: when can `query.trim()` be empty?

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
```diff
@@ -0,0 +1,132 @@
...
+                    .setPriority(priority)
+                    .setScheduledTime(request.getSessionTime().getEpochSecond())
+                    .createTDJobRequest();
+
+            TDJobOperator job = op.submitNewJob(req);
+            logger.info("Started {} job id={}:\n{}", engine, job.getJobId(), query);
+
+            joinJob(job);
+
+            // TODO: wrap query in a query using count and limit instead of storing and downloading the actual full query result?
```
frsyuki: I think there are 2 ideas.

(A) td_wait operator always checks that the query returns at least 1 row and the first column of the row is not false or NULL. And users use td_wait as following:
* if you want to check existence of some rows (at least 1 row), use `select true from tbl where [CONDITIONS] limit 1`.
* if you want to check number of rows (at least N rows), use `select count(*) > 100 from tbl where [CONDITIONS]`.

(B) td_wait operator requires `rows: N` option (default: 1). Users will set a query that returns some rows like: `select * where [CONDITIONS]`.
* if N is 1, td_wait operator wraps the query in `select true from (ORIGINAL_QUERY) sub limit 1`
* if N is larger than 1, td_wait operator wraps the query in `select count(*) >= N from (select 1 from (ORIGINAL_QUERY) raw limit N) sub`

(A) is more explicit. (B) is more implicit and has more opportunities to optimize automatically.
I was assuming (A) and this implementation is toward (B).

frsyuki: which do you think better?

frsyuki: In this case: a previous task loads data to a table "tbl" using Data Connector. It should load some rows between `${last_session_time}` to `${session_time}` time range.

With (A):
```
td_wait>:
  data: select true from tbl where td_time_range(time, '${last_session_time}', '${session_time}') limit 1
```

With (B):
```
td_wait>:
  data: select * from tbl where td_time_range(time, '${last_session_time}', '${session_time}')
```

frsyuki: ok, I think (B) is good.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitTableOperatorFactory.java
```diff
@@ -0,0 +1,148 @@
...
+            case "presto":
+                tableName = TDOperator.escapePrestoTableName(table);
+                break;
+            case "hive":
+                tableName = TDOperator.escapeHiveTableName(table);
+                break;
+            default:
+                throw new ConfigException("Unknown 'engine:' option (available options are: hive and presto): " + engine);
+        }
+        String query = "select count(1) from " + tableName + " limit " + expectedRows;
```
frsyuki: I think this count + limit is not working as expected. What you need is:

`select count(*) from (select * from "tableName" limit expectedRows) sub`

frsyuki: Above query works well with Hive. But Presto has an advanced optimizer for COUNT without WHERE.
Following query is faster in Presto:

`select count(*) from "tableName"`

danielnorberg: Ok, will fix 👍

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
danielnorberg: Ok, I'll implement (B). 👍

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
danielnorberg: I was experimenting with having only one `td_wait` operator. This is a leftover from that. Good catch. Will fix. 👍

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
```diff
@@ -0,0 +1,167 @@
...
+            int pollInterval = getPollInterval(params);
+
+            Config state = request.getLastStateParams().deepCopy();
+
+            // Start the query job
+            Optional<String> existingJobId = state.getOptional(JOB_ID, String.class);
+            if (!existingJobId.isPresent()) {
+                String query = templateEngine.templateCommand(workspacePath, params, "query", UTF_8);
+                int rows = params.get("rows", int.class, 1);
+                String jobId = startJob(request, params, query, rows);
```
frsyuki: how about moving startJob into following `try (TDOperator op = TDOperator.fromConfig(params)) {` block? Reason is that OperatorManager shows `Parameter '{}' is not used at task {}.` message if `params.get("engine", String.class, "presto")` is not called. TdWaitTableOperatorFactory also uses this way (existingJobId.isPresent is checked in `try (TDOperator op...` block).

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
```diff
@@ -0,0 +1,167 @@
...
+            logger.info("Started {} job id={}:\n{}", engine, job.getJobId(), query);
+
+            return job.getJobId();
+        }
+    }
+
+    private boolean fetchJobResult(TDJobOperator job, TDJob.Status status)
+    {
+        if (status != TDJob.Status.SUCCESS) {
+            throw new RuntimeException("Poll job failed: " + job.getJobId());
```
frsyuki: I think this is an expected error. So, here should throw TaskExecutionException so that OperatorManager suppresses stacktrace by default. Like TdOperatorFactory.joinJob method does. It also shows execution log of the job.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitTableOperatorFactory.java
```diff
@@ -0,0 +1,184 @@
...
+                    state.remove(JOB_ID);
+                    throw TaskExecutionException.ofNextPolling(pollInterval, ConfigElement.copyOf(state));
+                }
+
+                return TaskResult.empty(request);
+            }
+        }
+
+        private boolean fetchJobResult(int rows, TDJobOperator job)
+        {
```
frsyuki: I think here should check `status != TDJob.Status.SUCCESS` (as like TdWaitOperatorFactory.fetchJobResult does). This code throws `Got unexpected empty result for count job` which is unexpected. Here can throw TaskExecutionException with execution log of the job when the job failed.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
danielnorberg: 👍

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
danielnorberg: Ok, will do.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitTableOperatorFactory.java
danielnorberg: Ok, will do.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
danielnorberg: Going with (A) instead as it is not trivial to wrap queries in a robust way.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
```diff
@@ -0,0 +1,170 @@
...
+
+            ArrayValue row = firstRow.get();
+            if (row.size() < 1) {
+                throw new TaskExecutionException("Got empty row in result of query", ConfigElement.empty());
+            }
+
+            // Wait condition is fulfilled if the first column is not NULL and not FALSE.
+            Value firstCol = row.get(0);
+            return !firstCol.isNilValue() &&
+                    !(firstCol.isBooleanValue() && !firstCol.asBooleanValue().getBoolean());
```
frsyuki: I think some people who are familiar with MySQL will use "select 1" instead of "select true" by mistake because MySQL doesn't have boolean type.
How about adding this condition as well?:
`!(firstCol.isIntegerValue() && !firstCol.asIntegerValue().equals(BigInteger.valueOf(0)))`

frsyuki:
Added one comment. Everything else looks very good 👍

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
danielnorberg: 👍 Will do.

Issue #25, opened 2016-03-30
for_each parameterization capability is limited


danielnorberg:
`for_each>` parameters can be parameterized like below by utilizing YAML anchors:

```yaml
run: +parameterized_for_each

_export:
  foos: &FOOS
    - 1
    - 2

+parameterized_for_each:
  for_each>:
    foo: *FOOS
  _do:
    sh>: "echo hello ${foo}"
```

But it might be useful to allow for parameterizing `for_each>` using actual digdag parameters. E.g.

```yaml
run: +parameterized_for_each

_export:
  foos:
    - 1
    - 2

+parameterized_for_each:
  for_each>:
    foo: @foos
  _do:
    sh>: "echo hello ${foo}"
```

And using parameters explicitly set by e.g. a `py>` task:

```yaml
run: +main

+main:
  +export_foos:
    py>: tasks.export_foos
  +parameterized_for_each:
    for_each>:
      foo: ${foos}
    _do:
      sh>: "echo hello ${foo}"
```

```python
import digdag

def export_foos():
    digdag.env.store({"foos": [1, 2]})
```

Attempting to parameterize `for_each>` like this currently fails with the below error:

```
2016-03-30 10:18:17 +0900: Digdag v0.4.2
2016-03-30 10:18:18 +0900 [WARN] (main): Reusing the last session time 2016-03-29T00:00:00+09:00.
2016-03-30 10:18:18 +0900 [INFO] (main): Using session digdag.status/20160329T000000+0900.
2016-03-30 10:18:18 +0900 [INFO] (main): Starting a new session repository id=1 workflow name=+main session_time=2016-03-29T00:00:00+09:00
2016-03-30 10:18:19 +0900 [INFO] (0020@+main+export_foos): py>: tasks.export_foos
2016-03-30 10:18:19 +0900 [INFO] (0020@+main+parameterized_for_each): for_each>: {foo=[1,2]}
2016-03-30 10:18:19 +0900 [ERROR] (0020@+main+parameterized_for_each): Task failed
io.digdag.client.config.ConfigException: Expected array type for key 'foo' but got "[1,2]" (string)
	at io.digdag.client.config.Config.propagateConvertException(Config.java:400)
	at io.digdag.client.config.Config.readObject(Config.java:391)
	at io.digdag.client.config.Config.get(Config.java:235)
	at io.digdag.client.config.Config.getList(Config.java:277)
[...]

PR #122, opened 2016-06-15
fix ShOperator. add shell option.

toyama0919:
Hello.

I want use login shell.
In addition, I want to choose shell environment variable because it is fixed in "/bin/sh".

Thanks.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
@@ -1,7 +1,11 @@
 package io.digdag.standards.operator;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 import java.io.File;
 import java.io.InputStream;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


toyama0919:
This PR failed test?
My Local OSX is passed test.

frsyuki:
Is difference of shell implementations significant? What kinds of shell do you use?
I think that workflow definition should be reproducible on many environments as much as possible. This change makes workflow more dependent on the execution environments. Alternative idea is to add `shell: /bin/sh` as an option.

toyama0919:
Hmmm..

I want use server environments.
(rbenv, rvm, rails ,java environment... etc)
Then starting cost is very high.

My second idea, default_shell option.
What do you think?

```yaml
  sh>: echo hoge
  default_shell: zsh -l -c
```

frsyuki:
Is `default_` prefix necessary? I think `shell` option is the most flexible:

```
_export:
  sh:
    shell: zsh -l
+step1:
  sh>: echo this uses zsh with -l option
```

The default `shell` is `/bin/sh`.

toyama0919:
@frsyuki 
Ah, OK!
Thanks good advise.

toyama0919:
@frsyuki 
>The default value of shell option is /bin/sh
fixed.

digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
```diff
@@ -66,8 +70,13 @@ public TaskResult runTask()
             Config params = request.getConfig()
                 .mergeDefault(request.getConfig().getNestedOrGetEmpty("sh"));
 
+            String shell = params.get("shell", String.class, "/bin/sh -c");
             String command = params.get("_command", String.class);
-            ProcessBuilder pb = new ProcessBuilder("/bin/sh", "-c", command);
+
+            List<String> cmdline = new ArrayList<String>(Arrays.asList(shell.split(" ")));
```
danielnorberg: Splitting on whitespace seems a little brittle. How about making `shell` a list instead?

toyama0919:
@danielnorberg
Thanks advise!

fixed bellow settings.
Please review.

```ruby
_export:
  sh:
    shell:
      - /bin/zsh
      - '-l'
      - '-c'
```

frsyuki:
including -c is not good idea because it's assuming internal behavior of ShOperator

frsyuki:
@danielnorberg I think it's better if sh operator can accept string as well as array because of #136 or #25 issue. Necessary update is to parse the input using a better tokenizer. Maybe that's a future improvement (because I can't come up with easy code to parse).

frsyuki:
I added a commit and created #143. shell option is now more similar to shebang. So this also works:

```
+calling:
  sh>: p 1 + 1
  shell: [/usr/bin/env, ruby]
  #=> 2
```

toyama0919:
@frsyuki 
Oh, thanks.
That is easy to understand.

PR #125, opened 2016-06-16
Add shebang for Mac OS X to recognize this as valid executable

darklore:
OS X 10.9 cannot recognize digdag.jar as a valid executable without a shebang.
So that, add sh shebang.

### DIFF
--- a/digdag-cli/src/main/sh/selfrun.sh
+++ b/digdag-cli/src/main/sh/selfrun.sh
@@ -1,3 +1,4 @@
+#!/bin/sh
 
 : <<BAT
 @echo off


frsyuki:
It causes an error on Windows. Upgrade to Mac OS X >= 10.10. That's free.

darklore:
OK, I understood it.

UniqMartin:
Just out of curiosity: Have you considered providing a regular `.jar` file and two separate wrapper scripts (one for UNIX-like systems and one for Windows) instead of trying to combine everything in a single file in your regular releases? While this clever hack usually works, it relies on behavior that is nowadays considered obsolete on UNIX-like systems and the presence of a shebang line is heavily preferred.

frsyuki:
@UniqMartin @darklore Now I realized that this issue is for Homebrew.
Ease of installation matters more than wide compatibility, as a command-line client, but I understand that compatibility is also important for Homebrew.

How about installing a script that has following lines in homebrew the formula file?

```
#/bin/sh
exec /bin/sh digdag-(version).jar "$@"
```

I think this is same thing with having a shell-script wrapper file. Similar idea with `bin.write_exec_script`.
As a formula definition:

```ruby
class Digdag < Formula
  desc "Workload Automation System"
  homepage "https://github.com/treasure-data/digdag"
  url "https://dl.digdag.io/digdag-0.8.2.jar"
  sha256 "9e56fe2009a8f3ef62d2e5d3d1b786d0a2a38686c846402acfdcfbb06eae2843"

  bottle :unneeded

  depends_on :java => "1.8+"

  def install
    # Execute through /bin/bash for OS X 10.9.
    # See https://github.com/Homebrew/homebrew-core/pull/2012#issuecomment-226407948 for the detail.
    libexec.install "digdag-#{version}.jar" => "digdag.jar"
    (bin/"digdag").write "#!/bin/bash\nexec /bin/bash \"#{libexec/"digdag.jar"}\" \"$@\"\n".undent
  end

  test do
    ENV.java_cache
    assert_match version.to_s, shell_output("#{bin}/digdag --version")
  end
end
```

(we released v0.8.2 today)

UniqMartin:
@frsyuki Thank you very much for taking a closer look and for that detailed response! I understand your motivation for an easy to use single-file solution. I agree that providing a simple shell wrapper is probably the best solution, both because it keeps the difference between the upstream distribution and the Homebrew formula as minimal as possible and because you've just suggested it as the “official” solution.

@darklore Would you mind adjusting your Homebrew PR accordingly? Then we'll have one final review and can ship this finally. 😸

frsyuki:
@UniqMartin Thank you for your comment.
I created this PR Homebrew/homebrew-core#2261 for Embulk to apply the same thing (I created embulk, and it uses the same hack).

darklore:
@frsyuki  @UniqMartin Thank you very much for comments and valuable advice!
I'll modify my homebrew PR accodingly soon.

PR #127, opened 2016-06-17
make td> and td_load> operators poll instead of block

danielnorberg:
This is a first step towards making all the operators that start td jobs polling instead of blocking. Most of the polling logic should be possible to implement in a manner that it can be reused across all
td job operators.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/td/TDClientFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/td/TDClientFactory.java
@@ -1,6 +1,9 @@
 package io.digdag.standards.operator.td;
 
+import com.google.common.base.Optional;
+import com.treasuredata.client.ProxyConfig;
 import com.treasuredata.client.TDClient;
+import com.treasuredata.client.TDClientBuilder;
 import io.digdag.client.config.Config;
 import io.digdag.client.config.ConfigException;
 
@@ -14,7 +17,15 @@ class TDClientFactory
             throw new ConfigException("Parameter 'apikey' is empty");
         }
 
-        TDClient client = TDClient.newBuilder(false)
+        TDClientBuilder builder = TDClient.newBuilder(false);
+
+        Config proxyConfig = params.getNestedOrGetEmpty("proxy");
+        boolean proxyEnabled = proxyConfig.get("enabled", Boolean.class, false);
+        if (proxyEnabled) {
+            builder.setProxy(proxyConfig(proxyConfig));
+        }
+
+        TDClient client = builder
                 .setEndpoint(params.get("endpoint", String.class, "api.treasuredata.com"))
                 .setUseSSL(params.get("use_ssl", boolean.class, true))
                 .setApiKey(apikey)
@@ -22,5 +33,32 @@ class TDClientFactory
                 .build();
 
         return client;
+    }
+
+    private static ProxyConfig proxyConfig(Config config)
+    {
+        ProxyConfig.ProxyConfigBuilder builder = new ProxyConfig.ProxyConfigBuilder();
+
+        Optional<String> host = config.getOptional("host", String.class);
+        if (host.isPresent()) {
+            builder.setHost(host.get());
+        }
+
+        Optional<Integer> port = config.getOptional("port", Integer.class);
+        if (port.isPresent()) {
+            builder.setPort(port.get());
+        }
+
+        Optional<String> user = config.getOptional("user", String.class);
+        if (user.isPresent()) {
+            builder.setUser(user.get());
+        }
+
+        Optional<String> password = config.getOptional("password", String.class);
+        if (password.isPresent()) {
+            builder.setPassword(password.get());
+        }
+
+        return builder.createProxyConfig();
     }
 }
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
@@ -1,20 +1,22 @@
 package io.digdag.standards.operator.td;
 
-import java.util.Date;
-import java.io.Closeable;
-
+import com.google.common.base.Optional;
 import com.google.common.base.Throwables;
-import io.digdag.client.config.Config;
-import io.digdag.client.config.ConfigException;
-import io.digdag.util.RetryExecutor;
-import io.digdag.util.RetryExecutor.RetryGiveupException;
 import com.treasuredata.client.TDClient;
 import com.treasuredata.client.TDClientException;
 import com.treasuredata.client.TDClientHttpConflictException;
 import com.treasuredata.client.TDClientHttpNotFoundException;
 import com.treasuredata.client.TDClientHttpUnauthorizedException;
-import com.treasuredata.client.model.TDJobRequest;
 import com.treasuredata.client.model.TDExportJobRequest;
+import com.treasuredata.client.model.TDJobRequest;
+import io.digdag.client.config.Config;
+import io.digdag.client.config.ConfigException;
+import io.digdag.util.RetryExecutor;
+import io.digdag.util.RetryExecutor.RetryGiveupException;
+
+import java.io.Closeable;
+import java.util.Date;
+
 import static io.digdag.util.RetryExecutor.retryExecutor;
 
 public class TDOperator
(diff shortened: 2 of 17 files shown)


danielnorberg:
@frsyuki PTAL

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
```diff
@@ -78,46 +85,126 @@ public Operator newTaskExecutor(Path workspacePath, TaskRequest request)
...
+        @Override
+        public TaskResult runTask()
+        {
+            try (TDOperator op = TDOperator.fromConfigWithRetries(params)) {
+
+                // Generate and store domain key before starting the job
+                if (!existingDomainKey.isPresent()) {
+                    String domainKey = UUID.randomUUID().toString();
+                    state.set(DOMAIN_KEY, domainKey);
+                    throw TaskExecutionException.ofNextPolling(0, ConfigElement.copyOf(state.deepCopy()));
```
frsyuki: You don't need `deepCopy()` because `ConfigElement.copyOf` copies deeply.
ConfigElement is an immutable. Config is mutable.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java.orig
```diff
@@ -0,0 +1,557 @@
+package io.digdag.standards.operator.td;
```
frsyuki: I think you wrongly added this file

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDClientFactory.java
```diff
@@ -8,19 +9,27 @@
...
+    static TDClient clientFromConfig(Config params, boolean retries)
+    {
 
         String apikey = params.get("apikey", String.class).trim();
         if (apikey.isEmpty()) {
             throw new ConfigException("Parameter 'apikey' is empty");
         }
 
-        TDClient client = TDClient.newBuilder(false)
+        AbstractTDClientBuilder<TDClient> builder = TDClient.newBuilder(false)
```
frsyuki: why don't you use TDClientBuilder?

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -1,33 +1,43 @@
...
 
 public class TDOperator
         implements Closeable
 {
-    public static TDOperator fromConfig(Config config)
+    public static TDOperator fromConfig(Config config) {
+        return fromConfig(config, false);
+    }
+
+    public static TDOperator fromConfigWithRetries(Config config) {
```
frsyuki: I think TDOperator should use either TDClient's retry or TDOperator's retry. Using both of them means that retrying in TDClient AND retrying in TDOperator. It may retry for unexpected amount of time.
Unless all TDClient API become ready to retry, we should use TDOperator's retry only and always disable TDClient's retry.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
```diff
@@ -78,46 +85,126 @@ public Operator newTaskExecutor(Path workspacePath, TaskRequest request)
...
+
+                // Check if the job is done
+                String jobId = existingJobId.get();
+                TDJobOperator job = op.newJobOperator(jobId);
+                TDJobSummary status = checkJobStatus(job);
+                boolean done = status.getStatus().isFinished();
+                if (!done) {
+                    int prevPollInterval = state.get(POLL_INTERVAL, int.class, INITIAL_POLL_INTERVAL);
+                    int nextPollInterval = Math.min(MAX_POLL_INTERVAL, prevPollInterval * 2);
+                    state.set(POLL_INTERVAL, nextPollInterval);
```
frsyuki: I think it's easier to include number of retries in stateParams.
Polling interval can be calculated using `min(initialInterval * pow(2, retryCount), maxInterval)`.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
```diff
@@ -78,46 +85,126 @@ public Operator newTaskExecutor(Path workspacePath, TaskRequest request)
...
+            this.existingJobId = state.getOptional(JOB_ID, String.class);
+            this.existingDomainKey = state.getOptional(DOMAIN_KEY, String.class);
+        }
+
+        @Override
+        public TaskResult runTask()
+        {
+            try (TDOperator op = TDOperator.fromConfigWithRetries(params)) {
+
+                // Generate and store domain key before starting the job
```
frsyuki: This is good idea. Then we don't submit job twice.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
danielnorberg: 👍 will do.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
danielnorberg: Ok, will do.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDClientFactory.java
danielnorberg: Oops.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java.orig
danielnorberg: Oops.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
danielnorberg: Ok 👍

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
danielnorberg: Maybe there should be a way to persist task state without having to exit / throw `TaskExecutionException`. 

Maybe we should inject an interface that the operator can use to store state directly. This would avoid the extra `TaskExecutionException` and the associated overhead of querying the DB to update the task state etc.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDClientFactory.java
danielnorberg: Ah. Turns out that the setter methods return `AbstractTDClientBuilder`. That's pretty broken.

danielnorberg: https://github.com/treasure-data/td-client-java/pull/54

danielnorberg:
Turns out this interacts badly with the retrying functionality implemented by `BaseOperator`. The polling `TaskExecutionException` thrown by `TdOperatorFactory` is caught by `BaseOperator`, which then creates a new `TaskExecutionException` and discards the polling interval and state etc. 

I guess `BaseOperator` should not do treat a polling `TaskExecutionException` as an error. I guess it should just propagate it if the error is absent.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
frsyuki: > Maybe there should be a way to persist task state without having to exit / throw TaskExecutionException.

Agreed.

> Maybe we should inject an interface that the operator can use to store state directly

Injection by Guice happens for a OperatorFactory. So the injected class can't contain context info like taskId. So basically we need to change Operator API.
Here is my idea: https://github.com/treasure-data/digdag/issues/141#issuecomment-228463857
^ this is procedure-oriented API. This is hard to use it statically (e.g. generating Web UI components from the code). Declarative API is ideal but flexible declarative config API is hard to build.

frsyuki:
> he polling TaskExecutionException thrown by TdOperatorFactory is caught by BaseOperator

That's...not good. At least BaseOperator should exclude TaskExecutionException. So, this is yes > `it should just propagate it if the error is absent`.

Actually yesterday I thought that retrying should be implemented in WorkflowExecutor. Group retrying (`_retry` option at a non-leaf task) is implemented WorkflowExecutor. A reason is that error may happen in a queue. If an agent crashes during processing a task, task queue automatically re-assigns the task to another agent. But this may cause unexpected side effects such as duplicated data loading, etc.

In either cases (task execution error / queue error), error should be notified to WorkflowExecutor. Then users can configure the behavior using parameter consistently: retry the task automatically, back to a previous checkpoint and retry automatically, or make the attempt failed. Some operators might want to tell WorkflowExecutor that this task is idempotent (or not idempotent) after parsing options.

danielnorberg:
Yes, retrying should probably be in WorkflowExecutor. Let's put that on the backlog.

PR #138, opened 2016-06-23
add user info to revision

danielnorberg:
Authenticators can use this to annotate revisions with arbitrary user
information that can then be used by e.g. custom notification senders.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/repository/Revision.java
+++ b/digdag-core/src/main/java/io/digdag/core/repository/Revision.java
@@ -19,6 +19,8 @@ public abstract class Revision
 
     public abstract Optional<String> getArchivePath();
 
+    public abstract Optional<Config> getUserInfo();
+
     public static Revision copyOf(Revision other)
     {
         return ImmutableRevision.builder().from(other).build();
--- a/digdag-core/src/main/java/io/digdag/core/repository/Revision.java
+++ b/digdag-core/src/main/java/io/digdag/core/repository/Revision.java
@@ -19,18 +19,19 @@ public abstract class Revision
 
     public abstract Optional<String> getArchivePath();
 
-    public abstract Optional<Config> getUserInfo();
+    public abstract Config getUserInfo();
 
     public static Revision copyOf(Revision other)
     {
         return ImmutableRevision.builder().from(other).build();
     }
 
-    public static ImmutableRevision.Builder builderFromArchive(String name, ArchiveMetadata meta)
+    public static ImmutableRevision.Builder builderFromArchive(String name, ArchiveMetadata meta, Config userInfo)
     {
         return ImmutableRevision.builder()
             .name(name)
-            .defaultParams(meta.getDefaultParams().deepCopy());
+            .defaultParams(meta.getDefaultParams().deepCopy())
+            .userInfo(userInfo);
     }
 
     @Value.Check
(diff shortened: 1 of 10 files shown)


danielnorberg:
@frsyuki PTAL

digdag-core/src/main/java/io/digdag/core/repository/Revision.java
```diff
@@ -19,6 +19,8 @@
 
     public abstract Optional<String> getArchivePath();
 
+    public abstract Optional<Config> getUserInfo();
```
frsyuki: I think it's better to use `Config` instead of `Optional<Config>` because difference of `null` from `{}` is not always clear. Moreover, when `io.digdag.core.database.ConfigMapper` gets data from a text column, it doesn't distinguish null and `{}` (both of them become `{}`).

AuthRequestFilter can inject ConfigFactory at its constructor so that setProperty can replace absent to `{}` like `requestContext.setProperty("userInfo", result.getUserInfo().or(cf.create()))`

frsyuki:
Nice 👍 
How do you think about comment?

danielnorberg:
Sure, SGTM.

danielnorberg:
Added a commit to make the revision user info field non-optional. This now forces all creation sites of `Revision` to also have a `ConfigFactory` to create an empty `Config` to provide as userInfo. Is this really desirable?

Alternatively we could make `builderFromArchive` initiate the user info field to `{}` but then _it_ needs a `ConfigFactory`. 

Perhaps a deeper problem here is the fact that one needs a `ConfigFactory` to create a `Config`. Perhaps `Config` should not have a `ObjectMapper` field. Perhaps it should be less tightly coupled to jackson. Perhaps the internal representation should simply be a `Map<String, Object>` instead of `ObjectNode`?

digdag-server/src/main/java/io/digdag/server/Authenticator.java
```diff
@@ -1,51 +1,62 @@
...
+
+import javax.annotation.Nullable;
 import javax.ws.rs.container.ContainerRequestContext;
 
+import static org.immutables.value.Value.Style.ImplementationVisibility.PACKAGE;
+
 public interface Authenticator
 {
-    public static class Result
+    @Value.Immutable
```
frsyuki: This generates `io.digdag.server.ImmutableResult` (not `io.digdag.server.Authenticator.ImmutableResult`). I think that is not what you expect.
How about renaming `Authenticator.Result` to `AuthenticatorResult`?

digdag-server/src/main/java/io/digdag/server/Authenticator.java
```diff
@@ -1,51 +1,62 @@
...
+    @Value.Style(visibility = PACKAGE)
```
frsyuki: This interface needs to be public so that a custom authenticator module can instantiate it.

frsyuki:
Well, `Revision.builderFromArchive` already requires ConfigFactory through `meta.getDefaultParams().getFactory()`.
Having ObjectNode doesn't require ObjectMapper. Mapping between JsonNode and Java classes need ObjectMapper. So, `ConfigElement` doesn't has ObjectMapper but also doesn't have get or set methods.
Problem is when we should use `ConfigElement` and when we should use `Config`...and I don't have good ideas.

frsyuki:
Please merge, if you're also good 👍

PR #157, opened 2016-06-30
Remove type field from direct download/upload handles

frsyuki:
Direct download / upload works only for HTTP because digdag supports
REST API only and it seems reasonable limitation. Thus it doesn't make
sense to have "type" in `RestDirectDownloadHandle`, `RestDirectUploadHandle`,
`DirectDownloadHandle`, and `DirectUploadHandle` classes. Having type
field requires unnecessary check of "type" field every time and it won't
be implemented appropriately.

This change includes protocol change.
But Java client and CLI has backward compatibility. They can talk with new server and old server.


digdag-client/src/main/java/io/digdag/client/api/RestDirectDownloadHandle.java
```diff
@@ -1,20 +1,49 @@
...
 {
-    String getType();
-
+    @JsonValue
     String getUrl();
 
-    static ImmutableRestDirectDownloadHandle.Builder builder()
+    @JsonCreator
+    @Deprecated
+    static RestDirectDownloadHandle of(JsonNode node)
```
danielnorberg: Is this necessary?

Isn't disabling `FAIL_ON_UNKNOWN_PROPERTIES` enough? https://github.com/treasure-data/digdag/blob/master/digdag-client/src/main/java/io/digdag/client/DigdagClient.java#L109

frsyuki: This is necessary because this structure:

```
{
    "type": "http",
    "url": "http://example.com"
}
```

is changed to:

```
"http://example.com"
```

danielnorberg:
👍

PR #174, opened 2016-07-20
Rearchitecture task queue to support per-site max concurrency

frsyuki:
* New concept called "queue". A queue belongs to a site.
* Limit number of concurrently running threads per site and queue.
* "default queue tasks" are tasks that are not bound to a custom queue (they're bound to the implicit default queue of a site).
* "queue-bound tasks" are tasks that are bound to a custom queue.
* queues have a boolean flag.
  * If it is true, only "remote agents" run the tasks in the queue. On a hosted service environment, "remote agents" run on an user's data center.
  * If it is false, "shared agents" run the tasks in the queue (the flag is always false if it's the default queue). On a hosted service environment, "shared agents" run on the hosted environment.
* Some API updates to support remote agents through REST API in the future.


digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
```diff
@@ -0,0 +1,495 @@
...
+    }
+
+    private void sleepForEnqueue(long maxSleepMillis)
+    {
+        synchronized (localTaskNoticeHelper) {
+            try {
+                localTaskNoticeHelper.wait(maxSleepMillis);
+            }
+            catch (InterruptedException ex) {
+                return;
```
komamitsu: Why don't you call `Thread.currentThread().interrupt()` to restore the interrupted status?

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
```diff
@@ -0,0 +1,495 @@
...
+    {
+        int siteMaxConcurrency = queueConfig.getSiteMaxConcurrency(siteId);
+
+        try {
+            if (!localLockMap.tryLock(siteId, 500)) {
+                return ImmutableList.of();
+            }
+        }
+        catch (InterruptedException ex) {
+            return ImmutableList.of();
```
komamitsu: Why don't you call `Thread.currentThread().interrupt()` to restore the interrupted status?

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueConfig.java
```diff
@@ -0,0 +1,20 @@
...
+{
+    private final int defaultMaxConcurrency;
+
+    @Inject
+    public DatabaseTaskQueueConfig(Config systemConfig)
+    {
+        this.defaultMaxConcurrency = systemConfig.get("queue.db.max_concurrency", int.class, Integer.MAX_VALUE);
+    }
+
+    public int getSiteMaxConcurrency(int siteId)
```
komamitsu: We can't use a customized value for specific sites?

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
```diff
@@ -626,11 +626,117 @@ public void migrate(Handle handle)
...
+                    " drop column priority");
+            handle.update("alter table queued_tasks" +
+                    " drop column resource_type_id");
+            handle.update("create unique index queued_tasks_on_site_id_task_id on queued_tasks (site_id, task_id)");
+
+            if (isPostgres()) {
+                handle.update(
+                    "CREATE FUNCTION lock_shared_tasks(target_site_id int, target_site_max_concurrency bigint, limit_count int, lock_expire_seconds int, agent_id text) returns setof bigint as $$\n" +
+                    "BEGIN\n" +
+                    "  IF pg_try_advisory_xact_lock(23300, target_site_id) THEN\n" +
```
komamitsu: Why 23300?

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
```diff
@@ -626,11 +626,117 @@ public void migrate(Handle handle)
...
+                    "        where id = any(\n" +
+                    "          select queued_task_locks.id\n" +
+                    "          from queued_task_locks\n" +
+                    "          where lock_expire_time is null\n" +
+                    "          and site_id = target_site_id\n" +
+                    "          and not exists (\n" +
+                    "            select * from (\n" +
+                    "              select queue_id, count(*) as count\n" +
+                    "              from queued_task_locks\n" +
+                    "              where lock_expire_time is not null\n" +
```
komamitsu: `lock_expire_time is not null and lock_expire_time < cast(extract(epoch from statement_timestamp()) as bigint) + lock_expire_seconds` ?

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
```diff
@@ -626,11 +626,117 @@ public void migrate(Handle handle)
...
+                    "                and site_id = target_site_id\n" +
+                    "              group by queue_id\n" +
+                    "            ) runnings\n" +
+                    "            join queues on queues.id = runnings.queue_id\n" +
```
komamitsu: These JOIN and WHERE clauses can be included in the above sub query using HAVING?

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
```diff
@@ -626,11 +626,117 @@ public void migrate(Handle handle)
...
+                    "            where runnings.count >= queues.max_concurrency\n" +
+                    "              and runnings.queue_id = queued_task_locks.queue_id\n" +
+                    "          )\n" +
+                    "          and not exists (\n" +
+                    "            select count(*)\n" +
+                    "            from queued_task_locks\n" +
+                    "            where lock_expire_time is not null\n" +
```
komamitsu: Same as above

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueFactory.java
```diff
@@ -1,187 +1,40 @@
...
     {
-        this.store = store;
-        this.mapper = mapper;
+        this.database = database;
     }
 
+    @Override
     public String getType()
     {
         return "database";
```
komamitsu: How about making this value a constant variable?

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
```diff
@@ -0,0 +1,495 @@
...
+                .bind("queueId", queueId)
+                .bind("siteId", siteId)
+                .execute();
+        }) > 0;
+    }
+
+    @Override
+    public List<TaskRequest> lockSharedAgentTasks(int count, String agentId, int lockSeconds, long maxSleepMillis)
+    {
+        int i = 0;
```
komamitsu: Not used

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
frsyuki: Good point 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
frsyuki: it's just a random number. PostgreSQL uses two INT numbers or one BIGINT number as the identifier of a lock

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
frsyuki: I avoided it because it disables index-only scan (on PostgreSQL 9.6). `lock_expire_time IS NOT NULL` is included in a partial index but `lock_expire_time < ...` condition can't be evaluated using `queued_tasks_shared_grouping` or `queued_tasks_ordering` index. Thus it needs to fetch heap pages.

Effects of not having the condition are not big because `expireLocks` is called periodically.

Query plan without `lock_expire_time < ...`:

```
digdag_test=# explain analyze
select queued_task_locks.id
from queued_task_locks
where lock_expire_time is null
and site_id = 1
and not exists (
  select * from (
    select queue_id, count(*) as count
    from queued_task_locks
    where lock_expire_time is not null
      and site_id = 1
    group by queue_id
  ) runnings
  join queues on queues.id = runnings.queue_id
  where runnings.count >= queues.max_concurrency
    and runnings.queue_id = queued_task_locks.queue_id
)
and not exists (
  select count(*)
  from queued_task_locks
  where lock_expire_time is not null
    and site_id = 1
  having count(*) >= 10
)
order by queue_id, priority desc, id
limit 3
;
                                                                                      QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=6.32..25.92 rows=1 width=16) (actual time=0.010..0.010 rows=0 loops=1)
   InitPlan 1 (returns $0)
     ->  Aggregate  (cost=5.90..5.91 rows=1 width=8) (actual time=0.007..0.007 rows=0 loops=1)
           Filter: (count(*) >= 10)
           Rows Removed by Filter: 1
           ->  Index Only Scan using queued_tasks_shared_grouping on queued_task_locks queued_task_locks_2  (cost=0.13..5.90 rows=1 width=0) (actual time=0.004..0.004 rows=0 loops=1)
                 Index Cond: (site_id = 1)
                 Heap Fetches: 0
   ->  Result  (cost=0.41..20.01 rows=1 width=16) (actual time=0.009..0.009 rows=0 loops=1)
         One-Time Filter: (NOT $0)
         ->  Merge Anti Join  (cost=0.41..20.01 rows=1 width=16) (actual time=0.001..0.001 rows=0 loops=1)
               Merge Cond: (queued_task_locks.queue_id = queued_task_locks_1.queue_id)
               ->  Index Only Scan using queued_tasks_ordering on queued_task_locks  (cost=0.12..5.89 rows=1 width=16) (actual time=0.001..0.001 rows=0 loops=1)
                     Index Cond: (site_id = 1)
                     Heap Fetches: 0
               ->  Materialize  (cost=0.28..14.10 rows=1 width=4) (never executed)
                     ->  Nested Loop  (cost=0.28..14.10 rows=1 width=4) (never executed)
                           ->  GroupAggregate  (cost=0.13..5.91 rows=1 width=12) (never executed)
                                 Group Key: queued_task_locks_1.queue_id
                                 ->  Index Only Scan using queued_tasks_shared_grouping on queued_task_locks queued_task_locks_1  (cost=0.13..5.90 rows=1 width=4) (never executed)
                                       Index Cond: (site_id = 1)
                                       Heap Fetches: 0
                           ->  Index Scan using queues_pkey on queues  (cost=0.15..8.17 rows=1 width=8) (never executed)
                                 Index Cond: (id = queued_task_locks_1.queue_id)
                                 Filter: ((count(*)) >= max_concurrency)
 Planning time: 0.318 ms
 Execution time: 0.090 ms
(27 rows)
```


Query plan with `lock_expire_time < ...`:

```
digdag_test=# explain analyze
select queued_task_locks.id
from queued_task_locks
where lock_expire_time is null
and site_id = 1
and not exists (
  select * from (
    select queue_id, count(*) as count
    from queued_task_locks
    where lock_expire_time is not null
      and lock_expire_time < cast(extract(epoch from statement_timestamp()) as bigint)
      and site_id = 1
    group by queue_id
  ) runnings
  join queues on queues.id = runnings.queue_id
  where runnings.count >= queues.max_concurrency
    and runnings.queue_id = queued_task_locks.queue_id
)
and not exists (
  select count(*)
  from queued_task_locks
  where lock_expire_time is not null
    and lock_expire_time < cast(extract(epoch from statement_timestamp()) as bigint)
    and site_id = 1
  having count(*) >= 10
)
order by queue_id, priority desc, id
limit 3
;
                                                                                    QUERY PLAN
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=6.33..25.94 rows=1 width=16) (actual time=0.011..0.011 rows=0 loops=1)
   InitPlan 1 (returns $0)
     ->  Aggregate  (cost=5.91..5.92 rows=1 width=8) (actual time=0.007..0.007 rows=0 loops=1)
           Filter: (count(*) >= 10)
           Rows Removed by Filter: 1
           ->  Index Scan using queued_tasks_shared_grouping on queued_task_locks queued_task_locks_2  (cost=0.13..5.91 rows=1 width=0) (actual time=0.006..0.006 rows=0 loops=1)
                 Index Cond: (site_id = 1)
                 Filter: (lock_expire_time < (date_part('epoch'::text, statement_timestamp()))::bigint)
   ->  Result  (cost=0.41..20.02 rows=1 width=16) (actual time=0.011..0.011 rows=0 loops=1)
         One-Time Filter: (NOT $0)
         ->  Merge Anti Join  (cost=0.41..20.02 rows=1 width=16) (actual time=0.002..0.002 rows=0 loops=1)
               Merge Cond: (queued_task_locks.queue_id = queued_task_locks_1.queue_id)
               ->  Index Only Scan using queued_tasks_ordering on queued_task_locks  (cost=0.12..5.89 rows=1 width=16) (actual time=0.002..0.002 rows=0 loops=1)
                     Index Cond: (site_id = 1)
                     Heap Fetches: 0
               ->  Materialize  (cost=0.28..14.11 rows=1 width=4) (never executed)
                     ->  Nested Loop  (cost=0.28..14.11 rows=1 width=4) (never executed)
                           ->  GroupAggregate  (cost=0.13..5.92 rows=1 width=12) (never executed)
                                 Group Key: queued_task_locks_1.queue_id
                                 ->  Index Scan using queued_tasks_shared_grouping on queued_task_locks queued_task_locks_1  (cost=0.13..5.91 rows=1 width=4) (never executed)
                                       Index Cond: (site_id = 1)
                                       Filter: (lock_expire_time < (date_part('epoch'::text, statement_timestamp()))::bigint)
                           ->  Index Scan using queues_pkey on queues  (cost=0.15..8.17 rows=1 width=8) (never executed)
                                 Index Cond: (id = queued_task_locks_1.queue_id)
                                 Filter: ((count(*)) >= max_concurrency)
 Planning time: 0.447 ms
 Execution time: 0.063 ms
(27 rows)
```

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
frsyuki: Do you mean like this?:

```
...
and not exists (
  select queue_id, count(*) as count
  from queued_task_locks
  join queues on queues.id = queued_task_locks.queue_id
  where lock_expire_time is not null
    and site_id = 1
    and queued_task_locks.queue_id = queued_task_locks.queue_id
  group by queue_id
  having count(*) >= queues.max_concurrency
)
...
```

This query is invalid because HAVING requires GROUP BY keys or aggregation functions but queues.max_concurrency is none of them:

```
ERROR:  column "queues.max_concurrency" must appear in the GROUP BY clause or be used in an aggregate function
LINE 14:   having count(*) >= queues.max_concurrency
```

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
frsyuki: How?
This subquery doesn't depend on outer query. Thus PostgreSQL compiles this subquery to a InitPlan instead of a anti semi join. If this InitPlan returns 1 or more rows (meaning that target site already exceeds max_concurrency), all following plans won't be executed at all.

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueConfig.java
frsyuki: It's better to have but not implemented yet because I think few people use multiple sites.
On a hosted environment (TD), application should override this class using DI so that it gets max_concurrency from a remote database.

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueFactory.java
frsyuki: hm...I don't know that has any good effects or not.

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
frsyuki: 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
frsyuki: 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
frsyuki: 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
```diff
@@ -0,0 +1,495 @@
...
+                        )
+                        .execute();
+                }
+            });
+            if (c > 0) {
+                logger.warn("{} task locks are expired. Tasks will be retried.", c);
+            }
+        }
+        catch (Throwable t) {
+            logger.error("An uncaught exception is ignored. Lock expireation thread will be retried.", t);
```
komamitsu: `expiration`

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
```diff
@@ -0,0 +1,495 @@
...
+        }
+    }
+
+    public interface Dao
+    {
+        @SqlQuery("select shared_site_id from queues where id = :queueId")
+        Integer getSharedSiteId(@Bind("queueId") long queueId);
+
+        // optimized implementation of
```
komamitsu: Interesting...

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
```diff
@@ -0,0 +1,494 @@
...
+
+    private void deleteTask0(int siteId, long taskLockId, String agentId)
+        throws TaskNotFoundException, TaskConflictException
+    {
+        this.<Boolean, TaskNotFoundException, TaskConflictException>transaction((handle, dao, ts) -> {
+            int count;
+
+            count = dao.deleteQueuedTask(siteId, taskLockId);
+            if (count == 0) {
+                throw new TaskNotFoundException("Deleting lock does not exist: lock id=" + taskLockId + " site id=" + siteId);
```
komamitsu: Don't we need to try to remove the record from `queued_task_locks` just in case even if the return value of `dao.deleteQueuedTask()` is zero? Or create a REFERENCES constraint on the table to `queued_task.id`.

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
```diff
@@ -0,0 +1,494 @@
...
+                    if (data == null) {
+                        // queued_task is deleted after tryLockSharedAgentTasks call.
+                        // it is possible just because there are 2 different transactions.
+                    }
+                    else {
+                        String lockId = formatSharedTaskLockId(taskLockId);
+                        builder.add(decodeLockedTaskObject(data, lockId));
+                    }
+                }
+                return builder.build();
```
komamitsu: > // it is possible just because there are 2 different transactions.

If it can often happen, how about trying next site id like this?
```
                List<TaskRequest> taskRequests = builder.build();
                if (taskRequests.isEmpty()) {
                    continue;
                }
                return taskRequests;
```

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
komamitsu: Oops, overlooked it

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
komamitsu: Ah, I meant https://github.com/treasure-data/digdag/pull/174/files#r71672928, but ignore this comment as well

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueFactory.java
komamitsu: Yeah, you don't need to address it. But I prefer using constants to using string literals at many lines just in order to prevent typo.

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
```diff
@@ -914,7 +930,15 @@ public boolean taskFailed(int siteId, long taskId, String lockId, AgentId agentI
             taskFailed(new TaskControl(store, task), error)
         ).or(false);
         if (changed) {
-            dispatcher.taskFinished(siteId, lockId, agentId);
+            try {
+                dispatcher.taskFinished(siteId, lockId, agentId);
+            }
+            catch (TaskNotFoundException ex) {
+                logger.debug("Ignoring missing task entry error", ex);
```
komamitsu: Why is the log level lower than TaskConflictException?

komamitsu:
👍 
I left some comments, though.

frsyuki:
@komamitsu could you review bbf3fd3 as discussed at https://github.com/treasure-data/digdag/pull/179#discussion_r72044886? Thanks!

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
frsyuki: 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
frsyuki: I see. About inconsistency,

* If `queued_task_locks` doesn't exist but `queued_tasks` exists, the task won't be dequeued for ever. It is a problem in terms of garbage but not a big deal because it doesn't affect other logics.
* If `queued_task_locks` exists but `queued_tasks` doesn't exist, `lockSharedAgentTasks` will go into `// queued_task is deleted after tryLockSharedAgentTasks call.` section. This is not good because it consumes a garbage slot from max_concurrency...

But deleteTask won't be called anyways if this inconsistency happens. So, force-deleting records here doesn't solve this issue...
I think what we can is using transaction appropriately not to cause this issue.

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
frsyuki: I think it doesn't happen often. It is ok to cause some delay (maxSleepMillis at most) in this case.

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
frsyuki: probably it should be warn, too.

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueFactory.java
frsyuki: but this literal is used only once. getType is a sort of constant method.

komamitsu: So should we change this line https://github.com/treasure-data/digdag/blob/c1277203d608ef2ce9919f31cec0821751fabb8b/digdag-core/src/main/java/io/digdag/core/queue/TaskQueueManager.java#L33 to use this method?

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
```diff
@@ -0,0 +1,519 @@
...
+        long taskLockId = parseTaskLockId(lockId);
+        return forceDeleteTask0(taskLockId);
+    }
+
+    private boolean forceDeleteTask0(long taskLockId)
+    {
+        return this.transaction((handle, dao, ts) -> {
+            int taskCount = dao.forceDeleteQueuedTask(taskLockId);
+            int lockCount = dao.forceDeleteQueuedTaskLock(taskLockId);
+            return taskCount > 0 || lockCount > 0;
```
komamitsu: `taskCount > 0 && lockCount > 0` is an expected normal case?

komamitsu:
Added https://github.com/treasure-data/digdag/pull/174/files#r72228819. Other than that look good to me 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueFactory.java
frsyuki: ok, now I understand your concern.

frsyuki:
> taskCount > 0 && lockCount > 0 is an expected normal case?

yes. it is the expected normal case. `taskCount > 0 && lockCount <= 0` and `taskCount <= 0 && lockCount > 0` are unexpected.

komamitsu:
> yes. it is the expected normal case. taskCount > 0 && lockCount <= 0 and taskCount <= 0 && lockCount > 0 are unexpected.

So, let's replace `||` with `&&` at https://github.com/treasure-data/digdag/pull/174#discussion-diff-72228819R244

frsyuki:
@komamitsu why...? currently, forceDeleteTask returns true when some changes are committed in the database no matter whether the state was expected or not.

komamitsu:
Ah, okay. I thought the return value should mean whether the task counts are expected

PR #175, opened 2016-07-20
Support `postgresql` operator

komamitsu:


danielnorberg:
Nice!

Would be great to have tests for this operator. There is a postgresql server in the CI container. We could use that by checking for the `DIGDAG_TEST_POSTGRESQL` environment variable and creating temporary databases there as part of the test run, similar to how [TemporaryDigdagServer](https://github.com/treasure-data/digdag/blob/master/digdag-tests/src/test/java/acceptance/TemporaryDigdagServer.java) creates a new database for each digdag server instance.

frsyuki:
Can I spend some time to confirm following ideas?

* op name: postgresql→pg
* PostgreSQLBaseOperator.getPostgreSQLConnection→static PostgreSQLConnection.open
  * don't extend PostgreSQLBaseOperator
* use timestamp with time zone type as start_at time (column name "time" is not good because its meaning is unclear). It's better to use `now()` on RDBMS instead of calculating time on agent server using System.currentTimeMillis() because `now()` is consistent even if agent server's system clock is wrong.
* varchar status is probably not good. alternative idea is having boolean flag columns such as "done", "success", etc. ("success" column is probably unnecessary (useless) because this plugin can't always update status record when the transaction fails).
* How does this plugin recover in case a query fails after successful submit of the query + disconnection of the connection?
* tryLockStatusRecord is necessary. Digdag doesn't strongly guarantee single execution especially if network is unstable (e.g. remote agent).
* It seems weird to have inheritance between JdbcQueryHelper and JdbcQueryTxHelper.
I think (database operator set (JdbcQueryHelper)) is one class and JdbcQueryTxHelper is another class. Ideally, PostgreSQLOperatorFactory doesn't include postgresql-specific code if (database operator set) is well implemented). Otherwise, (database operator set) is unnecessary and PostgreSQLOperator can implement those methods directly.
* Is PostgreSQLTableParam necessary? Can't SchemaAndTableName replace it?
* `queryHelper.createStatusTable()` should be called only when `strictTransaction` is true
* `queryHelper.dropStatusRecord()` should be called only when `strictTransaction` is true
  * ideally, queryHelper should be created only when `strictTransaction` is true.
  * Or make it polymorphic and create NullTransactionHelper if `strictTransaction` is false.
* `JdbcQueryTxHelper.addStatusRecord` should ignore unique key constraint violation error. Better idea is probably replacing `getStatusRecord` with `getOrCreateStatusRecord` and never return null.
* strictly speaking, if the operator considers accidental concurrent execution, status record should be locked before checked. Thus, `getStatusRecord` should be `getOrCreateAndLockStatusRecord` as like `io.digdag.core.database.DatabaseProjectStoreManager.DatabaseProjectStore.putAndLockProject` method.
* why don't you move CSVWriter to digdag-plugin-utils so that TdOperatorFactory can also use it?
* CSVWriter.close shouldn't throw Exception. it can throw IOException
* PostgreSQL supports upsert now (since 9.5). Current update support will be obsoleted in near future. I think it should assume existence of upsert by design. Thus `uniq_keys` option is removed and add `merge_rule` (or `merge_actoin`) as like embulk-output-{postgresql,mysql} support.
* I think `update_query` option is unnecessary. It can create a prepared statement with the raw query to validate the statement first. Then, create another prepared statement with INSERT INTO or CREATE TABLE AS.
* PostgreSQL's UPDATE and INSERT queries can return results. Thus making it error is not strictly correct behavior.
* I think many people will want to run a SQL script (a SQL file that includes many statements). So, we need some discussion about showing results by default (when SELECT_ONLY).
* don't write messages directly to System.out because task logger can't take them.  Use CommandLogger as Rb/PyOperatorFactory uses. Modify and/or rename CommandLogger and TaskContextCommandLogger to TaskMessageLogger and TaskContextMessageLogger.
* I think executeQueryAndFetchResult can be simpler by passing a thin wrapper of ResultSet to QueryResultHandler instead of using event-driven API. JdbcSchema and JdbcColumn probably unnecessary. They are unnecessarily complicated for digdag (they're necessary complexity for embulk though).

komamitsu:
@danielnorberg Good idea. This p-r is mainly for reviewing the design of PostgreSQL operator. So I didn't write any tests. After reviewing, I'll write unit tests and integration test as you suggested.

Also, I'm a bit tied up now and asked @frsyuki to address his feedback. We also need to make getting status record from PostgreSQL and executing query & update status record atomic.

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
```diff
@@ -0,0 +1,193 @@
...
+                   );
+        }
+
+        @Override
+        public void cleanup()
+        {
+            executeStatement("delete old query status rows from " + escapeIdent(statusTableName) + " table",
+                    String.format(ENGLISH,
+                        "DELETE FROM %s WHERE query_id = ANY(" +
+                        "SELECT query_id FROM %s WHERE completed_at < now() - interval '%d seconds'" +
```
frsyuki: `interval '1' seconds` is the SQL standard. `interval '1 second'` is a PostgreSQL extension. I'm also surprised when I saw that query issued by Tableau through Prestogres but it's for sure.
See http://savage.net.au/SQL/sql-92.bnf.html#interval literal

komamitsu: The original code https://github.com/treasure-data/digdag/pull/175/commits/95f2bd0dfd9e8d5d25043152a9566fe6136528e0#diff-6796cda912b3d95467f3864be5e93189L106 was `interval '%d' seconds` not `interval '%d' second` and PostgreSQL didn't accept the syntax. So I made the change. But I should've changed that to `interval '%d' second` thinking of compatibility with SQL standard... I'll change it again.

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/AbstractJdbcOperator.java
```diff
@@ -180,7 +179,11 @@ public TaskResult runTask()
                     logger.warn("Error during cleaning up status table. Ignoring.", ex);
                 }
 
-                return TaskResult.defaultBuilder(request).build();
+                // It's possible that the query already finished successfully, but this task don't know that
+                // if the previous connection with database was disconnected during waiting result.
+                // So setting `completedAt` in state is needed even if the task was already finished.
+                state.set(COMPLETED_AT, Instant.now());
+                throw TaskExecutionException.ofNextPolling(0, ConfigElement.copyOf(state));
```
frsyuki: is this necessary? If this callback (setting completedAt stateParams) succeeds, another callback (making the task finished) should also succeed. So, I think that having this code doesn't improve risk significantly.

komamitsu: As we discussed on https://github.com/treasure-data/digdag/pull/179/files#r72044886, it can be happen that the task is retried due to an unexpected delay after the status record in PostgreSQL is expired. As a result, inserted data will be duplicated. So, this code stores a status to avoid that problem.

komamitsu:
Addressed comments on https://github.com/treasure-data/digdag/pull/187 and https://github.com/treasure-data/digdag/pull/198

PR #176, opened 2016-07-21
cli: generate workflow that can be pushed to td

danielnorberg:
Instead of generating a workflow that can currently only be run locally, generate a workflow that can be immediately pushed and started on the td hosted digdag, making it easier for new users to get started.

The assumption here is that current users are mostly td users looking to use the hosted digdag offering.


frsyuki:
ok 👍

PR #177, opened 2016-07-21
[WIP] make limits configurable

danielnorberg:
... and remove yet one more place where we're mutating global system
properties.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/Limits.java
+++ b/digdag-core/src/main/java/io/digdag/core/Limits.java
@@ -1,13 +1,24 @@
 package io.digdag.core;
 
+import io.digdag.client.config.Config;
+
+import javax.inject.Inject;
+
 public class Limits
 {
-    private static final long MAX_WORKFLOW_TASKS = Long.valueOf(
-            System.getProperty("io.digdag.limits.maxWorkflowTasks", "1000"));
+    private static final long DEFAULT_MAX_WORKFLOW_TASKS = Long.valueOf(
+            System.getProperty("io.digdag.limits.defaultMaxWorkflowTasks", "1000"));
 
-    // TODO (dano): this should be configurable by config file etc and not just system property
+    private final long maxWorkflowTasks;
 
-    public static long maxWorkflowTasks() {
-        return MAX_WORKFLOW_TASKS;
+    @Inject
+    public Limits(Config systemConfig)
+    {
+        maxWorkflowTasks = systemConfig.get("limits.max-workflow-tasks", Long.class, DEFAULT_MAX_WORKFLOW_TASKS);
+    }
+
+    public long maxWorkflowTasks()
+    {
+        return maxWorkflowTasks;
     }
 }
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/ForEachOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/ForEachOperatorFactory.java
@@ -27,10 +27,13 @@ public class ForEachOperatorFactory
         implements OperatorFactory
 {
     private static Logger logger = LoggerFactory.getLogger(ForEachOperatorFactory.class);
+    private Limits limits;
 
     @Inject
-    public ForEachOperatorFactory()
-    { }
+    public ForEachOperatorFactory(Limits limits)
+    {
+        this.limits = limits;
+    }
 
     public String getType()
     {
@@ -43,7 +46,7 @@ public class ForEachOperatorFactory
         return new ForEachOperator(workspacePath, request);
     }
 
-    private static class ForEachOperator
+    private class ForEachOperator
             extends BaseOperator
     {
         public ForEachOperator(Path workspacePath, TaskRequest request)
@@ -90,7 +93,7 @@ public class ForEachOperatorFactory
                 .build();
         }
 
-        private static List<Config> buildCombinations(ConfigFactory cf, Map<String, List<JsonNode>> entries)
+        private List<Config> buildCombinations(ConfigFactory cf, Map<String, List<JsonNode>> entries)
         {
             List<Config> current = new ArrayList<>();
             for (Map.Entry<String, List<JsonNode>> pair : entries.entrySet()) {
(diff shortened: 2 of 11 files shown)


danielnorberg:
@frsyuki PTAL

digdag-core/src/main/java/io/digdag/core/Limits.java
```diff
@@ -1,13 +1,24 @@
...
+import io.digdag.client.config.Config;
+
+import javax.inject.Inject;
+
 public class Limits
 {
-    private static final long MAX_WORKFLOW_TASKS = Long.valueOf(
-            System.getProperty("io.digdag.limits.maxWorkflowTasks", "1000"));
+    private static final long DEFAULT_MAX_WORKFLOW_TASKS = Long.valueOf(
+            System.getProperty("io.digdag.limits.defaultMaxWorkflowTasks", "1000"));
```
frsyuki: I think systemConfig should work as like this:

1. get values from JVM system property with "digdag." prefix. For example, `java -Ddigdag.notification.type=shell -Ddigdag.notification.shell.command=reboot" will set `{notification.type: shell, notification.shell.command: reboot}` to systemConfig.
2. CLI read config from `-c PATH.properties` file and set them to systemConfig. `server` command requires `-c` argument and doesn't read from default config file on home dir that is for local-mode execution.
3. some CLI commands parse cmdline arguments and set them to systemConfig. For example, `server` command parse `--task-log PATH` argument and set `{log-server.type: local, log-server.local.path: PATH}` to systemConfig.
4. CLI parse `-X KEY=VALUE` cmdline arguments and set them to systemConfig.

(Later numbers have higher priority)

With this idea, we don't need `io.digdag.limits.defaultMaxWorkflowTasks` because setting -Ddigdag.limits.max-workflow-tasks` means setting default limits.max-workflow-tasks. This needs less code and less documents. To take this idea, currently 1 doesn't work. 2 needs some fix.

digdag-core/src/main/java/io/digdag/core/Limits.java
```diff
@@ -1,13 +1,24 @@
 package io.digdag.core;
 
+import io.digdag.client.config.Config;
+
+import javax.inject.Inject;
+
 public class Limits
```
frsyuki: I wonder that class name could be WorkflowExecutorConfig because agents won't be able to use this, precisely speaking (commented below).

digdag-standards/src/main/java/io/digdag/standards/operator/ForEachOperatorFactory.java
```diff
@@ -27,10 +27,13 @@
         implements OperatorFactory
 {
     private static Logger logger = LoggerFactory.getLogger(ForEachOperatorFactory.class);
+    private Limits limits;
 
     @Inject
-    public ForEachOperatorFactory()
-    { }
+    public ForEachOperatorFactory(Limits limits)
```
frsyuki: I think this doesn't always work in this case:

* this agent runs on a remote machine.
* this agent doesn't have limits.max-workflow-tasks config (uses default value).
* server overwrites limits.max-workflow-tasks to a large number.
* a for_each> operator runs on the remote agent.

Number of tasks generated by this `for_each` operator may exceed default limits.max-workflow-tasks but it may be allowed on the server's environment.

frsyuki: Operators need to get limits from the server through TaskRequest (queue) or through TaskCallbackApi (REST API).

frsyuki: How about this (get Limits through TaskCallbackApi)?:

```diff
diff --git a/digdag-core/src/main/java/io/digdag/core/agent/TaskCallbackApi.java b/digdag-core/src/main/java/io/digdag/core/agent/TaskCallbackApi.java
index 735b9a3..03cef93 100644
--- a/digdag-core/src/main/java/io/digdag/core/agent/TaskCallbackApi.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/TaskCallbackApi.java
@@ -49,4 +49,6 @@ public interface TaskCallbackApi
             int projectId,
             String workflowName)
         throws ResourceNotFoundException;
+
+    Limits getLimits(int siteId);
 }
diff --git a/digdag-spi/src/main/java/io/digdag/spi/OperatorFactory.java b/digdag-spi/src/main/java/io/digdag/spi/OperatorFactory.java
index 911dafb..bb02d62 100644
--- a/digdag-spi/src/main/java/io/digdag/spi/OperatorFactory.java
+++ b/digdag-spi/src/main/java/io/digdag/spi/OperatorFactory.java
@@ -6,5 +6,5 @@ public interface OperatorFactory
 {
     String getType();
 
-    Operator newTaskExecutor(Path workspacePath, TaskRequest request);
+    Operator newTaskExecutor(TaskContext context, TaskRequest request);
 }
diff --git a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
index 83a2d77..bd1bd53 100644
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -256,7 +256,19 @@ public class OperatorManager
         if (factory == null) {
             throw new ConfigException("Unknown task type: " + type);
         }
-        Operator operator = factory.newTaskExecutor(workspacePath, mergedRequest);
+        Operator operator = factory.newTaskExecutor(new TaskContext() {
+            @Override
+            public Path getWorkspacePath()
+            {
+                return workspacePath;
+            }
+
+            @Override
+            public Limits getLimits()
+            {
+                return callback.getLimits(mergedRequest.getSiteId());
+            }
+        }, mergedRequest);
 
         return operator.run();
     }
```

TaskContext can have other methods such as `setStateParams(Config)`

danielnorberg:
Agreed 👍

kuenishi:
Any plan to merge this commit? I may have a use case for ~100000 tasks dynamically generated from file names in a directory (otherwise I have to re-invent multiprocess task scheduler as a single Digdag task). Also, my use case does not have remote agents and I think system administrator is responsible for syncing `limits.max-workflow-tasks` among agents and masters.

frsyuki:
@kuenishi here is the description why limits.max-workflow-tasks exists and what's necessary to increase it: https://github.com/treasure-data/digdag/issues/508
On your machine to test larger max-workflow-tasks, can simply give `-X io.digdag.limits.maxWorkflowTasks=N` option to `server` subcommand.

yoyama:
I will close in near future, if no one claim to keep it.

PR #179, opened 2016-07-22
[WIP] Deterministic pg operator

frsyuki:
TBD


digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
```diff
@@ -0,0 +1,188 @@
...
+    @Override
+    public String escapeIdent(String ident)
+    {
+        try {
+            StringBuilder buf = new StringBuilder();
+            escapeIdentifier(buf, ident);
+            return buf.toString();
+        }
+        catch (SQLException e) {
+            throw new IllegalArgumentException("Invalid identifier name (contains \\0 character): " + ident);
```
komamitsu: Pass `e` as a second parameter

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
```diff
@@ -0,0 +1,188 @@
...
+                   );
+        }
+
+        @Override
+        public void cleanup()
+        {
+            executeStatement("delete old query status rows from " + escapeIdent(statusTableName) + " table",
+                    String.format(ENGLISH,
+                        "DELETE FROM %s WHERE query_id = ANY(" +
+                        "SELECT query_id FROM %s WHERE completed_at < now() - interval '24' hours" +
```
komamitsu: It seems `24 hours` should be configurable

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
```diff
@@ -0,0 +1,188 @@
...
+                        escapeIdent(statusTableName))
+                    );
+        }
+
+        @Override
+        protected StatusRow lockStatusRow(UUID queryId)
+        {
+            try (Statement stmt = connection.createStatement()) {
+                ResultSet rs = stmt.executeQuery(String.format(ENGLISH,
+                            "SELECT completed_at FROM %s WHERE query_id = '%s' FOR UPDATE",
```
komamitsu: Without NOWAIT, it blocks the agent. Is it okay?

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
```diff
@@ -0,0 +1,188 @@
...
+                            escapeIdent(statusTableName),
+                            queryId.toString())
+                        );
+                if (rs.next()) {
+                    // status row exists and locked. get status of it.
+                    Timestamp completedAt = rs.getTimestamp(1);
```
komamitsu: This local variable is not used.

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/AbstractPersistentTransactionHelper.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        else {
+            // query is not completed. run the action.
+            action.run();
+            updateStatusRowAndCommit(queryId);
+            return true;
+        }
+    }
+
+    protected boolean beginTransactionAndLockStatusRow(UUID queryId)
```
komamitsu: 👍

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/AbstractJdbcOperator.java
```diff
@@ -0,0 +1,224 @@
...
+                        connection.executeUpdate(statement);
+                    }
+                });
+
+                if (!executed) {
+                    logger.debug("Query is already completed according to status table. Skipping statement execution.");
+                }
+
+                try {
+                    txHelper.cleanup();
```
komamitsu: I see. With cleaning up old garbage status records, we don't need to have the status of a query in Digdag.

But I have a concern in the following case:
1. Digdag workflow started a query`INSERT INTO sales_summary SELECT * FROM today_sales`
2. `queryId=1234` was created and stored in Digdag state
3. A status record `query_id=1234, created_at=2016-01-01 00:00:00` was inserted to status table in PostgreSQL
4. The query that includes INSERT INTO finished successfully
5. The status record was updated to `query_id=1234, created_at=2016-01-01 00:00:00, completed_at=2016-01-01 00:11:11`
6. But the Digdag agent process crashed before returning from `runTask()`
7. Also the retry of this task was somehow delayed more than 24 hours and another workflow or task removed the status record of this query from PostgreSQL
8. This query was retried after that and it executed INSERT INTO again since the status record was already removed

It's unlikely this case happens and maybe we can ignore it for now. What do you think?

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/AbstractJdbcResultSet.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Override
+    public List<Object> next()
+    {
+        try {
+            if (!resultSet.next()) {
+                return null;
+            }
+            getObjects(results);
```
komamitsu: The return value `Arrays.asList(results)` shares the instance variable`Object[] results` internally. We'd better instantiate a new `Object[]` here for a return value.

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/CsvWriter.java
```diff
@@ -2,15 +2,16 @@
...
 import java.io.IOException;
 import java.io.Writer;
+import java.io.Closeable;
 import java.util.List;
 import java.util.concurrent.Exchanger;
 
-public class CSVWriter
-    implements AutoCloseable
+public class CsvWriter
+    implements Closeable
```
komamitsu: Why did you change the interface to `Closable`? Just for limiting an exception class from `close()` method?

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
frsyuki: According to [PostgreSQL JavaDoc](https://jdbc.postgresql.org/development/privateapi/org/postgresql/core/Utils.html#escapeIdentifier-java.lang.StringBuilder-java.lang.String-), escapeIdentifier throws SQLException only if the string contains a \0 character. Stacktrace is

```
irb(main):003:0* org.postgresql.core.Utils.escapeIdentifier(java.lang.StringBuilder.new, "ab\0").toString
Java::OrgPostgresqlUtil::PSQLException: Zero bytes may not occur in identifiers.
	from org.postgresql.core.Utils.doAppendEscapedIdentifier(org/postgresql/core/Utils.java:206)
	from org.postgresql.core.Utils.escapeIdentifier(org/postgresql/core/Utils.java:188)
	from java.lang.reflect.Method.invoke(java/lang/reflect/Method.java:498)
	from RUBY.<eval>((irb):3)
```

I think we don't have to include stacktrace.

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
frsyuki: done

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
frsyuki: yes. What's your concern?

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/AbstractJdbcResultSet.java
frsyuki: hm. maybe. fixed.

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/CsvWriter.java
frsyuki: no. `close()` method can be declared to throw IOException even if the class implements AutoCloseable.
Maybe the reason is that Closeable.close is guaranteed be idempotent (can be called multiple times without exceptions) but AutoCloseable.close isn't required to be idempotent, according to [JavaDoc](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html).

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
frsyuki: 👍  I should have checked findbugs.

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/AbstractJdbcOperator.java
frsyuki: > delayed more than 24 hours

I see. That may happen. It seems better to store status on digdag status params as well.

frsyuki: I found that it may happen but it can't be solved using state params.

* State params are copied to a queue when the task is queued.
* State params are set when callback (either taskFinished or retryTask) to digdag server succeeded.
* If taskFinished callback to digdag server succeeded, the task won't be queued again (as long as digdag server doesn't have a bug around DB transaction).
* If task execution on an agent was delayed more than 24 hours, it means that the agent has old state params which was copied to task queue 24 hours ago. Checking this state params won't prevent duplicated task execution.

To mitigate this issue, task execution needs to check task status on the server before starting the task (as PerfectQueue does). If the task is marked as finished (success, error, error) on the server, agent skips the task. But this is still not the perfect solution because task execution may be delayed between successful status check and query execution.

frsyuki: I addressed this issue with #174 (bbf3fd37b347c01e68fa16e37dca9278df9b6f24), although it is not the perfect solution.

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
komamitsu: If the later agent gives up immediately with NOWAIT, I guessed Digdag's scheduler can use the agent resource for other tasks. I'm not sure if it's true, though.

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
```diff
@@ -103,7 +103,7 @@ public void cleanup()
             executeStatement("delete old query status rows from " + escapeIdent(statusTableName) + " table",
                     String.format(ENGLISH,
                         "DELETE FROM %s WHERE query_id = ANY(" +
-                        "SELECT query_id FROM %s WHERE completed_at < now() - interval '%d' seconds" +
+                        "SELECT query_id FROM %s WHERE completed_at < now() - interval '%d seconds'" +
```
frsyuki: `interval '1' seconds` is the SQL standard. `interval '1 second'` is a PostgreSQL extension. I'm also surprised when I saw that query issued by Tableau through Prestogres but it's for sure.
See http://savage.net.au/SQL/sql-92.bnf.html#interval literal

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
frsyuki: oh I see. yes, you're right.
If the target PostgreSQL supports SKIP LOCKED (PostgreSQL >= 9.5), here can be non-blocking so that this agent thread can run other operators meanwhile.

frsyuki: SKIP LOCKED doesn't return number of skipped records. We can't tell the record exists or not with simple use of SKIP LOCKED.
An idea is to use UNION ALL like this:

```
SELECT completed_at FROM %s WHERE query_id = '%s'
UNION ALL
SELECT completed_at FROM %s WHERE query_id = '%s' FOR UPDATE SKIP LOCKED
```

If it returns 2 rows, status row is locked. If it returns 1 row, status row exists but locked. If it returns 0 rows, status row doesn't exist.

In addition to this hack, plugin needs to check availability of SKIP LOCKED syntax. It sounds good idea to add a comment here about this idea but it's totally OK to put implementation off.

komamitsu: It's an interesting hack. But just using NOWAIT is simpler without any downside. And even PostgreSQL 8.1 can use NOWAIT. Why don't we use NOWAIT?

frsyuki: NOWAIT is better 👍

PR #182, opened 2016-07-27
Graceful shutdown of HTTP server and worker threads

frsyuki:
This change implements graceful shutdown to HTTP server and MultiThreadAgent threads.

* Undertow creates XnioWorker automatically but here doesn't use it. If we use it, `Undertow.stop` calls `xnioWorker.shutdownNow` immediately. However, it shouldn't be called until `GracefulShutdownHandler.shutodwn` is called to make sure that the HTTP handlers are not processing requests. Therefore, here creates XnioWorker and controls shutdown order manually.
* Upgraded Undertow to 1.4.x because `Undertow.Builder.setWorker` is not available in 1.3.
* A new shutdownHandler waits for completion of `ServerControl.destroy()`. It calls `@PreDestroy` methods. Thus shutting down digdag server will wait for completion of `@PreDestroy LocalAgentManager.shutdown` method.
* `LocalAgentManager.shutdown` calls `MultiThreadAgent.shutdown`. It waits for completion of all running tasks.


frsyuki:
@danielnorberg PTAL
Multi-process agent is a next step that might not be implemented soon :-p

digdag-core/src/main/java/io/digdag/core/agent/MultiThreadAgent.java
```diff
@@ -41,34 +44,56 @@ public MultiThreadAgent(AgentConfig config, AgentId agentId,
...
+        }
+        if (maximumCompletionWait.isPresent()) {
+            long seconds = maximumCompletionWait.get().getSeconds();
+            if (!executor.awaitTermination(seconds, TimeUnit.SECONDS)) {
+                logger.warn("Some tasks didn't finish within maximum wait time ({} seconds)", seconds);
+            }
+        }
+        else {
+            // no maximum wait time. waits for ever
+            while (!executor.awaitTermination(24, TimeUnit.HOURS))
```
danielnorberg: Maybe log something to at least indicate why the agent is not terminating?

danielnorberg:
Does this mean that the digdag rest api becomes unavailable until all currently running operators finish?

digdag-core/src/main/java/io/digdag/core/agent/MultiThreadAgent.java
frsyuki: 👍  Added.

frsyuki:
> Does this mean that the digdag rest api becomes unavailable until all currently running operators finish?

Yes. It's another reason why we should run agents on different machine.

danielnorberg:
Ok, please add a test verifying that the rest api is available while the server waits for task completion.

frsyuki:
@danielnorberg can I ask you to check new test?
The test runs only on unix systems because there're no good ways for Java to send signals to a PID on Windows.

frsyuki:
Undertow 1.4.0.Final seems released but not on maven central yet somehow.
https://github.com/undertow-io/undertow/commits/1.4.x

Release note:
https://issues.jboss.org/secure/ReleaseNote.jspa?projectId=12312521&version=12330950

digdag-tests/src/test/java/acceptance/ServerGracefulShutdownIT.java
```diff
@@ -0,0 +1,172 @@
...
+                    "-c", config.toString(),
+                    "-e", server.endpoint(),
+                    "server_graceful_shutdown", "sleep",
+                    "--session", "now");
+            assertThat(startStatus.code(), is(0));
+            attemptId = getAttemptId(startStatus);
+        }
+
+        // Wait for the task to start
+        for (int i = 0; i < 30; i++) {
```
danielnorberg: Use `TestUtils.expect()` instead?

https://github.com/treasure-data/digdag/blob/master/digdag-tests/src/test/java/utils/TestUtils.java#L196

danielnorberg:
👍

frsyuki:
Now I could download undertow 1.4.0.Final, somehow... Updated.

digdag-tests/src/test/java/acceptance/ServerGracefulShutdownIT.java
frsyuki: 👍  Updated.

danielnorberg:
👍  LGTM

PR #185, opened 2016-08-02
secret management

danielnorberg:
* Separate secrets from parameters.
* Add a pluggable secret storage facility with a default implementation that encrypts and stores secrets in a database table.
* Allow Authenticator implementations to provide a set of request context default secrets.
* Add REST api endpoints that can be used to set/delete/list secrets.
* Add a secrets cli command that can be used to set/delete/list secrets.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/AESGCMSecretCrypto.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/AESGCMSecretCrypto.java
@@ -1,0 +1,213 @@
+package io.digdag.core.database;
+
+import com.google.common.base.Preconditions;
+import com.google.common.base.Throwables;
+import io.digdag.core.crypto.SecretCrypto;
+import io.digdag.core.crypto.SecretCryptoException;
+
+import javax.crypto.AEADBadTagException;
+import javax.crypto.BadPaddingException;
+import javax.crypto.Cipher;
+import javax.crypto.IllegalBlockSizeException;
+import javax.crypto.NoSuchPaddingException;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+
+import java.nio.ByteBuffer;
+import java.security.InvalidAlgorithmParameterException;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.security.SecureRandom;
+import java.util.Arrays;
+import java.util.Base64;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static javax.crypto.Cipher.ENCRYPT_MODE;
+
+/**
+ * A {@link SecretCrypto} implementation that encrypts secrets using AES in GCM mode, padding to hide the actual size of the secret.
+ *
+ * <p>The encrypted record format is (offsets in bytes):</p>
+ *
+ * <pre>
+ * 0          6         10     14      26                  2074      2090
+ * +-----------------------------------------------------------------+
+ * | "aesgcm" | version | term | nonce | encrypted payload | gcm tag |
+ * +-----------------------------------------------------------------+
+ *      6B        4B       4B      12B         2048B           16B
+ * </pre>
+ *
+ * <p>The encrypted payload contains a 32 bit length integer and 2044 Bytes of text + padding. Both the length and the text + padding is encrypted.</p>
+ *
+ */
+public class AESGCMSecretCrypto implements SecretCrypto
+{
+    private static final SecureRandom SECURE_RANDOM = new SecureRandom();
+
+    private static final String NAME = "aesgcm";
+    private static final byte[] NAME_BYTES = NAME.getBytes(UTF_8);
+
+    private final SecretKey sharedSecret;
+
+    private static final int AES_KEY_SIZE = 128;
+    private static final int GCM_NONCE_LENGTH = 12;
+    private static final int GCM_TAG_LENGTH = 16;
+
+    private static final int TERM = 1;
+    private static final int VERSION = 1;
+
+    private static final int MAX_PLAINTEXT_LENGTH = 1024;
+
+    private static final int RECORD_SIZE = 2048;
+    private static final int LENGTH_SIZE = 4;
+    private static final int TEXT_SIZE = RECORD_SIZE - LENGTH_SIZE;
+    private static final int NAME_SIZE = NAME_BYTES.length;
+    private static final int TERM_SIZE = 4;
+    private static final int VERSION_SIZE = 4;
+    private static final int OPAQUE_SIZE =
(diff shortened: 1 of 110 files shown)


digdag-server/src/main/java/io/digdag/server/Authenticator.java
```diff
@@ -57,6 +59,19 @@ default boolean isAccepted()
         String getErrorMessage();
 
         Optional<Config> getUserInfo();
+
+        Optional<Map<String, String>> getSecrets();
```
frsyuki: I think it's better to make this `Optional<Supplier<Map<String, String>>>`. With `Optional<Map>`, Authenticator always needs to fetch security information from somewhere even when controller doesn't need them.

digdag-core/src/main/java/io/digdag/core/database/AESGCMSecretCrypto.java
```diff
@@ -0,0 +1,177 @@
...
+            throw Throwables.propagate(e);
+        }
+
+        assert cipherText.length == RECORD_SIZE + GCM_TAG_LENGTH;
+
+        byte[] opaque = new byte[OPAQUE_SIZE];
+        ByteBuffer output = ByteBuffer.wrap(opaque);
+
+        output.putInt(TERM);
+        output.putInt(VERSION);
```
frsyuki: Having TERM and VERSION is good so that we can change implementation of AESGCMSecretCrypto. But this doesn't solve a problem where we can't change implementation of SecretCrypto at all once we stored some data in DB because the new SecretCrypto can't decode data encrypted by the previous SecretCrypto (e.g. an user started digdag with default settings (DisabledSecretCrypto), and changing settings to use AESGCMSecretCrypto makes already-stored secrets invisible, or makes the server always returning HTTP 500).
I think that better idea is either:

* (A) give up extensibility of SecretCrypto interface completely. DatabaseSecretStore always uses AESGCMSecretCrypto (probably it will be simply renamed to SecretCrypto class) and the class stores encrypted-or-not-flag, algorithm-version, algorithm-name, etc. into the encoded data.

* (B) DatabaseSecretStore stores type of SecretCrypto in addition to the data using new `secrets.crypto_type text` field (text) or `secrets.crypto_type_id int` field (enum). All available SecretCrypto implementations will be injected to DatabaseConfig to decode stored data. In addition to that, default (latest) SecretCrypto is configured to encrypt new records. With this way, AESGCMSecretCrypto maybe doesn't have to include TERM or VERSION because we can create new classes like `AESGCMSecretCryptoV2`, `AESGCMSecretCryptoV3`, ... and let DatabaseSecretStore to choose appropriate one baed on `crypto_type[_id]` field. (TERM is still good to have as a magic header bytes for validation).

I think (B) is better, although it needs more code.

danielnorberg: Agree that we should store the type of crypto as well.

frsyuki:
I was thinking that returning `Config` is good as the plugin API but I also agree that it increases security concern because plugins may easily expose them through log messages.

But at the same time, calling SecretProvider.getSecretOptional and then call Config.getOptional for each config parameter need a lot of code...

How about these ideas?:

Idea 1:

* Add `io.digdag.util.OperatorConfig` class. This is a wrapper class for operator plugins to access options.
* OperatorConfig has similar methods to Config but it's immutable (no setter methods). It also doesn't have methods that involves security concerns such as `convert`, `getNested`, `getKeys`, and `getInternalObjectNode` methods.
* Operator plugins instantiate OperatorConfig from secretSelector (list of secret keys), SecretProvider, and Config. OperatorConfig looks up a key from SecretProvider first if it matches with secretSelector, and then from Config.

Idea 2:

* Rename `io.digdag.client.config.Config` class to `MutableConfig`.
* Add `io.digdag.client.config.Config` interface. Config interface has same methods to MutableConfig class excepting setter methods and insecure methods.
* Add `io.digdag.util.OperatorConfig` class that implements Config. OperatorConfig is created from SecretSelector, SecretProvider and Config. It looks up a key from SecretProvider first if a key matches with SecretSelector, then from another Config (SecretSelector is a list of strings).
* OperatorConfig class overrides toString so that it masks secret values with `***`. With this way, plugins won't get secret values unless plugins intentionally get them.
* Config is not always serializable (because OperatorConfig can't be serializable). Thus model classes need to use `ImmutableConfig` class instead. Add `io.digdag.client.config.ImmutableConfig` class that implements Config.

We may work on plugin API later. What's your opinions around plugin API?
If the data format stored on DB is fixed, let's merge 👍

digdag-core/src/main/java/io/digdag/core/database/AESGCMSecretCrypto.java
danielnorberg: I believe `TERM` is still useful for key rotation purposes.

danielnorberg:
I would prefer to have workflow _parameters_ and _secrets_ be separate concepts with separate interfaces that can better cater to the different respective requirements.

For workflow parameters convenience is paramount. Code interacting with workflow parameters should not need to worry about whether a parameter is a secret or not and whether it can be logged or not.

As for secrets, I believe we should strive to make it hard to leak them by accident. This is more important than convenience. Also, when reviewing code, it should be obvious when something is interacting with or touching secrets. This is part of the rationale for implementing the `SecretProvider` interface instead of using a more generic interface like Map/Config/etc for secrets.

I regard the current behavior of first getting credentials from secrets and then falling back to workflow parameters to be a regrettable anti-pattern. For a while we need to have the td operators be backwards compatible to give users time to migrate their apikeys etc to secrets. Moving forward however I believe that operators should be written to _only_ accept credentials from secrets and _not_ fall back to workflow parameters. 

As we can see in at least one recent TD customer issue, users are currently putting their apikeys in cleartext into their workflow files. It is uncomfortably likely that these workflow files are then stored and managed in an unsecure manner. E.g., they are mailed to colleagues and our customer support, pasted in chat rooms, get stored in our JIRA, perhaps pushed to git repositories etc. This means that by accident slack accounts, mail accounts, jira server, smartphone devices, etc are now insecurely storing credentials that can be used to access TD customer business data. I want to strongly discourage this practice. I believe one of the most effective ways to do this is to make it impossible to put credentials in workflow parameters.

As for the amount of extra code that is currently needed for operators to support getting credentials from both secrets and parameters, I do not believe that it is currently so great that it needs to be addressed by adding more tooling that abstracts these away. The td operators are already reusing the code that instantiates the td api clients etc in this fashion. Assuming that future operators will not be merging workflow parameters and secrets, I do not believe that the net reduction in lines of code and complexity would be worth it. I also believe that this would have negative security implications by making code audit harder etc.

digdag-core/src/main/java/io/digdag/core/database/AESGCMSecretCrypto.java
danielnorberg: Note that `DisabledSecretCrypto` really means disabled, not clear-text. Any attempt to use secrets without explicitly setting `digdag.secret-encryption-key` will fail with an error.

PR #187, opened 2016-08-02
Review for p-r #175

komamitsu:
Commits between 86eaa61 and e449c8a


digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
```diff
@@ -116,7 +119,7 @@ protected StatusRow lockStatusRow(UUID queryId)
         {
             try (Statement stmt = connection.createStatement()) {
                 ResultSet rs = stmt.executeQuery(String.format(ENGLISH,
-                            "SELECT completed_at FROM %s WHERE query_id = '%s' FOR UPDATE",
+                            "SELECT completed_at FROM %s WHERE query_id = '%s' FOR UPDATE NOWAIT",
```
frsyuki: I think you need to check SQLError because `NOWAIT` throws exception if the row is locked, according to postgresql's document.

frsyuki: SQL error code is "55P03":

```
embulk_demo=# select * from queues for update nowait;
ERROR:  55P03: could not obtain lock on row in relation "queues"
LOCATION:  heap_lock_tuple, heapam.c:4679
```

frsyuki: StatusRow may need another field like `EXISTS_CANNOT_LOCK`

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/AbstractJdbcOperator.java
```diff
@@ -180,7 +180,11 @@ public TaskResult runTask()
                     logger.warn("Error during cleaning up status table. Ignoring.", ex);
                 }
 
-                return TaskResult.defaultBuilder(request).build();
+                // It's possible that the query already finished successfully, but this task don't know that
+                // if the previous connection with database was disconnected during waiting result.
+                // So setting `completedAt` in state is needed even if the task was already finished.
```
frsyuki: I don't know why this is needed...as I commented at https://github.com/treasure-data/digdag/pull/179#discussion_r72129236.
If here calls REST API to set only state params, this is good. Because setting state params only probably less likely fails compared to setting the state to finished, and its overhead is small. The idea to add TaskContext.setStateParams is as commented at: https://github.com/treasure-data/digdag/pull/177#discussion_r71963995
But here doesn't use it. Making the task entry on the queue finished and retrying it again involves large overhead.
Without this call, this agent immediately calls back to digdag-server to mark the task finished. Once the task is marked as finished successfully, the task won't be re-executed. The risk of failure of the callback is same with the risk of failure of `ofNextPolling` callback.

digdag-standards/src/test/java/io/digdag/standards/operator/jdbc/JdbcOpTestHelper.java
```diff
@@ -0,0 +1,91 @@
...
+    public Path workpath()
+            throws IOException
+    {
+        return Files.createTempDirectory("jdbc_op_test");
+    }
+
+    public Config createConfig(Map<String, Object> configInput)
+            throws IOException
+    {
+        return loader.loadString(mapper.writeValueAsString(configInput)).toConfig(configFactory);
```
frsyuki: you can use this code to create Config instance:

```
import static io.digdag.client.DigdagClient.objectMapper;
...
return new ConfigFactory(objectMapper()).create()
                    set("host", "foobar1.org").
                    set("port", 6543).
                    set("user", "user1");
```

digdag-standards/src/main/java/io/digdag/standards/operator/pg/PgConnection.java
komamitsu: 👍 
I understand that the retry counter in Digdag will be incremented if the original exception is thrown as is. So we need to take care of that error.

PR #192, opened 2016-08-04
Log endpoint in client commands in debug level

frsyuki:
When there is a network issue, http client throws exceptions with message such as `Connection refused`. But the message doesn't include endpoint especially if the new user is unsure that configuration is correctly done or not.
This change add logging of endpoint so that the user can be confident that at least endpoint configuration is correct.


danielnorberg:
👍

PR #193, opened 2016-08-04
Disable HTTP contents logging that may include security information

frsyuki:
Apache httpclient logs HTTP headers and contents in DEBUG level by
default. But headers and contents may include security information. It
makes it difficult for users to use debug-level logging (`-l debug`).
Hesitation of using debug level logging makes debugging hard.
Debug logging should is useful as like #192.


danielnorberg:
👍

PR #195, opened 2016-08-05
td: add proxy.use_ssl option

frsyuki:


danielnorberg:
👍

frsyuki:
@danielnorberg I suspect that we should enable SSL proxy by default if protocol is SSL...how does http_proxy env var works for most of systems in the world? I think we should follow the same behavior.

PR #198, opened 2016-08-08
For code review of `pg_op` branche

komamitsu:


digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/LockConflictException.java
```diff
@@ -0,0 +1,28 @@
+package io.digdag.standards.operator.jdbc;
+
+import java.sql.SQLException;
+
+public class LockConflictException
+        extends RuntimeException
```
frsyuki: I think checked-exception (extending Exception instead of RuntimeException) works better for this use case because this exception must be caught and must not be re-thrown.

digdag-tests/src/test/java/acceptance/PgIT.java
```diff
@@ -0,0 +1,273 @@
...
+    {
+        return folder.getRoot().toPath().toAbsolutePath();
+    }
+
+    @Before
+    public void setUp()
+    {
+        if (isNullOrEmpty(POSTGRESQL)) {
+            throw new IllegalStateException("Environment variable `DIGDAG_TEST_POSTGRESQL` isn't set");
+        }
```
frsyuki: here should use `assumeThat(POSTGRESQL, not(isEmptyOrNullString()))` so that this test class itself is skipped if DIGDAG_TEST_POSTGRESQL is not set. DIGDAG_TEST_POSTGRESQL is not set when we want to test H2 database connectivity.

digdag-tests/src/test/java/acceptance/PgIT.java
```diff
@@ -0,0 +1,273 @@
...
+        setupRestrictedUser(props, database);
+
+        setupSourceTable(props, database);
+    }
+
+    @After
+    public void tearDown()
+    {
+        removeTempDatabase(props, database);
+    }
```
frsyuki: here should delete users as well as databases because users are not associated to a database in PostgreSQL.

frsyuki:
looks good other than above comments 👍

digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/LockConflictException.java
komamitsu: Oops, good catch!

Actually, I was planning to make it extend Exception and added definition `throws LockConfilictException`... https://github.com/treasure-data/digdag/pull/198/files/d798bc0b3283477b5f3179c0c07b804eac59822f#diff-6796cda912b3d95467f3864be5e93189R120

PR #201, opened 2016-08-09
refactor td operators + more non-blocking

danielnorberg:


danielnorberg:
@komamitsu PTAL

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -7,23 +7,32 @@
...
-import java.util.Date;
+import java.util.UUID;
 
+import static com.treasuredata.client.model.TDJob.Status.SUCCESS;
 import static io.digdag.util.RetryExecutor.retryExecutor;
 
 public class TDOperator
         implements Closeable
 {
+    private static final String JOB_ID = "job_id";
```
komamitsu: The original key names and https://github.com/treasure-data/digdag/pull/201/files#diff-3ade81f31ae01cbc057f3bc7abbc6b21R14 are lower camel cases. We'd better use an unified naming convention. Also, this change could cause an issue if remaining tasks stores the original keys and then tries to use them after the deployment.

komamitsu: Oh, you took care of the backward compatibility, nice!
https://github.com/treasure-data/digdag/pull/201/files#diff-9d50bfb9c6abc38992759addd61c9eb0R237

danielnorberg: Yeah, I think I'll revert the key name change.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -213,39 +221,75 @@ public TDJobOperator submitNewJobWithRetry(Submitter submitter)
...
-        return newJobOperator(client.partialDelete(database, table, from.getEpochSecond(), to.getEpochSecond()).getJobId());
-    }
+        ///////////////////////////////////////////////////////////////////////////////////////////
+        // TODO: remove this migration code
+        if (state.has("jobId")) {
+            Config jobState = state.getNestedOrSetEmpty(key);
+            if (!jobState.isEmpty()) {
+                throw new AssertionError();
+            }
+            jobState.setOptional("jobId", state.getOptional("jobId", String.class));
```
komamitsu: [minor] These codes are a bit confusing because it depends on the implementation that `state.getNestedOrSetEmpty()` returns a reference not a copy.

digdag-standards/src/main/java/io/digdag/standards/operator/td/BaseTdJobOperator.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    @Override
+    public final TaskResult runTask()
+    {
+        try (TDOperator op = TDOperator.fromConfig(params)) {
+
+            Optional<String> doneJobId = state.getOptional(DONE_JOB_ID, String.class);
+            TDJobOperator job;
+            if (!doneJobId.isPresent()) {
```
komamitsu: This check is only for avoiding calling `TDJobOperator#checkStatus` in case the previous attempt failed during processing result?

danielnorberg: Some operators run jobs in `processJobResult` and we need to keep calling `processJobResult` until it finishes "normally" without a polling exception. The above check is to prevent that the "main" job is then re-run.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
danielnorberg: Yeah, `getNestedOrSetEmpty` can be confusing. But if we keep it then I'm going to use it =)

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
```diff
@@ -81,42 +78,18 @@ private TdWaitOperator(Path workspacePath, TaskRequest request)
...
-                    throw TaskExecutionException.ofNextPolling(JOB_STATUS_API_POLL_INTERVAL, ConfigElement.copyOf(state));
-                }
-
-                // Fail the task if the job failed
-                if (status != SUCCESS) {
-                    String message = jobInfo.getCmdOut() + "\n" + jobInfo.getStdErr();
-                    throw new TaskExecutionException(message, ConfigElement.empty());
-                }
+                TDJobOperator job = op.runJob(state, POLL_JOB, this::startJob);
+                state.remove(POLL_JOB);
```
komamitsu: Why do you remove this state? Isn't it safer to left it so that `runJob()` can avoid re-run even when an unexpected retry occurs?

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
```diff
@@ -15,21 +14,20 @@
...
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 public class TdWaitOperatorFactory
         extends AbstractWaitOperatorFactory
         implements OperatorFactory
 {
+    public static final String POLL_JOB = "pollJob";
     private static Logger logger = LoggerFactory.getLogger(TdWaitOperatorFactory.class);
 
     private static final String JOB_ID = "jobId";
```
komamitsu: This constant is still needed?

komamitsu:
👍  I left some comments, though

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdWaitOperatorFactory.java
danielnorberg: This job must be re-run. It's polling for the query result to fulfill some criteria.

komamitsu: Ah, got it. 👍

PR #205, opened 2016-08-10
Run acceptance test if -Dtest.single=PATTERN is set explicitly

frsyuki:
Running a single acceptance test locally is useful to add an acceptance
test.

### DIFF
--- a/digdag-tests/build.gradle
+++ b/digdag-tests/build.gradle
@@ -18,9 +18,9 @@ test {
 
     // Exclude all acceptance tests if this is not an acceptance test run
     if (System.env.CI_ACCEPTANCE_TEST == null || !Boolean.valueOf(System.env.CI_ACCEPTANCE_TEST)) {
-        // run tests if -Dtest.single is explicitly set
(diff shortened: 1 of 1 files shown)


frsyuki:
@danielnorberg  PTAL

danielnorberg:
Isn't it odd to only do this for `-Dtest.single` and not for e.g. `--tests <pattern>` etc? How about just doing `CI_ACCEPTANCE_TEST=true` or checking for some other system property like `-Dtest.acceptance` ?

frsyuki:
I just didn't know about `--tests` option...apparently, -Dtest.single is deprecated and --tests is its successor: https://docs.gradle.org/current/userguide/java_plugin.html#N15A81
I added some more code to respect `--tests`.

`--tests <pattern>` starts java processes before filtering tests out if `forkEvery = 1` is set. `-Dtest.single=<pattern>` filters test out and starts a process. So `--tests` is much slower.

danielnorberg:
👍

PR #207, opened 2016-08-10
http_proxy support

danielnorberg:
Adds support for `http_proxy`, `HTTP_PROXY`, `https_proxy` and `HTTPS_PROXY` env vars to both digdag cli and the `td` operators.


digdag-cli/src/main/java/io/digdag/cli/Main.java
```diff
@@ -45,12 +44,18 @@
...
     private final PrintStream out;
     private final PrintStream err;
+    private final Map<String, String> env;
+
+    public Main(io.digdag.core.Version version, PrintStream out, PrintStream err) {
+        this(version, out, err, System.getenv());
+    }
 
-    public Main(io.digdag.core.Version version, PrintStream out, PrintStream err)
+    public Main(io.digdag.core.Version version, PrintStream out, PrintStream err, Map<String, String> env)
```
komamitsu: You put `env` right before `out` and `err` for other cli classes. So putting `env` before `out` here in the same way would be better.

digdag-cli/src/main/java/io/digdag/cli/Main.java
```diff
@@ -81,40 +86,40 @@ public int cli(String... args)
...
+        jc.addCommand("session", new ShowSession(version, env, out, err), "sessions");
+        jc.addCommand("attempts", new ShowAttempts(version, env, out, err));
+        jc.addCommand("attempt", new ShowAttempt(version, env, out, err));
+        jc.addCommand("reschedule", new Reschedule(version, env, out, err));
+        jc.addCommand("backfill", new Backfill(version, env, out, err));
+        jc.addCommand("log", new ShowLog(version, env, out, err), "logs");
+        jc.addCommand("kill", new Kill(version, env, out, err));
+        jc.addCommand("task", new ShowTask(version, env, out, err), "tasks");
+        jc.addCommand("schedule", new ShowSchedule(version, env, out, err), "schedules");
+        jc.addCommand("delete", new Delete(env, version, out, err));
```
komamitsu: Why don't you put `env` after `version` like others?

digdag-cli/src/main/java/io/digdag/cli/Run.java
```diff
@@ -216,9 +216,13 @@ public void run(String workflowNameArg, String matchPattern)
     {
         Properties systemProps = loadSystemProperties();
 
+        systemProps.setProperty("environment", DigdagClient.objectMapper().writeValueAsString(env));
```
komamitsu: Is this needed? For future use?

digdag-client/src/main/java/io/digdag/client/config/Config.java
```diff
@@ -270,7 +270,7 @@ public Object get(String key, JavaType type, Object defaultValue)
     @SuppressWarnings("unchecked")
     public <E> E get(String key, TypeReference<E> type, E defaultValue)
     {
-        return (E) get(key, mapper.getTypeFactory().constructType(type));
+        return (E) get(key, mapper.getTypeFactory().constructType(type), defaultValue);
```
komamitsu: 👍

digdag-cli/src/main/java/io/digdag/cli/Main.java
danielnorberg: Will do! 👍

digdag-cli/src/main/java/io/digdag/cli/Main.java
danielnorberg: Will do! 👍

digdag-cli/src/main/java/io/digdag/cli/Run.java
danielnorberg: Forgot to remove that. Good catch!

digdag-standards/src/test/java/io/digdag/standards/operator/td/TDOperatorTest.java
```diff
@@ -30,6 +31,7 @@
 @RunWith(MockitoJUnitRunner.class)
 public class TDOperatorTest
 {
+    public static final ImmutableMap<String, String> EMPTY_ENV = ImmutableMap.of();
```
danielnorberg: Good catch! 👍

digdag-tests/src/test/java/acceptance/td/TdIT.java
```diff
@@ -114,6 +135,108 @@ public void testRunQuery()
     }
 
     @Test
+    public void testRunQueryWithEnvProxy()
+            throws Exception
+    {
+        List<FullHttpRequest> jobIssueRequests = Collections.synchronizedList(new ArrayList<>());
+
+        proxyServer = DefaultHttpProxyServer
```
komamitsu: How about moving these codes to a new private method that returns HttpProxyServerBootstrap and calling it from this method and testRunQueryOnServerWithEnvProxy() to reduce duplicated codes?

komamitsu:
I left some minor comments. Other than that look good to me! 👍

PR #210, opened 2016-08-13
Add further explanations for sh operator

kiyoto:
I believe many "first timers" prefer to run a shell command inline (as opposed to a separate file). Showing an example.

### DIFF
--- a/digdag-docs/src/operators.rst
+++ b/digdag-docs/src/operators.rst
@@ -107,7 +107,15 @@ sh>: Shell scripts
 
 **sh>:** operator runs a shell script.
 
-TODO: add more description here
+Run a shell command (`/bin/sh`)
+
+.. code-block:: yaml
+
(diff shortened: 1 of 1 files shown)


danielnorberg:
👍

digdag-docs/src/operators.rst
```diff
@@ -121,6 +129,14 @@ TODO: add more description here
 
   * :command:`sh>: tasks/workflow.sh --task1`
 
+The shell defaults to `/bin/sh`. If an alternate shell such as `zsh` is desired, use the `shell` option in the `_export` section.
+
+.. code-block:: yaml
+
+    _export:
+      sh:
+        shell: /usr/bin/zsh
```
frsyuki: `shell:` takes an array like `shell: [/usr/bin/zsh]`

PR #216, opened 2016-08-16
Handle TaskLimitExceededException to avoid returning status code 500

komamitsu:


digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
```diff
@@ -167,6 +167,9 @@ public Response startAttempt(RestSessionAttemptRequest request)
             RestSessionAttempt res = RestModels.attempt(attempt, def.getProject().getName());
             return Response.ok(res).build();
         }
+        catch (LimitExceededException ex) {
+            return Response.status(Response.Status.FORBIDDEN).entity(ex.getMessage()).build();
```
danielnorberg: `403 Forbidden` is typically used e.g. when the user does not have sufficient permissions. I believe `400 Bad Request` would be more suitable.

danielnorberg:
nice! 👍

PR #218, opened 2016-08-16
Add session_attempts.finished_at

frsyuki:
The main purpose is making it possible to know how much time attempts are taking. This sort of metrics is useful to know how users are interacting with digdag.
It needed to parse JSON stored in task_archives table to know finish time.
Feature requested by @diana-shealy.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
@@ -733,6 +733,27 @@ public class DatabaseMigrator
         }
     };
 
+    private final Migration MigrateAddFinishedAtToSessionAttempts = new Migration() {
+        @Override
+        public String getVersion()
+        {
+            return "20160816021018";
+        }
+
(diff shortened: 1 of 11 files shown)


frsyuki:
@danielnorberg How do you think about naming?

* finished_at
* completed_at
* end_at
* done_at

As far as I looked up dictionary, finished_at makes sense because it's set even when the attempt fails.

danielnorberg:
`finished_at` sgtm. 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
```diff
@@ -756,7 +756,27 @@ public void migrate(Handle handle)
                             .addTimestamp("updated_at", "not null")
                             .build());
 
-            handle.update("create index secrets_on_site_id_and_project_id_and_scope_and_key on secrets (site_id, project_id, scope, key)");
```
danielnorberg: Ouch.

PR #221, opened 2016-08-17
Add server.jmx.port option to start JMX agent

frsyuki:
Main goal is to get GC status using JMX. It's also possible to export
custom metrics by adding `@Managed` to an injected class and
exposing the class using `ExportBinder.newExporter` method.

### DIFF
--- a/digdag-server/src/main/java/io/digdag/server/ServerConfig.java
+++ b/digdag-server/src/main/java/io/digdag/server/ServerConfig.java
@@ -33,6 +33,8 @@ public interface ServerConfig
 
     public Optional<String> getAccessLogPath();
 
+    public Optional<Integer> getJmxPort();
+
     public String getAccessLogPattern();
 
     public boolean getExecutorEnabled();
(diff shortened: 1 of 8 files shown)


frsyuki:
@komamitsu PTAL

digdag-server/src/main/java/io/digdag/server/ServerConfig.java
```diff
@@ -33,6 +33,8 @@
 
     public Optional<String> getAccessLogPath();
 
+    public Optional<Integer> getJmxPort();
```
komamitsu: [minor] `public` isn't needed here.

digdag-tests/src/test/java/acceptance/ServerJmxIT.java
```diff
@@ -0,0 +1,49 @@
...
+import javax.management.MBeanServerConnection;
+import javax.management.ObjectName;
+
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.junit.Assert.assertThat;
+
+public class ServerJmxIT
+{
+    public static final Pattern JMX_PORT_PATTERN = Pattern.compile("\\s*JMX agent started on port (\\d+)\\s*");
```
komamitsu: [trivial] We can set it to`private`

komamitsu:
👍

PR #226, opened 2016-08-22
rebrandable cli + refactoring

danielnorberg:
* Add `-Dio.digdag.cli.programName=<name>` system property for setting the name that the cli should use in help/usage text.

* Make it possible to specify `-c/--config` before the command on the command line. E.g. `digdag -c /custom/config workflows`.

* Wire the cli commands using Guice to eliminate a lot of boilerplate.


danielnorberg:
@frsyuki PTAL

frsyuki:
Looks very good 👍

PR #229, opened 2016-08-22
fix td> store_last_result + add test

danielnorberg:
* fix broken storeParams decoration in `BaseTdJobOperator`
* Store all columns in first row, not just one column.


frsyuki:
oh i see....looks good 👍

PR #231, opened 2016-08-23
use prebuilt ci build image

danielnorberg:


frsyuki:
nice improvements 👍

circle.yml
```diff
@@ -4,6 +4,7 @@ machine:
     version: oraclejdk8
   environment:
     TERM: dumb  # don't let gradle use fancy ansi seq code
+    BUILD_IMAGE: danieln/digdag-build:20160823085939Z-4c3c94d6925e4e476fca04260ec6be6352351de2
```
frsyuki: this image is built using `docker/push.sh` command, right? it would be great if here has a comment that says about it.

danielnorberg: Yeah, and it should be `digdag/build` instead of `danieln/digdag-build`. Created a digdag docker hub account we can use for that.

danielnorberg:
We cannot use oracle jdk for this build image. We'd have to use openjdk (e.g. zulu). Do we still want to go ahead with this?

PR #233, opened 2016-08-24
UI: Added workflow duration fields

dzannotti:
Added duration column per session in /workflows/:id
Added duration column per attempt in /sessions/:id
Added Last attempt duration field in brief (top view) in /sessions/:id


digdag-ui/console.jsx
```diff
@@ -13,6 +13,7 @@ import moment from 'moment';
 import pako from 'pako';
 import path from 'path';
 import yaml from 'js-yaml';
+import Duration from 'duration';
```
dzannotti: not too fond of this dep, but rather this than reinvent the wheel

danielnorberg:
👍 LGTM!

danielnorberg:
should we merge and deploy?

dzannotti:
Yep, let's hold off on the release as i'm about to add another PR

PR #235, opened 2016-08-26
Added job link to tasks table

dzannotti:
This adds a job link to the tasks table (for now only td jobs id from storeParams are considered linkable)

### DIFF
--- a/digdag-ui/console.jsx
+++ b/digdag-ui/console.jsx
@@ -55,6 +55,7 @@ type AuthItem = {
 
 type ConsoleConfig = {
   url: string;
+  jobUrl: (id:string) => string;
   navbar: ?{
     brand: ?string;
     logo: ?string;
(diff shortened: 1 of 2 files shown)


digdag-ui/console.jsx
```diff
@@ -815,12 +816,23 @@ function formatTaskState(state) {
   }
 }
 
+const JobLink = ({ params }:{params: Object}) => {
```
danielnorberg: The job id in the task `storeParams` is only present _after_ the task has finished. I believe we also want to link to any currently running td job, while the task is running. To do that we can inspect the task `stateParams` which will look like this while it's running a td job:

```yaml
job:
  domainKey: 6d649edd-05b8-45f4-8ffb-aebc6883576f
  jobId: '85201301'
  pollIteration: 2
```

danielnorberg:
👍

PR #234, opened 2016-08-26
load td.conf

danielnorberg:
* Add digdag client configuration plugin interface and an implementation that loads apikey from `~/.td/td.conf` and sets endpoint etc. This is disabled by default and can be enabled via `-Dio.digdag.standards.td.client-configurator-enabled=true`.

* Make the local secret store manager load secrets from a set of backing secret stores and add a secret store implementation that provides secrets td client configuration secrets from `~/.td/td.conf`

* Allow plugins to configure their injector, similar to a guice module.

PR #237, opened 2016-08-26
Load system plugins from classpath

frsyuki:
If io.digdag.spi.Plugin are included in classpath when digdag-cli
starts, load them as system plugins. This is useful to customize digdag
using plugin mechanism. io.digdag.spi.Extension can do similar thing but
Extension is to override some built-in modules of digdag-core, while
io.digdag.spi.Plugin is used to add new module to exntension points.
They are used for different purposes.

This is a pull-req for #234.


danielnorberg:
SGTM. Not sure why this change made tests start failing. Will take a look.

PR #238, opened 2016-08-27
Workflow name calculated from file name must normalize path separator

frsyuki:
Workflow files in subdirectories will have
`+<dir_name>/<sub_dir_name>/<file_name>` as its workflow name. On
Windows, because path separator is '\', it will be
`+<dir_name>\<sub_dir_name>\<file_name>`. This is invalid because
workflow name must not contain backslash (validated using
ModelValidator.checkWorkflowName).


danielnorberg:
Nice. Please add a test as well.

frsyuki:
Added some unit tests.
I found another bug where ProjectArchiveLoader fails if there is a directory named *.dig. It's also fixed.
@danielnorberg please also check #248

frsyuki:
This assumes #239 (gradle 3.0) because pom file generation (called at `ci/validate.sh`) had a bug.

PR #240, opened 2016-08-29
optimize startup time when command line is empty

danielnorberg:
It's common to run a command line without any arguments to just get a quick usage help. This cuts down the time for that to ~200 ms on this laptop.


frsyuki:
👍

PR #243, opened 2016-08-29
Updated installation for Windows

frsyuki:
* `PowerShell -Command ...` runs a PowerShell command.
* Invoke-WebRequest command downloads a file.
* setx sets an environment variable.

Environment variable will be saved in current user's registry (`HKEY_CURRENT_USER\Environment`).


danielnorberg:
👍

PR #246, opened 2016-08-30
move secret scope precedence to provider

danielnorberg:
As the SecretProvider is the interface that is used to fetch secrets in a workflow scope, it makes more sense to make the precedence decision there and simplify SecretStore.

Also:

* Add secrets table index (with unique constraint)
* Add database secret store test


frsyuki:
Looks good 👍 
SecretAccessContext.java's Optional seems unnecessary.

danielnorberg:
Yup, removed that.

PR #252, opened 2016-08-31
docs: use '>' always even for operators that don't use _command

frsyuki:


digdag-docs/src/operators.rst
```diff
@@ -563,7 +563,7 @@ TODO: add more description here
         apikey: YOUR/API_KEY
 
     +step1:
-      _type: td_table_export
+      td_table_export>:
       database: mydb
       table: mytable
```
danielnorberg: Maybe we should make `td_table_export>` take `database.table` as a command?

digdag-docs/src/operators.rst
```diff
@@ -758,7 +758,7 @@ To use Gmail SMTP server, you need to do either of:
       to: [me@example.com]
 
     +step2:
-      _type: mail
+      mail>:
       body: this is email body in string
       subject: workflow started
       to: [me@example.com]
```
danielnorberg: How about making the `mail>` operator take `to:` as a command?

danielnorberg:
👍

digdag-docs/src/operators.rst
frsyuki: Should it be `subject`, `body`, or to...
I think current implementation is ok because basically operators take file path at `>: ...`.

Issue #70, opened 2016-04-21
project structure and workflow suffix changes


danielnorberg:
- no `digdag.yml` project file 
- workflow file suffixes change from `.yml` to `.dig`
- all `.dig` files in project directory are picked up automatically by digdag
- digdag cli gains a `--project` option to specify the project directory
- `digdag run` `-f` option is removed
- `digdag init` creates: 
  - `.digdag` directory in project directory
  - `.digdag/config` file
- digdag cli looks for `.digdag` directory, starting in current directory and then recurses into parent directories (like git).
- `digdag push` pushes entire project (as identified by `.digdag`) in some parent dir, even if executed in a subdir.
- filenames referenced by a `.dig` file are resolved relative to the path of the `.dig` file so that given a project with:

```
.digdag/config
subdir/workflow.dig
queries/foo.sql
```

Then a user can both be in the project dir and do:

```
digdag run subdir/workflow
```

Or

```
cd subdir/
digdag run workflow
```

with the same result. The `subdir/workflow.dig` file must reference the query file as `../queries/foo.sql`.
- digdag does not change "CWD" when executing a workflow.
- `digdag run` `<workflow name>` parameter can include `.dig` suffix so that the below invocation semantics are identical:
  - `digdag run foo`
  - `digdag run foo.dig`

yoyama:
Close the old issue. Please reopen or file new one if you need.

PR #254, opened 2016-09-01
Update call> operator to read workflow files in a subdirectory

frsyuki:
Dividing a large workflow definition into multiple files is helpful to
organize a complex workflow definition. `call>` is more straightforward
(and less hacky) compared to `!include` in terms of syntax.

This changes `call>` operator to be able to load workflow definition from a file so that it can call workflows in a subdirectory. *.dig files in subdirectories won't be recognized as workflows.

This change includes removal of `params` argument from
`ConfigLoaderManager.loadParameterizedFile` method. `params` argument was necessary when it was using jinja2 to preprocess a file. It's unnecessary any more since digdag evaluates parameters when a workflow runs.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/Arguments.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/Arguments.java
@@ -47,7 +47,7 @@ public class Arguments
 
         // -P files
         if (paramsFile != null) {
-            overwriteParams.merge(loader.loadParameterizedFile(new File(paramsFile), cf.create()));
+            overwriteParams.merge(loader.loadParameterizedFile(new File(paramsFile)));
         }
 
         // -p options
--- a/digdag-cli/src/main/java/io/digdag/cli/Show.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/Show.java
@@ -59,11 +59,10 @@ public class Show
     private void show(Injector injector, String workflowPath)
             throws Exception
     {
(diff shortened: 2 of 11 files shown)


frsyuki:
A discussion is whether `_workdir` should be set or not.

For example, with following files, root.dig uses `call>` to call sub.dig, and sub.dig uses query.sql and nested.dig files:

```
root.dig
sub/sub.dig
sub/queries/query.sql
sub/nested/nested.dig
```

If `call>` operator doesn't set `_workdir` (current implementation), sub.dig will use relative path names like this:

```
td>: queries/query.sql
call>: nested/nested.sql
```

If `call>` operator sets _workdir to `sub/`, sub.dig will use absolute path names like this:

```
td>: sub/queries/query.sql
call>: sub/nested/nested.sql
```

frsyuki:
Setting `_workdir` lets users run a subworkflow individually. hm...

danielnorberg:
I believe a workflow file should refer to files relative to the location of the workflow file.

I.e. in the above example, nested.dig should refer to query.sql using the path `../queries/query.sql`.

When resolving the file path, resolve it in the directory of the workflow file. This way nested workflow files can be called by other workflows in other directories using `call>` as well as run directly by `digdag run`.

`digdag run` itself should look for the `.digdag` directory in cwd (or a parent directory) according to https://github.com/treasure-data/digdag/issues/70 and use the directory containing `.digdag` as the project root. File references should not be allowed to go outside the project directory.

frsyuki:
Updated implementation is #290.

PR #256, opened 2016-09-01
Fix race condition around SessionAttemptControl.archiveTasks

frsyuki:
`WorkflowExecutor.propagateSessionArchive` lists up finished root tasks
using one transaction. Then, use another transaction to archive the
attempts of the root tasks. This is intentional to avoid long lock
because archiving tasks is expected to take long time especially if the
attempt has many tasks.

However, because they use two transactions, another thread may archive
an attempt between the two transactions. Therefore, the later
transaction needs to re-check state of the attempt before actually
archiving it.

Without this fix, a thread may try to archive an empty attempt because the other thread archives (=deletes) all tasks. And because task_archives table has unique index on attempt id, the transaction fails with `duplicate key value violates unique constraint "task_archives_pkey"` message. So, this problem doesn't cause inconsistent data.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
+++ b/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
@@ -621,7 +621,7 @@ public class WorkflowExecutor
                 .map(task -> {
                     return sm.lockAttemptIfExists(task.getAttemptId(), (store, summary) -> {
                         if (summary.getStateFlags().isDone()) {
-                            // already archived. This means that another thread archvied
+                            // already archived. This means that another thread archived
                             // this attempt after findRootTasksByStates call.
                             return false;
                         }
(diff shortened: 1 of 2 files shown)


frsyuki:
@komamitsu PTAL

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
```diff
@@ -620,9 +620,16 @@ private boolean propagateSessionArchive()
                 .stream()
                 .map(task -> {
                     return sm.lockAttemptIfExists(task.getAttemptId(), (store, summary) -> {
-                        SessionAttemptControl control = new SessionAttemptControl(store, task.getAttemptId());
-                        control.archiveTasks(archiveMapper, task.getState() == TaskStateCode.SUCCESS);
-                        return true;
+                        if (summary.getStateFlags().isDone()) {
+                            // already archived. This means that another thread archvied
```
komamitsu: `archived`

komamitsu:
👍

komamitsu:
Adding an unit test for this issue would be nice when you get a chance.

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
frsyuki: 👍

frsyuki:
Thank you for your review!
Please go ahead your work that is blocked by this issue. I'm going to add tests later.

PR #257, opened 2016-09-01
Retry workflow executor loop on unexpected error

frsyuki:
Thread created at `ServerBootstrap.initialize` was calling
`control.destroy` when WorkflowExecutor throws an unexpected exception.
But this is bad because `DeploymentManager.undeploy` is called and thus
HTTP handlers will start returning errors (`500 Internal Server Error`)
to all requests with `Deployment digdag-server.war has stopped` error
logs until the server restarts.

This changes the behavior to not undeploy DeploymentManager. Instead,
workflow executor loop is retried for ever.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/LocalAgentManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/LocalAgentManager.java
@@ -15,8 +15,8 @@ public class LocalAgentManager
         implements BackgroundExecutor
 {
     private final Supplier<MultiThreadAgent> agentFactory;
-    private Thread thread;
-    private MultiThreadAgent agent;
+    private volatile Thread thread;
+    private volatile MultiThreadAgent agent;
 
     @Inject
     public LocalAgentManager(
--- a/digdag-core/src/main/java/io/digdag/core/agent/MultiThreadAgent.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/MultiThreadAgent.java
@@ -102,7 +102,14 @@ public class MultiThreadAgent
                 }
             }
             catch (Throwable t) {
-                logger.error("Uncaught exception. Ignoring.", t);
+                logger.error("Uncaught exception during acquiring tasks from a server. Ignoring. Agent thread will be retried.", t);
+                try {
+                    // sleep before retrying
+                    Thread.sleep(1000);
+                }
+                catch (InterruptedException ex) {
+                    Thread.currentThread().interrupt();
+                }
             }
         }
     }
--- a/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
+++ b/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
@@ -342,6 +342,18 @@ public class WorkflowExecutor
         }
     }
 
+    public void noticeRunWhileConditionChange()
+    {
+        propagatorLock.lock();
+        try {
+            // don't set propagatorNotice but break wait at runWhile
+            propagatorCondition.signalAll();
+        }
+        finally {
+            propagatorLock.unlock();
+        }
+    }
+
     public void run()
             throws InterruptedException
     {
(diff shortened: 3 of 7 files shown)


digdag-core/src/main/java/io/digdag/core/agent/LocalAgentManager.java
```diff
@@ -15,8 +15,8 @@
         implements BackgroundExecutor
 {
     private final Supplier<MultiThreadAgent> agentFactory;
-    private Thread thread;
-    private MultiThreadAgent agent;
+    private volatile Thread thread;
```
komamitsu: [minor] This keyword doesn't seem to be needed because this value is called only in `synchronized` method.

digdag-core/src/main/java/io/digdag/core/agent/MultiThreadAgent.java
```diff
@@ -102,7 +102,14 @@ public void run()
...
             }
             catch (Throwable t) {
-                logger.error("Uncaught exception. Ignoring.", t);
+                logger.error("Uncaught exception during acquiring tasks from a server. Ignoring. Agent thread will be retried.", t);
+                try {
+                    // sleep before retrying
+                    Thread.sleep(1000);
+                }
+                catch (InterruptedException ex) {
+                    Thread.currentThread().interrupt();
```
komamitsu: 👍

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
```diff
@@ -342,6 +342,18 @@ private void noticeStatusPropagate()
         }
     }
 
+    public void noticeRunWhileConditionChange()
```
komamitsu: This method can lead to exponential backoff style waiting in runWhile(), right?

digdag-server/src/main/java/io/digdag/server/WorkflowExecutorLoop.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public void eagerShutdown()
+            throws InterruptedException
+    {
+        startShutdown();
+    }
+
+    private void startShutdown()
+    {
+        if (stop == false) {
```
komamitsu: [trivial] You can write `!stop` instead here

digdag-server/src/main/java/io/digdag/server/WorkflowExecutorLoop.java
```diff
@@ -0,0 +1,113 @@
...
+
+public class WorkflowExecutorLoop
+        implements BackgroundExecutor
+{
+    private static final Logger logger = LoggerFactory.getLogger(WorkflowExecutorLoop.class);
+
+    private final Supplier<Thread> threadFactory;
+    private final WorkflowExecutor workflowExecutor;
+
+    private volatile Thread thread = null;
```
komamitsu: [minor] `volatile` can be removed. But it's okay to leave it just in case

digdag-server/src/main/java/io/digdag/server/ServerBootstrap.java
```diff
@@ -106,23 +108,6 @@ public Injector initialize(ServletContext context)
...
-            // TODO create global site
-            LocalSite site = injector.getInstance(LocalSite.class);
-
-            Thread thread = new Thread(() -> {
-                try {
-                    site.run();
-                }
-                catch (Exception ex) {
-                    logger.error("Uncaught error", ex);
-                    control.destroy();
```
komamitsu: 👍

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
frsyuki: Yes. It is necessary to shorten shutdown duration that returns 503 Service Temporarily Unavailable errors to REST clients.
`runWhile` method sleeps MAX_INTERVAL (5 seconds) at most if there're no actively updating tasks. It uses `propagatorCondition.await(time)` to sleep. noticeRunWhileConditionChange breaks the sleep by sending notification to `propagatorCondition`.

komamitsu:
👍

digdag-core/src/main/java/io/digdag/core/agent/LocalAgentManager.java
frsyuki: hm...maybe I'm still not 100% understanding about volatile...

PR #258, opened 2016-09-01
Report uncaught errors through JMX 

frsyuki:
Background executor threads try to recover from uncaught errors as much
as possible. However, because those errors are unexpected, system
administrators should be notified. This change introduces new
ErrorReporter interface so that administrators can customize the behavior
when uncaught errors are thrown. Default implementation reports counts
using JMX.

#257 is an example that should be notified.

This change assumes #257. Changes since #257 is https://github.com/treasure-data/digdag/compare/retry-workflow-executor-loop...report-unexpected-errors

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/ErrorReporter.java
+++ b/digdag-core/src/main/java/io/digdag/core/ErrorReporter.java
@@ -1,0 +1,16 @@
+package io.digdag.core;
+
+public interface ErrorReporter
+{
+    void reportUncaughtError(Throwable error);
+
+    static ErrorReporter empty()
+    {
+        return new ErrorReporter()
+        {
+            @Override
+            public void reportUncaughtError(Throwable error)
+            { }
+        };
+    }
+}
--- a/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
+++ b/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
@@ -346,7 +346,7 @@ public class WorkflowExecutor
     {
         propagatorLock.lock();
         try {
-            // don't set propagatorNotice but break wait at runWhile
+            // don't set propagatorNotice but break wait in runWhile
             propagatorCondition.signalAll();
         }
         finally {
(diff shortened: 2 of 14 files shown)


frsyuki:
@komamitsu  PTAL

digdag-core/src/main/java/io/digdag/core/ErrorReporter.java
```diff
@@ -0,0 +1,16 @@
+package io.digdag.core;
+
+public interface ErrorReporter
```
danielnorberg: I guess we could use the `Thread.UncaughtExceptionHandler` interface.

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
```diff
@@ -342,6 +342,18 @@ private void noticeStatusPropagate()
         }
     }
 
+    public void noticeRunWhileConditionChange()
+    {
+        propagatorLock.lock();
+        try {
+            // don't set propagatorNotice but break wait at runWhile
```
danielnorberg: s/at/in

digdag-core/src/main/java/io/digdag/core/ErrorReporter.java
frsyuki: perhaps naming is not good but, that's because threads retry operations when uncaught exceptions happen. For example, `WorkflowExecutorLoop.run` may catch uncaught error when database transaction caused error (such as DNS error, statement timeout, etc.). Appropriate recovery behavior is another discussion but retrying the hole operation is an option. Retrying won't be an option with `Thread.UncaughtExceptionHandler` handler.

digdag-tests/src/test/java/acceptance/ServerJmxIT.java
```diff
@@ -46,4 +70,30 @@ public void scheduleStartTime()
...
+        try (JMXConnector con = connectJmx(server)) {
+            MBeanServerConnection beans = con.getMBeanServerConnection();
+
+            Object uncaughtErrorCount = beans.getAttribute(ObjectName.getInstance("io.digdag.core", "name", "ErrorReporter"), "UncaughtErrorCount");
+            assertThat(uncaughtErrorCount, is(0));
+
+            // oops, tasks table is broken!?
+            try (DataSourceProvider dsp = new DataSourceProvider(server.getRemoteTestDatabaseConfig())) {
+                Statement stmt = dsp.get().getConnection().createStatement();
+                stmt.execute("drop table tasks cascade");
```
komamitsu: 👍

komamitsu:
👍 
BTW, how about adding AirbrakeErrorReporter or something later?

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
frsyuki: 👍 I asked people around me to know how in and at are different...and still not 100% sure about the difference....

frsyuki:
@komamitsu sounds good idea. maybe sentry.

PR #259, opened 2016-09-02
Schema migrator should lock schema before migration

frsyuki:
This change addresses 3 issues to make schema migration less awful.

* Schema migrator should not run concurrently on multiple threads.
  Migrator acquires an exclusive table lock to avoid it.

* Applying migration scripts should not skip migration scripts that are
  not applied yet but older than the latest applied version. This is
  because a pull-request that has larger version of migration script
  could be merged earlier than another pull-request. Even with this
  case, because obviously there're no dependency between those two
  pull-requests (this is an assumption), all of the migration scripts
  should be applied.

* Migration definition files should be separated into files so that
  commit to DatabaseMigrator.java file doesn't cause many conflicts
  that are not obvious to merge.

Other problems such as "server can't start until migration finishes" are out of scope for now.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseConfig.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseConfig.java
@@ -243,4 +243,9 @@ public interface DatabaseConfig
 
         return props;
     }
+
+    static boolean isPostgres(String databaseType)
+    {
+        return databaseType.equals("postgresql");
+    }
 }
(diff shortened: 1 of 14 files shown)


frsyuki:
@danielnorberg can you comment about change? This is a short-term solution so that we can run at least 2 ~ 3 servers.

danielnorberg:
LGTM 👍

frsyuki:
👍

PR #263, opened 2016-09-08
Test with PostgreSQL should respect all parameters

frsyuki:
ci/run_test_pg.sh is setting idleTimeout and minimumPoolSize parameters
but they were not used.


danielnorberg:
👍

PR #264, opened 2016-09-08
Nested transaction must not open a new connection to avoid deadlock

frsyuki:
Deadlock happens in following scenario:

* Thread1 starts a transaction and locks a row (such as lockTaskIfExists).
* Thread2 opens a connection and tries to lock the same row. This is
  blocked because thread1 already locks the row.
* Thread1 tries to open another connection in the transaction. This
  could be blocked for ever by Thread2 if number of opened connection
  reached maximumPoolSize when thread2 opened a connection.

`WorkflowExecutor.getTaskRequest` behaves like thread1 (`sm.getAttemptWithSessionById` in `lockTaskIfExists`).
`WorkflowExecutor.enqueueTask` also behaves like thread1 (`DatabaseTaskQueueServer.enqueue` in `lockTaskIfExists`).

WorkflowExecutorParallelStressTest reproduces this issue. It reproduces as a timeout error (`Connection is not available, request timed out after 30000ms`)
This issue doesn't happen h2 database because digdag doesn't use connection pool with h2.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
@@ -288,7 +288,7 @@ public abstract class BasicDatabaseStoreManager <D>
                                     "from within the callback. See cause " +
                                     "for the original exception.", ex);
                         }
-                        handle.commit();
+                        validateTransactionAndCommit(handle);
                         return retval;
                     }
                     finally {
@@ -340,7 +340,7 @@ public abstract class BasicDatabaseStoreManager <D>
                                     "from within the callback. See cause " +
                                     "for the original exception.", ex);
                         }
-                        handle.commit();
+                        validateTransactionAndCommit(handle);
                         return retval;
                     }
                     finally {
@@ -431,13 +431,21 @@ public abstract class BasicDatabaseStoreManager <D>
     private void validateTransactionAndCommit(Handle handle)
         throws TransactionFailedException
     {
-        // Validate connection before COMMIT. This is necessary because PostgreSQL actually runs
-        // ROLLBACK silently when COMMIT is issued if a statement failed during the transaction.
-        // It means that BasicDatabaseStoreManager.transaction method doesn't throw exceptions
-        // but the actual transaction is rolled back. Here checks state of the transaction by
-        // calling org.postgresql.jdbc.PgConnection.isValid method.
+        // Validate the connection before COMMIT. Purpose of this method is to avoid
+        // following unexpected behavior of PostgreSQL:
         //
-        // Here assumes that HikariDB doesn't overwrite isValid.
+        // 1. A thread starts a transaction
+        // 2. The thread issues a statement and it fails (e.g. unique key violation)
+        // 3. The thread catch the exception and does other things
+        // 4. The thread issues COMMIT
+        // 5. PostgreSQL silently translate COMMIT to ROLLBACK without errors. The thread
+        //    assumes that the transaction is committed but actually everything is rolled back.
+        //
+        // To avoid this issue, the thread needs to send a dummy SELECT statement after step 4.
+        // PostgreSQL makes the statement failed ("ERROR:  25P02: current transaction is aborted,
+        // commands ignored until end of transaction block").
+        //
+        // Here assumes that HikariDB doesn't overwrite PgConnection.isValid method.
         boolean isValid;
         try {
             isValid = handle.getConnection().isValid(30);
--- a/digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
@@ -359,19 +359,19 @@ public abstract class BasicDatabaseStoreManager <D>
     }
 
     @SuppressWarnings("unchecked")
-    public <T, E1 extends Exception, E2 extends Exception> T transactionWithRetry(RetryableTransactionActionWithExceptions<T, D, E1, E2> action, Class<E1> exClass1, Class<E2> exClass2) throws E1, E2
+    public <T, E1 extends Exception, E2 extends Exception> T transactionWithLocalRetry(RetryableTransactionActionWithExceptions<T, D, E1, E2> action, Class<E1> exClass1, Class<E2> exClass2) throws E1, E2
     {
         // TODO
         // Here should throw error if currentTransaction.get() != null to avoid race condition under high load.
         // In following scenario, a deadlock happens:
-        // * thread1 acquires a row lock on RDBMS (such as lockTaskIfExists), then calls transactionWithRetry.
+        // * thread1 acquires a row lock on RDBMS (such as lockTaskIfExists), then calls transactionWithLocalRetry.
         // * thread2 opens a connection and try to lock the same row on RDBMS (this is blocked).
-        // * thread1's transactionWithRetry calls dbi.open to open another connection. But this could be blocked
+        // * thread1's transactionWithLocalRetry calls dbi.open to open another connection. But this could be blocked
         //   for ever if number of opened connection reached maximumPoolSize when thread2 opened a connection.
-        // However, ScheduleExecutor.backfill is calling putAndLockSession (that calls transactionWithRetry)
+        // However, ScheduleExecutor.backfill is calling putAndLockSession (that calls transactionWithLocalRetry)
         // within lockScheduleById transaction. There're no good solutions to avoid both issues for now.
         //
-        //checkState(currentTransaction.get() == null, "Nested retrying transaction is not allowed");
+        //checkState(currentTransaction.get() == null, "Nested transaction with local retry is not allowed");
         try {
             return transactionRetryExecutor.runInterruptible(() -> {
                 TransactionState ts = new TransactionState();
--- a/digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
@@ -251,6 +251,7 @@ public abstract class BasicDatabaseStoreManager <D>
                 throw Throwables.propagate(innerException);
             }
             catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
                 throw Throwables.propagate(ex);
             }
         }
@@ -301,6 +302,7 @@ public abstract class BasicDatabaseStoreManager <D>
                 throw Throwables.propagate(innerException);
             }
             catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
                 throw Throwables.propagate(ex);
             }
         }
@@ -353,6 +355,7 @@ public abstract class BasicDatabaseStoreManager <D>
                 throw Throwables.propagate(innerException);
             }
             catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
                 throw Throwables.propagate(ex);
             }
         }
@@ -420,6 +423,7 @@ public abstract class BasicDatabaseStoreManager <D>
             throw Throwables.propagate(innerException);
         }
         catch (InterruptedException ex) {
+            Thread.currentThread().interrupt();
             throw Throwables.propagate(ex);
         }
     }
@@ -472,6 +476,7 @@ public abstract class BasicDatabaseStoreManager <D>
                 throw Throwables.propagate(innerException);
             }
             catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
                 throw Throwables.propagate(ex);
             }
         }
@@ -498,6 +503,7 @@ public abstract class BasicDatabaseStoreManager <D>
                 throw Throwables.propagate(innerException);
             }
             catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
                 throw Throwables.propagate(ex);
             }
         }
@@ -525,6 +531,7 @@ public abstract class BasicDatabaseStoreManager <D>
                 throw Throwables.propagate(innerException);
             }
             catch (InterruptedException ex) {
+                Thread.currentThread().interrupt();
                 throw Throwables.propagate(ex);
             }
         }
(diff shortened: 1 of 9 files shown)


frsyuki:
@komamitsu Can I ask you to review?

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
```diff
@@ -224,45 +230,48 @@ public int getIndex()
...
+                        currentTransaction.set(null);
+                    }
+                });
+            }
+            catch (RetryGiveupException ex) {
+                Throwable innerException = ex.getCause();
+                throw Throwables.propagate(innerException);
+            }
+            catch (InterruptedException ex) {
+                throw Throwables.propagate(ex);
```
komamitsu: Restoring interrupted status would be better

komamitsu: If the interrupted state will be restored where this RuntimeException is caught, it's okay. If not, restoring the state here would be better.

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
```diff
@@ -274,42 +283,105 @@ public int getIndex()
...
+            }
         }
     }
 
     @SuppressWarnings("unchecked")
-    public <T, E1 extends Exception, E2 extends Exception> T transaction(TransactionActionWithExceptions<T, D, E1, E2> action, Class<E1> exClass1, Class<E2> exClass2) throws E1, E2
+    public <T, E1 extends Exception, E2 extends Exception> T transactionWithLocalRetry(RetryableTransactionActionWithExceptions<T, D, E1, E2> action, Class<E1> exClass1, Class<E2> exClass2) throws E1, E2
```
komamitsu: Can you clarify the policy when to call `transactionWithLocalRetry()` not other `transaction()`?

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
```diff
@@ -274,42 +283,105 @@ public int getIndex()
...
     {
+        // TODO
+        // Here should throw error if currentTransaction.get() != null to avoid race condition under high load.
+        // In following scenario, a deadlock happens:
+        // * thread1 acquires a row lock on RDBMS (such as lockTaskIfExists), then calls transactionWithLocalRetry.
```
komamitsu: These logics of locking sounds a bit complicated to me. It may prevent other developers from trying to look into around here. Maybe we'd better simplify these lock logics or write the policies of lock logics as a document later.

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
```diff
@@ -274,42 +283,105 @@ public int getIndex()
...
+            }
         }
     }
 
     @SuppressWarnings("unchecked")
-    public <T, E1 extends Exception, E2 extends Exception> T transaction(TransactionActionWithExceptions<T, D, E1, E2> action, Class<E1> exClass1, Class<E2> exClass2) throws E1, E2
+    public <T, E1 extends Exception, E2 extends Exception> T transactionWithLocalRetry(RetryableTransactionActionWithExceptions<T, D, E1, E2> action, Class<E1> exClass1, Class<E2> exClass2) throws E1, E2
     {
+        // TODO
+        // Here should throw error if currentTransaction.get() != null to avoid race condition under high load.
```
komamitsu: We don't need to have the same consideration on other `transaction()`?

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
```diff
@@ -274,42 +283,105 @@ public int getIndex()
...
                                     "for the original exception.", ex);
                         }
-                        if (!ts.retryNext) {
-                            validateTransactionAndCommit(handle);
-                            return retval;
-                        }
-                        handle.rollback();
-                        ts.retryNext = false;
-                        ts.retryCount += 1;
+                        handle.commit();
```
komamitsu: We don't need to call `validateTransactionAndCommit()` instead?

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
```diff
@@ -380,62 +455,78 @@ private void validateTransactionAndCommit(Handle handle)
...
+        Handle nested = currentTransaction.get();
+        if (nested != null) {
+            return action.call(nested, nested.attach(daoIface));
         }
-        catch (InterruptedException ex) {
-            throw Throwables.propagate(ex);
+        else {
+            try {
+                return transactionRetryExecutor.runInterruptible(() -> {
+                    try (Handle handle = dbi.open()) {
```
komamitsu: Is it possible to call a nested transactional methods like `transaction()` in `action.call()`? If so, this `handle` is needed to be set to `currentTransaction`?

digdag-core/src/main/java/io/digdag/core/database/DatabaseModule.java
```diff
@@ -22,7 +22,23 @@ public void configure(Binder binder)
...
+        // case because creating a handle opens another connection. Opening a connection in a transaction causes
+        // deadlock as following:
+        //
+        // * Thread1 starts a transaction and acquires a row lock (such as lockTaskIfExists).
+        // * Thread2 opens a connection and try to lock the same row. This is blocked blocked because Thread1
+        //   already locks the row.
+        // * Thread1 tries to open another connection in the transaction. This could be blocked for ever by
+        //   Thread2 if number of opened connection reached maximumPoolSize when thread2 opened a connection.
+        //
+        binder.bind(DBI.class).toProvider(DbiProvider.class).in(Scopes.SINGLETON);
```
komamitsu: 👍

komamitsu:
Added some comments. Can you take a look at them?

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
frsyuki: It must be better to restore interrupt state. I fix it 👍

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
frsyuki: Purpose of transactionWithLocalRetry is specifically put-and-lock implementations: putAndLockProject and putAndLockSession. These 2 operations want to insert a record if not exist, lock it within the same transaction, and rollback everything if something failed in the transaction.

putAndLockProject is not a big issue because it won't be called within another transaction.

putAndLockSession needs above behavior because session is created only when one doesn't exist. If exists, an attempt must exist. The current implementation for PostgreSQL (<= 9.4) and H2 is to INSERT first and retry the transaction if conflict happens. Because the conflict is an expected situation, here shouldn't raise exceptions.

frsyuki: For future, I think that putAndLockProject and putAndLockSession should use following approach:

* Assume PostgreSQL >= 9.5 that supports INSERT ON DUPLICATE KEY UPDATE.
* If it's H2 database, use a lock in Java (io.digdag.core.database.LocalLockMap).

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
frsyuki: I think this transactionWithLocalRetry method itself should be removed away

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
frsyuki: I also tried to do that but I couldn't because DatabaseSecretControlStore already depends on it. Well, actually DatabaseSecretControlStore.setProjectSecret seems broken...(it throws conflict exception if 2 threads run the method concurrently).

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
frsyuki: Hm...yes, maybe.
validateTransactionAndCommit takes effect when we need to rollback the transaction but inner action didn't throw an exception. Rollback is mandatory when a SQL statement fails. And usually, inner action throws an exception if a SQL statement fails. validateTransactionAndCommit can prevent unexpected behavior (commit is called but nothing is committed) when an inner action ignored SQL exceptions. But it has overhead (it issues a SELECT statement).

transactionWithLocalRetry always doesn't follow it (RetryableTransactionActionWithException) because it simply sets retry flag and doesn't throw exception when a SQL statement fails.

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
frsyuki: ah, it is theoretically possible but no code does it

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
frsyuki: oh H2 supports MERGE statement. Here can use it. I hope PostgreSQL also supports MERGE...it's a part of ANSI SQL:2003 standard.

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
frsyuki: I updated code and comments.

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
komamitsu: I see.

> Moreover, we shouldn't write such code...do you think BasicDatabaseStoreManager should have a assertion to make nested transaction() in autoCommit failed?

Ideally, it's best to prevent those codes at compile time...

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
```diff
@@ -431,13 +431,21 @@ public int getIndex()
...
         //
-        // Here assumes that HikariDB doesn't overwrite isValid.
+        // 1. A thread starts a transaction
+        // 2. The thread issues a statement and it fails (e.g. unique key violation)
+        // 3. The thread catch the exception and does other things
+        // 4. The thread issues COMMIT
+        // 5. PostgreSQL silently translate COMMIT to ROLLBACK without errors. The thread
+        //    assumes that the transaction is committed but actually everything is rolled back.
+        //
+        // To avoid this issue, the thread needs to send a dummy SELECT statement after step 4.
```
komamitsu: Which class/method issues the SELECT statement after COMMIT now?

komamitsu:
👍

komamitsu:
> Thread2 opens a connection and tries to lock the same row. This is blocked because thread1 already locks the row.

@frsyuki 
How about using `NOWAIT` for `FOR UPDATE`? With `NOWAIT`, if Thread2 fails to get the lock (55P03?) and releases the connection immediately, Thread1 can obtain a connection. Thread2 should retry later after an appropriate wait.

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
frsyuki: PgConnection.isValid does it.

frsyuki: oops...the comment is wrong. it should be "before" step 4.

frsyuki:
@komamitsu I think it doesn't work well because of two reasons:

* UPDATE and DELETE statements also lock rows. This happens if setSecrets is nested in putAndLockProject transaction. But UPDATE and DELETE don't support NOWAIT.
* when two locking operations are nested, failure of a nested operation requires retry of the entire transaction. For example, putAndLockSession is nested in lockScheduleById. putAndLockSession can't retry locally when NOWAIT fails. PostgreSQL requires rollback of the entire transaction when a single statement fails. It needs to retry the entire lockScheduleById transaction. But from lockScheduleById method's point of view, it's hard to tell it is safely retryable or not because exception might be wrapped and it's not sure how it's wrapped.

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
danielnorberg: Afaik pg does not yet support merge.

digdag-core/src/main/java/io/digdag/core/database/BasicDatabaseStoreManager.java
danielnorberg: Should `DatabaseSecretControlStore` use some other method instead of `transaction()`? Maybe we could add some comments to the methods in `BasicDatabaseStoreManager` so it's easier to know when to use what.

danielnorberg:
I think I would prefer the transaction handle to be explicitly managed and passed in to methods instead of being magically managed in the background using thread local storage.

komamitsu:
Yeah. It might help to avoid this kind of issue. I feel the current implementation does a bit too implicit although the code size is reduced.

frsyuki:
Passing transaction handle around makes components tightly coupled as you see code in #267.

danielnorberg:
The `setProjectSecret` operation is not expected to run in the same transaction as other operations (as it might use another backing store like vault), but I believe that it's fair to expect that `{Project,Workflow,Schedule}{Control,}Store`, etc are all backed by one and the same RDBMS database and that their respective operations can be part of the same transaction. I would rather have that coupling be made explicit instead of being implicit magic.

PR #266, opened 2016-09-08
Added authentication skip if config is empty

dzannotti:
This allows the client to skip the login screen if auth items is empty (which implies that the authentication data comes from cookies)


danielnorberg:
👍

PR #267, opened 2016-09-09
Secret consistency

frsyuki:
If secrets table has two entries for the same scope & key, selecting a secret
is undeterministic. On PostgreSQL, results may change after vacuum.
This duplicated entry may happen even using DELETE + INSERT transaction
because DELETE doesn't lock the unexistent row. So, INSERT may run
concurrently. Concurrent put of secrets may results in this state.

Unique index solves it. However, because anyways INSERT may run concurrently,
it may fail with unique constraint violation error.

A solution is to run DELETE and INSERT when a project is locked.


digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20160908175551_KeepSecretsUnique.java
```diff
@@ -0,0 +1,21 @@
...
+    public void migrate(Handle handle, MigrationContext context)
+    {
+        if (context.isPostgres()) {
+            // make sure records are unique
+            handle.update("delete from secrets where id = any(select id from (select row_number() over (partition by site_id, project_id, scope, key order by id) as i, id from secrets) win where i > 1)");
+        }
+        else {
+            // h2 doesn't support window function...
+        }
+        handle.update("create unique index secrets_unique_on_site_id_and_project_id_and_scope_and_key on secrets (site_id, project_id, scope, key)");
```
danielnorberg: 👍 This should've been an unique index from the start. Oops.

danielnorberg:
1. We probably want to keep the ability to use other secret storage backends like vault, etc.
2. We will likely add site-level secrets in the future. It would seem to make sense to gather those secret set/get methods in the Secret{Control}Store interfaces instead of moving them to the Project{Control}Store interfaces.

Could we go ahead with the database schema migration and hold off on the other changes while we continue to discuss them?

danielnorberg:
It seems undesirable to have long-lived row locks e.g. while making requests to a secret storage system like vault. Therefore we do not want to invoke `setProjectSecret` in the scope of the transaction and lock of `putAndLockProject`.

Taking a step back, it seems that what this PR is attempting to address is the potential for concurrent project push operations to result in an "inconsistent" set of `PROJECT_DEFAULT` secrets. The `PROJECT_DEFAULT` scope is designed to be a fallback of last resort if the user has not specified a suitable secret which is then stored in the `PROJECT` scope.

I believe that it is acceptable for concurrent project pushes to result in either of the set of secrets being persisted. I believe that it is not necessary (or sufficiently desirable) to ensure that the `PROJECT_DEFAULT` secrets are stored atomically wrt to the project store operation.

It might seem desirable to perform the `setProjectSecret` invocations (https://github.com/treasure-data/digdag/blob/master/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java#L528) in a single transaction so that in the event of concurrent project pushes A and B, either the set of `PROJECT_DEFAULT` secrets from A _or_ B are stored (where their keys overlap). However, the backing store might not offer this capability. And again, we do not want to wrap this in a row lock.

frsyuki:
We should avoid long-lived locking generally but it is totally reasonable to lock rows for consistency. For this particular case, locking a project is acceptable to prevent inconsistency because of concurrent push of a project or concurrent update of secrets. Task execution won't be blocked because SELECT without FOR UPDATE won't be blocked.

When we think about other secret storages in the future, desired order is putting secrets first and putting project/revision later so that we accept garbage keys on the secret storages and reject inconsistency on the database. Because database is the visible entity for users. The other order can cause inconsistency; uploading a project returned a HTTP error but project is updated, or uploading a project returned HTTP 200 OK but secrets are not updated. It's difficult to investigate this kind of problems if happened. This concerns me because inquiry about a failure of something might force us to investigate lots of logs in case RDB became unstable.

To implement appropriate consistency control without transaction, secret storage needs more code so that garbage entries on the secret storage is associated to a revision as well as a project where newly assigned revision_id is rolled back and not reused any more when putting secrets failed. With this way, secrets will be kept remained as garbage but doesn't affect consistency. This means putting secrets within a transaction where the project is locked.

frsyuki:
hmm...don't you rethink about putting secrets in a transaction?
If we put secrets while a project is locked, #264 can make it transactional using a thread-local handler without changing interfaces. It's a simple move of 2 lines. Although the code becomes hard to understand why it becomes transactional.

PR #268, opened 2016-09-09
Add unique index on secrets table

danielnorberg:
Cherry picked from https://github.com/treasure-data/digdag/pull/267

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSecretControlStore.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSecretControlStore.java
@@ -18,9 +18,21 @@ class DatabaseSecretControlStore
 
     DatabaseSecretControlStore(DatabaseConfig config, DBI dbi, int siteId, SecretCrypto crypto)
     {
-        super(config.getType(), Dao.class, dbi);
+        super(config.getType(), dao(config.getType()), dbi);
         this.siteId = siteId;
         this.crypto = crypto;
+    }
+
+    private static Class<? extends Dao> dao(String type)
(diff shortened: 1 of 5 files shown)


digdag-core/src/main/java/io/digdag/core/database/DatabaseSecretControlStore.java
```diff
@@ -61,9 +72,29 @@ public void deleteProjectSecret(int projectId, String scope, String key)
...
+    interface PgDao
+            extends Dao
+    {
+        @Override
         @SqlUpdate("insert into secrets" +
                 " (site_id, project_id, scope, key, engine, value, updated_at)" +
-                " values (:siteId, :projectId, :scope, :key, :engine, :value, now())")
-        int insertProjectSecret(@Bind("siteId") int siteId, @Bind("projectId") int projectId, @Bind("scope") String scope, @Bind("key") String key, @Bind("engine") String engine, @Bind("value") String value);
+                " values (:siteId, :projectId, :scope, :key, :engine, :value, now())" +
+                " on conflict (site_id, project_id, scope, key) do update set engine = :engine, value = :value, updated_at = now()")
```
komamitsu: INSERT ~ ON CONFLICT is supported by PostgreSQL 9.5 or newer. I'm not sure if we support PostgreSQL 9.4 or older as Digdag-server's database...

komamitsu:
LGTM 👍 

@frsyuki Can you take a look at this just in case?

frsyuki:
it's ok to drop support of PostgreSQL < 9.5 since now.

PR #269, opened 2016-09-09
[ui]: Accepts SHA from process.env

dzannotti:
With this it should accept a third party SHA or use the repo one (useful for external config builds)


danielnorberg:
👍

PR #271, opened 2016-09-09
Longer timeout in TdWaitIT + Hive queries

frsyuki:
Occasionally TdWaitIT is failing due to timeout.
This PR tries to fix it by increasing timeout duration.

Hive queries are taking time especially using low-tier price plan
that doesn't have guaranteed performance.


frsyuki:
@danielnorberg PTAL~

danielnorberg:
👍

PR #274, opened 2016-09-12
[WIP] assign secret scopes to digdag and td config

danielnorberg:
- Assign PROJECT scope to digdag config
- Assign PROJECT_DEFAULT scope to td.conf

This ensures that e.g. td.apikey set in digdag config has higher priority than the apikey picked up from td.conf.


frsyuki:
I see. Looks good 👍 
It might be good idea to create `ScopedSecretStore` interface where SecretStore adapter checks scope automatically before calling `ScopedSecretStore.get`.

yoyama:
It is old PR and not fit current situation. Close it.

PR #276, opened 2016-09-12
Pass project path to operators

frsyuki:
Operators were getting working directory path (`newTaskExecutor(workspacePath, request)`). This path is the directory where the workflow definition file is placed. Operators did not allow users to use files outside of this directory (e.g. `td>: ../query.sql` is an error).

However, this is not useful when #254 adds supports for workflows placed in a subdirectories. `td>: ../query.sql` used in `my_project/subdir/common.dig` should be allowed because query.sql is still in the project path.

To implement it, operator needs both project path and working directory path. Relative path is resolved using working directory path, and the calculated path is validated using project path.

This PR changes operator API so that operators get project path (`newOperator(projectPath, request)`).

Internal implementation is:

* Project path is given by `WorkspaceManager`.
  * `ExtractArchiveWorkspaceManager` (used in server mode)) extracts project archive to a temporary directory and uses the directory path as the project path.
  * `LocalWorkspaceManager` (used in local mode) uses `_project_path` parameter as the project path.
* Working directory path = project path + `_workdir` parameter.

There're no big behavior changes because digdag doesn't support workflows in subdirectories (WorkflowResourceMatcher doesn't match with them). So this PR is supposed to be a refactoring.


frsyuki:
@komamitsu can I ask you to review this?

digdag-plugin-utils/src/main/java/io/digdag/util/Workspace.java
```diff
@@ -126,9 +151,12 @@ public String templateCommand(TemplateEngine templateEngine, Config params, Stri
                 catch (FileNotFoundException | NoSuchFileException ex) {
                     throw new ConfigException("File not found: " + ex.getMessage(), ex);
                 }
-                catch (RuntimeException | IOException ex) {
+                catch (IOException ex) {
                     throw new ConfigException("Failed to read a template file: " + command + ": " + ex.getClass(), ex);
                 }
+                catch (RuntimeException ex) {
+                    throw new ConfigException("" + ex.getMessage() + ": " + command, ex);
```
komamitsu: [minor] Throwable#getMessage returns String, so we can remove the first empty string literal.

digdag-plugin-utils/src/main/java/io/digdag/util/BaseOperator.java
```diff
@@ -16,17 +16,17 @@
...
     protected final List<Config> inputs;
     protected final List<Config> outputs;
 
-    public BaseOperator(Path workspacePath, TaskRequest request)
+    public BaseOperator(Path projectPath, TaskRequest request)
     {
-        this.workspacePath = workspacePath;
-        this.workspace = new Workspace(workspacePath);
+        this.projectPath = projectPath;
+        this.workspace = Workspace.ofTaskRequest(projectPath, request);
```
komamitsu: [minor] This variable is closed in `run()` not `close()` or something explicitly. So it seems to be a bit confusing life time to me. How about removing this variable and modifying `run()` like this?
```
    @Override
    public TaskResult run(TaskExecutionContext ctx)
    {
        RetryControl retry = RetryControl.prepare(request.getConfig(), request.getLastStateParams(), false);
        try {
            try (Workspace workspace = Workspace.ofTaskRequest(projectPath, request)) {
                return runTask(ctx, workspace);
            }
        }
```

komamitsu:
👍 
Although I added some minor comments

digdag-plugin-utils/src/main/java/io/digdag/util/Workspace.java
frsyuki: This works when `ex.getMessage()` returns null...although String.format is preferred anyway

digdag-plugin-utils/src/main/java/io/digdag/util/BaseOperator.java
frsyuki: I see. hm...but I think operators want to use this.workspace at its constructor

frsyuki: I think `Operator.close` should be added so that OperatorManager calls it.

danielnorberg:
Maybe `TaskExecutionContext` could've been used instead of adding `Path projectPath` to `newOperator`? https://github.com/treasure-data/digdag/pull/276/files#diff-c051095e7c4a54050d7bf29e9de6c393R37

The point of `TaskExecutionContext` is to make it possible to make these kinds of changes in a backwards compatible manner.

frsyuki:
@danielnorberg `TaskExecutionContext` is given to `runTask` method but `BaseTdJobOperator` is using workspace (that needs projectPath) in its constructor to read query files.

danielnorberg:
@frsyuki I'd prefer to pass in `TaskExecutionContext` or some other injection interface to the constructor as well so that future changes like this can be done in a backwards compatible manner.

PR #282, opened 2016-09-14
Scrap backward compatibility of deprecated Operator.run() method

frsyuki:
#276 already spoiled plugin backward compatibility


danielnorberg:
👍  Let's ensure that we can do future changes without breaking backwards compatibility by getting evolvable operator interfaces into place so that we don't have to make method signature changes like in #276.

frsyuki:
👍

PR #283, opened 2016-09-14
[WIP] Closeable operator

frsyuki:
This is a follow-up of https://github.com/treasure-data/digdag/pull/276#r78507211

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -317,9 +317,9 @@ public class OperatorManager
         OperatorContext context = new DefaultOperatorContext(
                 projectPath, mergedRequest, secretProvider, privilegedVariables);
 
-        Operator operator = factory.newOperator(context);
-
-        return operator.run();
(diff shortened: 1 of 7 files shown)


danielnorberg:
In the long run I would like us to consider executing each operator in a container and utilize the container runtime/platform to clean up after each containerized operation execution.

frsyuki:
That is an interesting idea.

frsyuki:
I think local-mode execution is important because digdag should be able to cover use cases where people use their local machine as a develop environment to gradually improve workflow definitions.
For long term, Digdag should make it possible to use local files even on server environment rather than prohibiting them. Cleaning up all generated files is not a desired direction, in my opinion.
But always deleting `<project-dir>/.digdag/tmp` seems ok without calling `Operator.call`.

danielnorberg:
Yeah, and I think using docker containers for all operators could allow use of local files and "persistent workspaces" etc.

frsyuki:
rebased

yoyama:
I guess this PR enable that each operator can override close().
I think LGTM. But we need additional modification because this PR is old and some operators may be added.
So I added [WIP] to the title.

yoyama:
Close because no progress.

PR #290, opened 2016-09-16
Call operator supports files in subdirectory

frsyuki:
Dividing a large workflow definition into multiple files is helpful to
organize a complex workflow definition. call> is more straightforward
(and less hacky) compared to !include in terms of syntax.

This changes `call>` operator to be able to load workflow files from
any *.dig files in the project archive including files in subdirectories.

When `call>` calls a workflow in a subdirectory, the called workflow uses
the subdirectory as the working directory. This is a significant difference from
`!include`.


frsyuki:
@danielnorberg could you take a look?
I checked hosted digdag environment and found that no one uses project_id parameter.

danielnorberg:
👍  nice!

PR #291, opened 2016-09-19
Optimize WorkflowExecutor.propagateAllBlockedToReady

frsyuki:
This method takes very long time if number of BLOCKED tasks
increased because it has a loop in it. Because the loop blocks
shutdown process, server returns 503 Service Unavailable to
clients for long time.

This method was fetching all BLOCKED tasks and get their parent_id.
If this method uses SQL to get the parent_id list directly, number of
the loop and data transfer from database to the server become several
times smaller.

```
=> select count(*) from (select id, parent_id, state from tasks where state = 0) orig;
  count
---------
 5278779
(1 row)

Time: 2877.191 ms
```

```
=> select count(*) from (select distinct parent_id from tasks where state = 0) optimized;
 count
--------
 135461
(1 row)

Time: 3972.704 ms
```

Query plan:

```
=> explain analyze select distinct parent_id from tasks where parent_id > 12345 and state = 0 order by parent_id limit 100;
                                                                            QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.43..356.94 rows=100 width=8) (actual time=0.039..3.068 rows=100 loops=1)
   ->  Unique  (cost=0.43..240934.82 rows=67581 width=8) (actual time=0.037..2.779 rows=100 loops=1)
         ->  Index Only Scan using tasks_on_parent_id_and_state on tasks  (cost=0.43..227809.86 rows=5249982 width=8) (actual time=0.033..1.553 rows=1088 loops=1)
               Index Cond: ((parent_id > 12345) AND (state = 0))
               Heap Fetches: 1023
 Planning time: 0.102 ms
 Execution time: 3.184 ms
(7 rows)
```

This query is scalable against number of tasks because query plan doesn't include SORT node.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -319,6 +319,25 @@ public class DatabaseSessionStoreManager
     }
 
     @Override
+    public List<Long> findDirectParentsOfBlockedTasks(long lastId)
+    {
+        return autoCommit((handle, dao) ->
+                handle.createQuery(
+                    "select distinct parent_id" +
+                    " from tasks" +
+                    " where parent_id > :lastId" +
+                    " and state = " + TaskStateCode.BLOCKED_CODE +
+                    " order by parent_id" +
+                    " limit :limit"
+                    )
+                .bind("lastId", lastId)
+                .bind("limit", 100)
+                .mapTo(Long.class)
+                .list()
+            );
+    }
+
+    @Override
     public boolean requestCancelAttempt(long attemptId)
     {
         return transaction((handle, dao, ts) -> {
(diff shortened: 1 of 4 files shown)


frsyuki:
I tried to combine the query with `trySetChildrenBlockedToReadyOrShortCircuitPlannedOrCanceled` but generated query becomes slow. It's better to avoid long-running single UPDATE query to not lock rows for long time.

```
explain
update tasks
set updated_at = now(), state = case
when task_type = 1 then 5
when (state_flags & 1) != 0 then 9
else 1
end
where parent_id = any(array(
  with recursive parents (id) as (
    (
      select parent_id as id
      from tasks
      where parent_id is not null and state = 0
      order by parent_id
      limit 1
    )
    union all
    select (
      select parent_id as id
      from tasks
      where parent_id > parents.id and state = 0
      order by parent_id
      limit 1
    )
    from parents
    where parents.id is not null
  )
  select parents.id
  from parents
  join tasks on parents.id = tasks.id
  where tasks.state in (5, 7)
))
and not exists (
  select * from tasks up
  join task_dependencies dep on up.id = dep.upstream_id
  where dep.downstream_id = tasks.id
  and up.parent_id = tasks.parent_id
  and up.state not in (7)
);
```

frsyuki:
@komamitsu  could you take a look?

komamitsu:
👍

PR #294, opened 2016-09-20
Limit number of active attempts

### LINKED ISSUE #293: Limiting number of active workflow attempts per site
There're schedules and REST API (`PUT /api/attempts`) that issue new attempts. `require>` also issues a  new attempt but it should be implemented as an alias of REST API.

frsyuki:
Fixes #293.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/Limits.java
+++ b/digdag-core/src/main/java/io/digdag/core/Limits.java
@@ -5,9 +5,18 @@ public class Limits
     private static final long MAX_WORKFLOW_TASKS = Long.valueOf(
             System.getProperty("io.digdag.limits.maxWorkflowTasks", "1000"));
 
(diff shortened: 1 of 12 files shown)


danielnorberg:
I guess we also need the `ScheduleExecutor` to be smart about this limit and not attempt to start more sessions than allowed.

frsyuki:
My current idea is that `ScheduleExecutor` puts execution of a schedule off for several minutes if `WorkflowExecutor` throws AttemptLimitExceededException. Do you mean that `ScheduleExecutor` needs to be even smarter?

danielnorberg:
That sounds good enough 👍

frsyuki:
Rebased from master.
@danielnorberg PTAL

danielnorberg:
👍

PR #302, opened 2016-09-23
Use upsert instead of putAndLock retrying transactions

frsyuki:
This simplifies code. PostgreSQL >= 9.5 is assumed since secrets store.

PR #303, opened 2016-09-23
Make LimitExceededException checked

frsyuki:
This found that:

* inserting notification task also is also limited but it seems bad

This depends on #302 and #294.


danielnorberg:
Yeah, I think inserting tasks into a workflow to perform user notifications is inherently broken. It's a brittle hack. Just imagine that a workflow execution ends up in such a state that no further tasks can execute (in this case due to the task limit being reached, but assume that there might be other states where tasks fail to execute as well), then inserting additional tasks to notify the user of their broken workflow session attempt is quite futile. 

Ideally there'd be another reliable execution mechanism that can be leveraged to perform tasks like user notifications. Perhaps we can factor out reliable task execution into another lower layer, separate from workflow execution. Then both the workflow execution engine and the user notification layer, mail sending mechanism, etc could utilize the reliable task execution mechanism.

frsyuki:
Updated. @danielnorberg could you take a look?
712a0f0e11727389f832dc4fa02ca301736bcd05 is the main commit. Changes of `BasicDatabaseStoreManager.transaction` and `ScheduleExecutor` are basically for propagating checked exceptions from a transaction to outer callers.

frsyuki:
`TaskQueueServer` can be used for any task execution in terms of its interface design. However, task execution worker (=`MultiThreadAgent`) assumes that elements in the queues are always `TaskRequest` and then `OperatorManager` calls `WorkflowExecutor.taskSucceeded` callback. So, more isolation of Agent from WorkflowExecutor is necessary to use `TaskQueueServer` as a generic task execution mechanism.

frsyuki:
rebased

digdag-core/src/main/java/io/digdag/core/database/DatabaseScheduleStoreManager.java
```diff
@@ -148,23 +169,29 @@ public DatabaseScheduleControlStore(Handle handle)
...
         @Override
-        public boolean updateNextScheduleTime(int schedId, ScheduleTime nextTime)
+        public void updateNextScheduleTime(int schedId, ScheduleTime nextTime)
+            throws ResourceNotFoundException
         {
             int n = dao.updateNextScheduleTime(schedId,
                     nextTime.getRunTime().getEpochSecond(),
                     nextTime.getTime().getEpochSecond());
-            return n > 0;
+            if (n <= 0) {
```
danielnorberg: `n < 0` if you got an integer overflow or what? `ETOOMANYROWS` =)

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -848,7 +848,7 @@ public boolean setDoneState(long taskId, TaskStateCode beforeState, TaskStateCod
             return n > 0;
         }
 
-        public boolean setDoneStateShortCircuit(long taskId, TaskStateCode beforeState, TaskStateCode afterState, Config error)
+        public boolean setErrorStateShortCircuit(long taskId, TaskStateCode beforeState, TaskStateCode afterState, Config error)
```
danielnorberg: Looks like this PR got a bunch of stuff from #310. Rebasing problem?

danielnorberg: Oh. That PR was against this PR..?

danielnorberg: Would be easier to review them separately, or is there a reason they need to go into one PR?

digdag-core/src/main/java/io/digdag/core/database/DatabaseScheduleStoreManager.java
frsyuki: schedules.id is serial (int). so it never happens. I simply doesn't trust my understanding of jDBI or H2 enough.

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
frsyuki: oh right. Merging #310 first was my mistake.
#310 was to this PR because it caused conflicts.

digdag-core/src/main/java/io/digdag/core/database/DatabaseScheduleStoreManager.java
danielnorberg: Ok, maybe we want to `assert n >= 0`?

danielnorberg:
👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseScheduleStoreManager.java
frsyuki: That seems better

frsyuki:
Thank you for your review! 👍

PR #306, opened 2016-09-23
[UI]: Remove td forced requirement

dzannotti:
I noticed that the recent changes forces request to td-api before showing the UI. This might not be ideal for every client, so i gated it behind a config flag


danielnorberg:
👍

PR #307, opened 2016-09-23
[UI] Fixed build/dev for linux

dzannotti:
By updating flow and node-sass, because both versions used were known to cause issues in linux envs


digdag-ui/model.js
```diff
@@ -70,7 +70,7 @@ export type Task = {
 export type Attempt = {
   id: number;
   project: IdName;
-  workflow: NameOptionalId;
+  workflow: NameId;
```
danielnorberg: The `Attempt` type should probably match the underlying rest api type.

https://github.com/treasure-data/digdag/blob/master/digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java#L23

digdag-ui/model.js
```diff
@@ -27,9 +27,9 @@ export type IdName = {
   name: string;
 }
 
-export type NameOptionalId = {
+export type NameId = {
   name: string;
-  id: ?number;
+  id: number;
```
danielnorberg: https://github.com/treasure-data/digdag/blob/master/digdag-client/src/main/java/io/digdag/client/api/NameOptionalId.java

dzannotti: True and agreed, however the JS ui handles the id as being always present (see https://github.com/treasure-data/digdag/blob/master/digdag-ui/console.jsx#L299) Should the correct behaviour be checking against the existance of that id js side?

danielnorberg: Indeed. Perhaps it should.

Issue #46, opened 2016-04-16
plural REST API with ?name= parameter replaces singular REST API


frsyuki:
There are 3 singular REST API:
- `GET /api/project?name=<name>`
- `GET  /api/projects/{id}/workflow?name=name[&revision=name]`
- `GET  /api/workflow?project=<name>&name=<name>[&revision=<name>]`

Replace them by adding following REST API:
- `GET /api/projects?name=<name>`
- `GET /api/projects/{id}/workflows?name=<name>[&revision=name]`
- (no replacement for `/api/workflow`. Use `/api/projects?name=<name` and `/api/projects/{id}/workflows` instead)

If a resource exist, these new REST API return an array with one element in it. If a resource doesn't exist, it returns an empty array (not 404 Not Found). With this way, client can tell that a project doesn't exist when `GET /api/projects/{id}/workflows?name=<name>[&revision=name]` returns 404 Not Found because it returns an empty array if a project exists but workflow doesn't exist.

PR #309, opened 2016-09-26
disable/enable schedule

danielnorberg:
Make it possible to mark a schedule as disabled, effectively suspending or "pausing" a scheduled workflow until it is enabled again.

Scheduled sessions that have already been started are not affected.

When a schedule is re-enabled, it picks up where it stopped. Sessions will not be skipped as a consequence of a disabling and later enabling a schedule.

Also made it possible to fetch schedules for a single project and a single workflow.

```
GET  /api/schedules?project_id={id}
GET  /api/schedules?project_id={id}&workflow={name}
```


danielnorberg:
@frsyuki PTAL

digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20160926123456_AddDisabledColumnToSchedules.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.skife.jdbi.v2.Handle;
+
+public class Migration_20160926123456_AddDisabledColumnToSchedules
+        implements Migration
+{
+    @Override
+    public void migrate(Handle handle, MigrationContext context)
+    {
+        handle.update("alter table schedules add column disabled_at bigint");
```
frsyuki: An index needs update to keep `lockReadySchedules` fast even if there are many disabled schedules:

```
        if (context.isPostgres()) {
            handle.update("drop index schedules_on_next_run_time");
            handle.update("create index schedules_on_next_run_time where deleted_at is null");
        }
```

digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20160926123456_AddDisabledColumnToSchedules.java
```diff
@@ -0,0 +1,13 @@
+package io.digdag.core.database.migrate;
+
+import org.skife.jdbi.v2.Handle;
+
+public class Migration_20160926123456_AddDisabledColumnToSchedules
```
frsyuki: `timestamp` type (`timestamp with time zone` on postgresql) seems better in this case.

I used `bigint` for session_time and next_run_time because I wanted to avoid confusion of different handling of timestamp precision depending on databases (nano, micro, or seconds). Precision of schedules and session_time are fixed to seconds. But `foobar_at` columns should have more precision.

```
        handle.update("alter table schedules add column disabled_at bigint");
```

danielnorberg: Ok. I guess I'll do `disabled_at = now()` instead of generating the timestamp in Java and fighting with pg & h2 timestamp parsing.

digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20160926123456_AddDisabledColumnToSchedules.java
danielnorberg: 👍

digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20160926123456_AddDisabledAtColumnToSchedules.java
```diff
@@ -0,0 +1,25 @@
...
+                    " add column disabled_at timestamp with time zone");
+        }
+        else {
+            handle.update("alter table schedules" +
+                    " add column disabled_at timestamp");
+        }
+
+        if (context.isPostgres()) {
+            handle.update("drop index schedules_on_next_run_time");
+            handle.update("create index schedules_on_next_run_time on schedules (next_run_time) where deleted_at is null");
```
frsyuki: sorry, my comment was wrong... `deleted_at` -> `disabled_at`

frsyuki:
I think it `GET  /api/projects/{id}/schedules?workflow={workflowName}` is the way to go...! So I wrote code: #313. You suggested #46 design as far as I remember.

danielnorberg:
👍  Thanks!

PR #310, opened 2016-09-26
Remove unnecessary UPDATE call when leaf tasks successfully finished

frsyuki:
taskSucceeded callback uses setRunningToPlannedSuccessful to change task
state to PLANNED without flags (flags are set only when state is in
PLANNED). setDoneFromDoneChildren changes task state to SUCCESS if the
task is in PLANNED state and there're no flags if there're no children.
This means that taskSucceeded callback can set SUCCESS state immediately
if it doesn't add children.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -848,7 +848,7 @@ public class DatabaseSessionStoreManager
             return n > 0;
         }
 
-        public boolean setDoneStateShortCircuit(long taskId, TaskStateCode beforeState, TaskStateCode afterState, Config error)
+        public boolean setErrorStateShortCircuit(long taskId, TaskStateCode beforeState, TaskStateCode afterState, Config error)
         {
             long n = dao.setDoneState(taskId, beforeState.get(), afterState.get());
             if (n > 0) {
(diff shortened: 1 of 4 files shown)


digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
```diff
@@ -100,6 +100,8 @@
  *   taskSucceeded:
  *     lockedTask.setRunningToPlannedSuccessful:
  *       : PLANNED
+ *     lockedTask.setRunningToShortCircuitSuccess:
```
danielnorberg: Maybe mention here that this path is only taken if no subtask was added?

danielnorberg:
LGTM 👍

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
frsyuki: 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -848,7 +848,7 @@ public boolean setDoneState(long taskId, TaskStateCode beforeState, TaskStateCod
             return n > 0;
         }
 
-        public boolean setDoneStateShortCircuit(long taskId, TaskStateCode beforeState, TaskStateCode afterState, Config error)
+        public boolean setErrorStateShortCircuit(long taskId, TaskStateCode beforeState, TaskStateCode afterState, Config error)
```
danielnorberg: Looks like this PR got a bunch of stuff from #310. Rebasing problem?

PR #312, opened 2016-09-27
[UI]: Switch to standard linting

dzannotti:
For coherence with td-console, i switched the linting to js-standard,
@danielnorberg sadly huge changeset because of the semicolon removal


danielnorberg:
Relint all the codes! I dream of doing this with the java code =D

👍

PR #314, opened 2016-09-27
Bundle Moment.js

### LINKED ISSUE #215: strftime with variable
Formatting timestamp is one of the common use cases of `${...}` expression.
Vanilla JavaScript doesn't have good support for it.
Having a strftime library such as https://github.com/samsonjs/strftime as a built-in library should be useful.

Discussions are:

* How should we recommend users to convert session_time (string) to a Date object?
* How to set default timezone? (I think users will expect that `timezone: ` parameter is used as the default timezone).

frsyuki:
Bundles [Moment.js](http://momentjs.com/) 2.15.1 to the JavaScript config eval engine.
Fixes #215.


frsyuki:
@danielnorberg PTAL

danielnorberg:
👍

westwick:
woot!

PR #315, opened 2016-09-28
Add /workflows REST API

frsyuki:
This improves latency of typical operations including `digdag workflows`
command.

New indexes are for PostgreSQL so that it can run the query without Sort or HashAggregate nodes as following. Query execution time doesn't increase depending number of old deleted projects or old revisions.

```
=> explain select wd.*, wc.config, wc.timezone, proj.id as proj_id, proj.name as proj_name, proj.deleted_name as proj_deleted_name, proj.deleted_at as proj_deleted_at, proj.site_id, proj.created_at as proj_created_at, rev.name as rev_name, rev.default_params as rev_default_params from (select * from workflow_definitions wf where wf.revision_id = any(array(select max(r.id) from revisions r join projects p on r.project_id = p.id where p.site_id = 1 and p.deleted_at is null group by r.project_id)) and wf.id > 10 order by wf.id limit 10) wd join revisions rev on rev.id = wd.revision_id join projects proj on proj.id = rev.project_id join workflow_configs wc on wc.id = wd.config_id order by wd.id;
                                                                 QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=8.70..34.10 rows=9 width=368)
   ->  Nested Loop  (cost=8.56..25.82 rows=9 width=213)
         ->  Nested Loop  (cost=8.42..23.87 rows=10 width=129)
               ->  Limit  (cost=8.28..14.55 rows=10 width=26)
                     InitPlan 1 (returns $1)
                       ->  GroupAggregate  (cost=0.28..8.13 rows=7 width=8)
                             Group Key: r.project_id
                             ->  Nested Loop  (cost=0.28..8.03 rows=7 width=8)
                                   ->  Index Scan using projects_on_site_id_and_id on projects p  (cost=0.14..3.15 rows=1 width=4)
                                         Index Cond: (site_id = 1)
                                   ->  Index Only Scan using revisions_on_project_id_and_id on revisions r  (cost=0.14..4.80 rows=7 width=8)
                                         Index Cond: (project_id = p.id)
                     ->  Index Scan using workflow_definitions_pkey on workflow_definitions wf  (cost=0.14..12.68 rows=20 width=26)
                           Index Cond: (id > 10)
                           Filter: (revision_id = ANY ($1))
               ->  Index Scan using revisions_pkey on revisions rev  (cost=0.14..0.91 rows=1 width=107)
                     Index Cond: (id = wf.revision_id)
         ->  Index Scan using projects_pkey on projects proj  (cost=0.14..0.18 rows=1 width=88)
               Index Cond: (id = rev.project_id)
   ->  Index Scan using workflow_configs_pkey on workflow_configs wc  (cost=0.14..0.91 rows=1 width=159)
         Index Cond: (id = wf.config_id)
```

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseProjectStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseProjectStoreManager.java
@@ -579,6 +579,7 @@ public class DatabaseProjectStoreManager
                         " where p.site_id = :siteId" +
                         " and p.deleted_at is null" +
                         " group by r.project_id" +
+                    " )) " +
                     " and wf.id > :lastId" +
                     " order by wf.id" +
                     " limit :limit" +
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseProjectStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseProjectStoreManager.java
@@ -536,13 +536,12 @@ public class DatabaseProjectStoreManager
                 " join (" +
                     // list id of active (non-deleted) latest revisions in this site.
                     // this is not efficient but gave up optimization on h2.
-                    "select project_id, max(id) as revision_id from revisions" +
-                    " where project_id in (" +
-                        "select id from projects" +
-                        " where site_id = :siteId" +
-                        " and deleted_at is null" +
-                        ")" +
-                    " group by project_id" +
+                    "select r.project_id, max(r.id) as revision_id" +
+                    " from revisions r" +
+                    " join projects p on r.project_id = p.id" +
+                    " where p.site_id = :siteId" +
+                    " and p.deleted_at is null" +
+                    " group by r.project_id" +
                 ") a on wd.revision_id = a.revision_id" +
                 " join revisions rev on a.revision_id = rev.id" +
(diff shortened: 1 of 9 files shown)


frsyuki:
@danielnorberg PTAL

digdag-core/src/main/java/io/digdag/core/database/DatabaseProjectStoreManager.java
```diff
@@ -533,6 +563,31 @@ public void deleteSchedules(int projId)
...
+                    // order by id and limit before join
+                    "select * from workflow_definitions wf" +
+                    " where wf.revision_id = any(array(" +
+                        // list id of active (non-deleted) latest revisions in this site
+                        "select max(r.id)" +
+                        " from revisions r" +
+                        " join projects p on r.project_id = p.id" +
+                        " where p.site_id = :siteId" +
+                        " and p.deleted_at is null" +
+                        " group by r.project_id" +
```
danielnorberg: missing `))` here

frsyuki: oops! Thank you for fixing it.

PR #322, opened 2016-09-30
move project compilation to server

danielnorberg:
Moving the project archive metadata compilation to the server makes it easier for the console to implement workflow editing.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/client/Archive.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/client/Archive.java
@@ -86,14 +86,11 @@ public class Archive
         ConfigFactory cf = injector.getInstance(ConfigFactory.class);
         ConfigLoaderManager loader = injector.getInstance(ConfigLoaderManager.class);
 
-        // read parameters
-        Config overrideParams = loadParams(cf, loader, loadSystemProperties(), paramsFile, params);
-
         // load project
         Path projectPath = (projectDirName == null) ?
             Paths.get("").toAbsolutePath() :
             Paths.get(projectDirName).normalize().toAbsolutePath();
-        injector.getInstance(Archiver.class).createArchive(projectPath, Paths.get(output), overrideParams);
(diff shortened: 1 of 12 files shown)


frsyuki:
nice improvements. gogo

digdag-cli/src/main/java/io/digdag/cli/client/Archive.java
```diff
@@ -86,14 +86,11 @@ private void archive(Injector injector)
         ConfigFactory cf = injector.getInstance(ConfigFactory.class);
         ConfigLoaderManager loader = injector.getInstance(ConfigLoaderManager.class);
 
-        // read parameters
-        Config overrideParams = loadParams(cf, loader, loadSystemProperties(), paramsFile, params);
```
frsyuki: Two fields (`params` and `paramsFile `) should be removed now?

digdag-cli/src/main/java/io/digdag/cli/client/Push.java
```diff
@@ -36,12 +36,6 @@
     @Parameter(names = {"--project"})
     String projectDirName = null;
 
-    @DynamicParameter(names = {"-p", "--param"})
-    Map<String, String> params = new HashMap<>();
-
-    @Parameter(names = {"-P", "--params-file"})
-    String paramsFile = null;
```
frsyuki: please also update usage() method

digdag-core/src/main/java/io/digdag/core/archive/ProjectArchive.java
```diff
@@ -60,7 +60,7 @@ public void listFiles(PathConsumer consumer)
     }
 
     // reused by ProjectArchiveLoader.load
-    static void listFiles(Path projectPath, PathConsumer consumer)
+    public static void listFiles(Path projectPath, PathConsumer consumer)
```
frsyuki: should this method be public?

PR #323, opened 2016-09-30
[WIP] Changing REST API id to String

frsyuki:
working in progress...

Advices by @cesarandreu:

* string ids let you treat ids as opaque identifiers
* a common strategy is to use hashids: http://hashids.org/
* database still has incrementing numeric ids, but it exposes that as a string id which means you don't expose how many of X resource have been created; e.g. you can see how many users have been created, so you can get an idea of the service's size
* string id -> hash id is possible later

I agree with his idea. We can think string formats later. Related article: [Sharding & IDs at Instagram](http://instagram-engineering.tumblr.com/post/10853187575/sharding-ids-at-instagram). First goal is to just convert numbers to strings so that later changes don't break backward compatibility of REST API.

Another advice by @kyprifog is that REST APIs returning an array should wrap it in an object like `{"projects": [...]}`:

```
I want to make as part of server side pagination is to always wrap collection objects, it makes expectations of errors or pagination easier
{error: "Bad stuff", status: 422}
{jobs: [{...},{...}], page_info: {page: 1, next_page: ...}}
is much more consistent because you can always assume the underlying data type is hash
```

I think this also makes sense. Returning an array wrapped in an object also lets us add more information in the object without breaking backward compatibility.


danielnorberg:
This will break the console. Perhaps bake this into api v2 as per https://github.com/treasure-data/digdag/pull/333#issuecomment-251852487 ?

PR #324, opened 2016-10-01
[WIP] Allow ${...} syntax return objects and arrays

frsyuki:
Purpose of this PR is (was) implementing "Idea 3" of https://github.com/treasure-data/digdag/issues/25#issuecomment-203236773 to allow `key: ${foos}` syntax to put array (or objects) as a value of `key`.

However, I found a problem that `Config.get("key", String.class)` fails if value is an array or object.
 This means that following code fails:

```
_export:
  key:
    nested: v

+t:
  echo>: ${key}  # this is an object
```

But this works:

```
_export:
  key:
    nested: v

+t:
  echo>: foo${key}bar  # this is a string
```

I think this behavior is unexpected.

So, "Idea 4" could be better.

### DIFF
--- a/digdag-client/src/main/java/io/digdag/client/config/Config.java
+++ b/digdag-client/src/main/java/io/digdag/client/config/Config.java
@@ -22,7 +22,7 @@ import static java.util.Locale.ENGLISH;
 
 public class Config
 {
-    protected final ObjectMapper mapper;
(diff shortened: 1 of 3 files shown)


frsyuki:
This didn't work. I created #328 instead.

PR #326, opened 2016-10-03
td: retry on netsplits and server errors

danielnorberg:


digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -31,6 +34,12 @@
 public class TDOperator
         implements Closeable
 {
+    private static final Logger logger = LoggerFactory.getLogger(TDOperator.class);
+
+    private static final int INITIAL_RETRY_WAIT = 100;
```
komamitsu: This initial retry interval seems a bit too short to me. How about 500?

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -252,42 +262,99 @@ public TDJobOperator runJob(Config state, String key, JobStarter starter)
...
+            catch (TDClientException e) {
+                logger.warn("failed to start job: domainKey={}", domainKey.get(), e);
+                if (isDeterministicClientException(e)) {
+                    throw e;
+                }
+                throw errorPollingException(state, key, jobState);
+            }
+
+            // Reset error polling
+            jobState = jobState.withErrorPollIteration(Optional.absent());
```
komamitsu: If you add `jobState.resetErrorPollIteration()` and call it here, it's easier to understand for me.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
danielnorberg: This is only for the initial ephemeral netsplit handling.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
danielnorberg: jobState is immutable

danielnorberg: I guess we could use a mutable builder instead, but I'm not sure I'd prefer that...

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
komamitsu: [minor] Yeah, but I have a little concern that this short retry wait might make the api server busier when td-api is already busy and unstable.

komamitsu:
LGTM with awesome tests 👍  although I felt the initial retry interval for td-api a bit too short.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
danielnorberg: Per discussion on slack, will change this to two retries with 300 ms initial wait.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
danielnorberg: Builder doesn't have getters. There's a new `@Value.Modifiable` annotation we could use to generate a value class with setters, but when jackson serializing an object of that class it includes at least one unwanted field. Will stick with the immutable value class for now.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
danielnorberg: Actually, as some operators still use the retry executor directly without the persistent retry that this PR implements, I'll revert this to the original 3 retries with 500ms initial wait for now. We can change this later when all operators have been state-machinified.

PR #327, opened 2016-10-03
Return 404 Not Found when http method name is unknown

frsyuki:
Before:

* Path is under /api (handled by RESTeasy's HttpServlet30Dispatcher)

  * If a method for the path is NOT available: 404 Not Found

  * If a method for the path is available:

    * if `@GET`, etc. annotation is same with the requested method: 200 OK

    * otherwise: 405 Method Not Allowed

* Path is not under /api

  * Resource is always unavailable (no servlets are mapped to this path)

  * If method is either of GET, HEAD, POST, PUT, DELETE, OPTIONS, or
    TRACE: 404 Not Found
    (handled by Undertow's DefaultServlet which extends
     javax.servlet.http.HttpServlet)

  * otherwise: 501 Not Implemented

This change maps a new servlet to /api that returns 404 Not Found for
any requesting methods so that behavior becomes consistent with paths
under /api.

javax.servlet.http.HttpServlet was returning 501 Not Implemented: http://www.docjar.com/html/api/javax/servlet/http/HttpServlet.java.html

### DIFF
--- a/digdag-guice-rs/src/main/java/io/digdag/guice/rs/GuiceRsServletContainerInitializer.java
+++ b/digdag-guice-rs/src/main/java/io/digdag/guice/rs/GuiceRsServletContainerInitializer.java
@@ -2,6 +2,7 @@ package io.digdag.guice.rs;
 
 import java.util.Map;
 import java.util.Set;
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


danielnorberg:
👍

PR #329, opened 2016-10-04
handle space in cli path

danielnorberg:
Verified that this works on windows with paths containing a space.


danielnorberg:
@frsyuki PTAL

frsyuki:
👍

PR #330, opened 2016-10-04
cli: windows: fix selfupdate

danielnorberg:
On windows the file suffix must be `.bat`. Mac OS X and Linux do not care.


danielnorberg:
@frsyuki PTAL

frsyuki:
👍

PR #331, opened 2016-10-04
Set timeout on HTTP connections

frsyuki:
Digdag servers behind load balancers usually have protection against too slow clients (e.g. AWS ELB has [60 seconds idle timeout](https://aws.amazon.com/blogs/aws/elb-idle-timeout-control/)).
This PR is useful when digdag is used without load balancers, or to be conservatively protected against accidental operations, or we want more flexible control on timeout behavior.


danielnorberg:
👍

PR #333, opened 2016-10-04
[WIP] Collection-returning REST API wrap list in an object

frsyuki:
See #323 for discussion.
This assumes #323.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/client/Backfill.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/client/Backfill.java
@@ -92,7 +92,8 @@ public class Backfill
                 truncatedTime.getSessionTime().toInstant(),
                 retryAttemptName,
                 Optional.fromNullable(count),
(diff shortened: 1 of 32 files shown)


danielnorberg:
Nice but this will break the UI. How about adding api v2?

I.e.: 

1. Add `/api/v2/...` endpoints returning these wrapped collections.
2. Keep v1 endpoints while migrating the console.
3. Remove v1 endpoints.

frsyuki:
That is good idea but I'm also afraid of having too many duplicated code. Let me try...
I think this collection change is ok because legacy v1 API can be a wrapper of v2. But ID change could be difficult because it needs to duplicate all rest model classes.

danielnorberg:
Yeah, I wouldn't want to have this code duplicated either. I'd remove v1 as as soon as possible after migrating the console.

And yes, all the rest model classes would also need to be duplicated.

frsyuki:
See #407.

PR #336, opened 2016-10-06
Add API for operators to reset variables before merging

frsyuki:
This solves following unexpected behavior:

* Task +run1 stores td.last_results = `{a: "A1", b: "B"}`
* Task +run2 stores td.last_results = `{a: "A2", d: "D"}`
* Following tasks see td.last_results = `{a: "A2", b: "B", d: "D"}`
* But expected result is td.last_results = `{a: "A2", d: "D"}`

My first idea was to let operators declare list of output parameters (like `return asList("td.last_results")`) so that digdag core (OperatorManager) takes care of resetting parameters. However, this PR forces operators to use `taskResultBuilder.resetStoreParameters(asList(ConfigKey.of("td.last_results")))`. This is because I couldn't find good API design that is tolerant of future extensions...

ConfigKey uses dot-separated format to represent nested keys. I also considered JSONPath, ObjectPath, and RFC 6901 / JSR 374 JSON Pointer. I think dot-separated format is the best because users are already familiar with it because of secret access format and JavaScript's field access syntax.


frsyuki:
@danielnorberg Can I ask you to review this?

danielnorberg:
nice 👍

PR #340, opened 2016-10-06
Add access to secrets to scripting operators

frsyuki:
Users can grant access to secret parameters to scripting operators (sh>, py>, rb>) by explicitly using new `_env` directive. Secrets are set as environment variables.

Operators access to the variables through `PrivilegedVariables` interface. Operators don't need to declare access to the secrets in `Operator.secretSelectors()` method because PrivilegedVariables are supposed to be privileged that are granted by the config explicitly as like `_secrets` directive. This also means that operators can access to secrets only when the operator statically declare them in advance (*), or workflow config explicitly allows them dynamically.

(\* to achieve this, another pull-request is necessary that moves Operator.secretSelectors method to OperatorFactory: #344).

`_env` must be local config at this moment. Setting `_env` in `_export` or store params don't take effect. This is because I'm expecting update of security model of `_env` in the future and it shouldn't break working workflow definitions. The update is about this security concern: 

```
+my_task:
  untrusted_3rd_party_plugin>: do something
```

When this `untrusted_3rd_party_plugin` runs, it can generate following tasks without any permission from users:

```
+my_task^sub+generated_by_untrusted_plugin_get:
  _env:
    IT: {secret: my.secure.data}
  sh>: echo "$IT" > tmp.txt

# same thing happens with previous _secret + _env model
+my_task^sub+generated_by_untrusted_plugin_get:
  _secret:
    my.secure.data: true
  _env:
    IT: {secret: my.secure.data}
  sh>: echo "$IT" > tmp.txt

+my_task^sub+generated_by_untrusted_plugin_use:
  untrusted_3rd_party_plugin>: use tmp.txt
```

This issue happens when `sh>` operator is trusted (trusted by secret access policy) even if `untrusted_3rd_party_plugin` has no permission to access secrets. Currently, this concern is out of this PR's focus excepting that `_env` must be in local config to be prepared for future changes. Future changes may be something that restricts access to secrets from generated subtasks by the permissions of the generating task. This idea may also need some changes to the plugin system so that users (not system administrators) can grant access to secrets per operator in a workflow.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -291,12 +291,6 @@ public class OperatorManager
 
         SecretStore secretStore = secretStoreManager.getSecretStore(mergedRequest.getSiteId());
 
-        PrivilegedVariables privilegedVariables = GrantedPrivilegedVariables.build(
-                // _env must be local config
-                mergedRequest.getLocalConfig().getNestedOrGetEmpty("_env"),
-                mergedRequest.getConfig(),
-                GrantedPrivilegedVariables.privilegedSecretProvider(secretStore, mergedRequest.getProjectId()));
-
         Config grants = mergedRequest.getConfig().getNestedOrGetEmpty("_secrets");
 
         SecretFilter operatorSecretFilter = SecretFilter.of(
@@ -313,6 +307,11 @@ public class OperatorManager
 
         DefaultSecretProvider secretProvider = new DefaultSecretProvider(
                 secretContext, secretAccessPolicy, grants, operatorSecretFilter, secretStore);
+
+        PrivilegedVariables privilegedVariables = GrantedPrivilegedVariables.build(
+                mergedRequest.getConfig().getNestedOrGetEmpty("_env"),
+                mergedRequest.getConfig(),
+                GrantedPrivilegedVariables.privilegedSecretProvider(secretContext, secretAccessPolicy, secretStore));
 
         TaskExecutionContext taskExecutionContext = new DefaultTaskExecutionContext(
                 privilegedVariables, secretProvider);
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -309,7 +309,7 @@ public class OperatorManager
                 secretContext, secretAccessPolicy, grants, operatorSecretFilter, secretStore);
 
         PrivilegedVariables privilegedVariables = GrantedPrivilegedVariables.build(
-                mergedRequest.getConfig().getNestedOrGetEmpty("_env"),
+                mergedRequest.getLocalConfig().getNestedOrGetEmpty("_env"),
                 mergedRequest.getConfig(),
                 GrantedPrivilegedVariables.privilegedSecretProvider(secretContext, secretAccessPolicy, secretStore));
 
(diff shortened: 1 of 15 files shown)


frsyuki:
@danielnorberg could you take a look?
I used `_env` because I can't imagine use cases of PrivilegedVariables other than setting environment variables. If there're some, other names such as `_var` might be better.

digdag-core/src/main/java/io/digdag/core/agent/CheckedConfig.java
```diff
@@ -9,50 +9,33 @@
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.digdag.client.config.Config;
 
-public class CheckedConfig
```
danielnorberg: These changes are a revert of the previous commit which was also a revert..?

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -295,7 +308,13 @@ protected TaskResult callExecutor(Path projectPath, String type, TaskRequest mer
         DefaultSecretProvider secretProvider = new DefaultSecretProvider(
                 secretContext, secretAccessPolicy, grants, operatorSecretFilter, secretStore);
 
-        TaskExecutionContext taskExecutionContext = new DefaultTaskExecutionContext(secretProvider);
+        PrivilegedVariables privilegedVariables = GrantedPrivilegedVariables.build(
+                mergedRequest.getLocalConfig().getNestedOrGetEmpty("_env"),
```
danielnorberg: If "_env" is only used for environment variables for e.g. the `sh` operator, then it should only be in the code path of those operators?

digdag-core/src/main/java/io/digdag/core/agent/CheckedConfig.java
frsyuki: previous revert commit was separating this PR from #343 but later I found that this PR actually depends on #343.

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
frsyuki: Maybe, yes. But in current implementation, Operator has total control of starting a new process. To set "_env" to a new process, "_env" needs to go through Operator. And because OperatorManager doesn't know whether the Operator starts a new process or not, OperatorManager needs to pass "_env" to Operator every time. Because of this implementation limitation, _env is passed to Operator every time. Do you think we should change this? Then, an idea is to add `@com.google.common.annotations.Beta` annotation to `OperatorContext.getPrivilegedVariables()`.

danielnorberg: I assumed the purpose `_env` is for setting environment variables of the process spawned by `sh>` I.e. when the user intends to spawn a process. I.e., it only has meaning in the context of `sh>`.

Do you mean that `_env` would be used to control environment variables for e.g. processes spawned to execute a sandboxed non-sh operator? I.e. even when the user intent is not specifically to spawn a process?

frsyuki: I mean that `_env` would be used to provide a way to access secrets for any scripting operators. `py>` and `rb>` operators use scripts written by users. They need "something" to use secrets. And the "something" I found is `_env`.

frsyuki: I assumed that all script processors can use environment variables. This might be wrong but as many systems do (Travis CI, Circle CI, Heroku, Docker, psql command...), using environment variable is a common way to pass credentials to a runtime. I think it is a reasonable assumption.

frsyuki: How important is it restrict scope of this code path only to `sh>` operator? Why not `py>` and `rb>` scripts have chance to use secrets?

frsyuki:
rebased

frsyuki:
@danielnorberg do you have any chance to say LGTM here?

danielnorberg:
👍

frsyuki:
Thank you for your review!

PR #342, opened 2016-10-07
UI: Refactor webpack config

dzannotti:
This refactors entirely the build scripts for UI. The result for end users is backwards compatible.


danielnorberg:
👍  but please have someone who knows web look at it if necessary =)

PR #343, opened 2016-10-07
TaskRequest.getLocalConfig should include evaluated config

frsyuki:
`${...}` was not evaluated before.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/CheckedConfig.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/CheckedConfig.java
@@ -9,48 +9,31 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import io.digdag.client.config.Config;
 
-public class CheckedConfig
+class CheckedConfig
     extends Config
 {
-    private static class HashSetWithAllFlag<T>
-            extends HashSet<T>
+    static class UsedKeysSet
+            extends HashSet<String>
     {
         private boolean all = false;
 
-        public void setAll(boolean v)
+        public void setAllUsed(boolean v)
         {
             this.all = v;
         }
 
-        public boolean isAll()
+        public boolean isAllUsed()
         {
             return all;
         }
     }
 
-    private final HashSetWithAllFlag<String> usedKeys;
+    private final UsedKeysSet usedKeys;
 
-    public CheckedConfig(Config config)
-    {
-        this(config, new HashSetWithAllFlag<>());
-    }
-
-    private CheckedConfig(Config config, HashSetWithAllFlag<String> usedKeys)
+    CheckedConfig(Config config, UsedKeysSet usedKeys)
     {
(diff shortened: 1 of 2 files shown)


frsyuki:
@komamitsu I think that your #339 needs this.

digdag-core/src/main/java/io/digdag/core/agent/CheckedConfig.java
```diff
@@ -9,50 +9,33 @@
...
 {
-    private static class HashSetWithAllFlag<T>
-            extends HashSet<T>
+    static class UsedKeysSet
+            extends HashSet<String>
     {
         private boolean all = false;
 
-        public void setAll(boolean v)
+        public void setAllUsed(boolean v)
```
komamitsu: `public` can be removed

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -234,17 +236,27 @@ private void runWithWorkspace(Path projectPath, TaskRequest request)
...
         }
 
-        CheckedConfig checkedConfig = new CheckedConfig(config);
+        // Now config is evaluated using JavaScript. Copy values from it
+        // to create evaluated localConfig (original localConfig is not
+        // evaluated).
+
+        Config localConfig = config.getFactory().create();
+        for (String localKey : request.getLocalConfig().getKeys()) {
+            localConfig.set(localKey, config.get(localKey, JsonNode.class).deepCopy());
```
komamitsu: What if `localConfig` has `schema: myschema` and `config` has `schema: {"foo": "bar"}`? If I understand correctly, it seems an unexpected value in `config` can be copied to `localConfig`.

How about evaluating `localConfig` using `evalEngine`?

komamitsu:
@frsyuki Sorry for my late reply and thanks! This would be what I really wanted. I left some comments. Could you check them?

digdag-core/src/main/java/io/digdag/core/agent/CheckedConfig.java
frsyuki: 👍

digdag-core/src/main/java/io/digdag/core/agent/CheckedConfig.java
```diff
@@ -62,7 +45,7 @@ public Config deepCopy()
     @Override
     public ObjectNode getInternalObjectNode()
     {
-        this.usedKeys.setAll(true);
+        this.usedKeys.setAllUsed(true);
```
frsyuki: The purpose of `CheckedConfig` class is showing `Did you mean ...?` message when some local config keys are not used. CheckedConfig tracks that which keys are accessed by the operator and which keys are not.
CheckedConfig assumes that all keys are accessed when `getInternalObject` method is called.

frsyuki:
@komamitsu thank you. I replied.

komamitsu:
LGTM! 👍

PR #345, opened 2016-10-10
UI: Add workflows page

dzannotti:
This adds a new homepage for the UI (workflows) as per discussion

screen shot 2016-10-10 at 17 21 20


danielnorberg:
Nice!

Might it be useful to also have a `project` column in the `Workflows` table?

PR #346, opened 2016-10-10
Releasing artifacts to bintray

### LINKED ISSUE #156: Release jar artifacts to bintray for plugin development
Plugins need to depend on io.digdag:digdag-spi package. But now jars are released to a maven repository built on a private S3 bucket. It's not publicly accessible.
Idea here is to move the repository to bintray. Actions:

* [x] Create new organization on bintray
  * https://bintray.com/digdag
* [x] Invite core developers to the organization
* [ ] Update digdag download proxy server (dl.digdag.io) to support bintray. See embulk dl proxy: https://github.com/embulk/plugin-crawler
* [x] Optional: Move all old releases to bintray
* [ ] Update build.gradle to release files to bintray.

frsyuki:
Fixes #156.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,31 +8,19 @@ apply plugin: 'com.github.johnrengelman.shadow'
 apply plugin: 'idea'
 apply plugin: 'maven'
 
-def getGitHash = { ->
-    try {
-        def stdout = new ByteArrayOutputStream()
-        exec {
-            commandLine 'git log -n 1 --pretty=format:%ad-%H --date format-local:%Y%m%dT%H%M%S'.split(' ')
-            environment 'TZ', 'UTC'
-            standardOutput = stdout
-        }
-        return stdout.toString().trim()
-    }
-    catch (Exception ex) {
-        return "${new Date().format("yyyyMMdd'T'HHmmss", TimeZone.getTimeZone("UTC"))}-unknown"
-    }
-}
-def gitHash = getGitHash()
-
 allprojects {
     group = 'io.digdag'
     version = '0.8.18-SNAPSHOT'
 
     ext {
         isSnapshotRelease = version.endsWith('-SNAPSHOT')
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -185,24 +258,24 @@ subprojects {
                     version project.version
                 }
             }
-        }.withXml {
-            project.configurations[JavaPlugin.RUNTIME_CONFIGURATION_NAME].allDependencies.findAll {
```
danielnorberg: Original purpose of this task was to generate a pom for local development use. 

Is the purpose of the change to this task to generate a pom for publishing?

frsyuki: Task of `pom` is still for local development (and validation in `ci/validate.sh`). `pom` directive in `publishing` is for publishing. They share the same code (`pomXmlModifier`) to generate the same `<dependencies>` section.

frsyuki: I added another commit so that all projects use the same version.

danielnorberg:
Given how magic maven `-SNAPSHOT` is, I'm worried that using non-`-SNAPSHOT` versioned poms with intellij when developing might cause intellij etc to use cached artifacts and/or download artifacts instead of rebuilding. For local development purposes I think I'd prefer that the generated poms to use `-SNAPSHOT` when the build.gradle version is `-SNAPSHOT`.

frsyuki:
I updated releasing mechanism of -SNAPSHOT as following:

* If `-PsnapshotVersion=...` is set, uses the given version instead of -SNAPSHOT. Otherwise, keeps using -SNAPSHOT.
* New task `snapshotRelease` runs `./gradlew -PsnapshotVersion=<git> release`.

danielnorberg:
On both travis and circle:

```
fatal: unknown date format format-local:%Y%m%dT%H%M%S
```

I guess the version of git provided is too old.

danielnorberg:
👍

PR #347, opened 2016-10-11
reusable polling retry executor & add td store_last_result download retry

danielnorberg:

PR #348, opened 2016-10-11
s3_wait: handle netsplits & add reusable poller

danielnorberg:
Continuation of #347

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/PollingWaiter.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/PollingWaiter.java
@@ -1,7 +1,7 @@
 package io.digdag.standards.operator;
 
-import autovalue.shaded.com.google.common.common.base.Throwables;
 import com.google.common.base.Optional;
+import com.google.common.base.Throwables;
 import io.digdag.client.config.Config;
 import io.digdag.client.config.ConfigElement;
 import io.digdag.spi.TaskExecutionException;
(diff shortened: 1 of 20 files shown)


danielnorberg:
@komamitsu PTAL

digdag-standards/src/main/java/io/digdag/standards/operator/PollingWaiter.java
```diff
@@ -0,0 +1,120 @@
...
+
+        Config taskState = pollState.getNestedOrSetEmpty(TASK);
+
+        Optional<T> result;
+
+        try {
+            result = f.apply(taskState);
+        }
+        catch (Exception e) {
+            throw Throwables.propagate(e);
```
komamitsu: We don't need to retry when an exception occurs?

digdag-standards/src/main/java/io/digdag/standards/operator/PollingWaiter.java
```diff
@@ -0,0 +1,120 @@
...
+        }
+
+        if (!result.isPresent()) {
```
komamitsu: It seems that this class can be merged with PollingRetryExecutor that has many codes by adding a new predicate that tests a result to replace this line.

digdag-standards/src/main/java/io/digdag/standards/operator/PollingWaiter.java
danielnorberg: My intent was to have the `PollingRetryExecutor` handle retries separately from the `PollingWaiter` and keep waiting and retrying logic and state separate.

komamitsu:
👍

PR #350, opened 2016-10-12
push build image to docker hub

danielnorberg:
Push a build image to docker hub and use that in travis/circle ci instead of building/caching locally to prevent frequent CI failures due to ephemeral image build failures.


danielnorberg:
@komamitsu PTAL

danielnorberg:
@frsyuki PTAL

danielnorberg:
Note that this uses zulu openjdk instead of oracle jdk to avoid licensing issues.

docker/Makefile
```diff
@@ -0,0 +1,18 @@
+GIT_SHOW = $(shell git show --pretty=format:'%H-%ad' --date 'format:%Y%m%dT%H%M%S' | head -n 1)
+GIT_STATUS = $(shell git status --porcelain --ignore-submodules=dirty 2> /dev/null | tail -n1)
```
komamitsu: [trivial] `-n1` -> `-n 1` as well as `GIT_SHOW`

komamitsu:
👍

frsyuki:
LGTM 👍 
A minor comment is that having git hash first and date later is not good when we list images up because sort order becomes random. So having date first is better. That's something I found with #346.

docker/Makefile
frsyuki: I think you can use `git log -n 1` instead of `git show ... | head -n 1`: 
`git log -n 1 --pretty=format:%ad-%H --date format-local:%Y%m%dT%H%M%S`

PR #353, opened 2016-10-17
fix init command on windows

danielnorberg:
Do not use the platform-sensitive `Path` and `Paths` when dealing with java resource paths.

Java resource path delimiters are always `/` regardless of underlying platform file system.


danielnorberg:
@komamitsu PTAL

komamitsu:
👍

PR #357, opened 2016-10-19
lazy git hash

danielnorberg:
Avoid getting git hash when interpreting build.gradle to avoid failing due to old git on travis/circle ci.


danielnorberg:
@frsyuki PTAL

frsyuki:
I see. I was assuming that evaluation of `commandLine "./gradlew", "-PsnapshotVersion=${getGitHash()}", "release"` line happens only when `snapshotRelease` task is invoked but apparently not.

frsyuki:
I merged to `bintray` branch.
Thank you!

PR #361, opened 2016-10-21
http operator

danielnorberg:


danielnorberg:
@komamitsu PTAL

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
```diff
@@ -0,0 +1,389 @@
...
+            ContentResponse response = PollingRetryExecutor.pollingRetryExecutor(state, "request")
+                    .withErrorMessage("HTTP request failed")
+                    .run(s -> execute(request));
+
+            return result(response, storeContent);
+        }
+
+        private LinkedHashMultimap<String, String> headers()
+        {
+            List<JsonNode> entries = params.getListOrEmpty("headers", JsonNode.class);
```
komamitsu: Why don't you use simply Object not Array of 1 element like this?
```
headers:
  Accept-Encoding: gzip
  Connection: keep-alive
```

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
```diff
@@ -0,0 +1,389 @@
...
+                request.content(contentProvider, contentType.orNull());
+            }
+
+            LinkedHashMultimap<String, String> headers = headers();
+            for (Map.Entry<String, String> header : headers.entries()) {
+                request.header(header.getKey(), header.getValue());
+            }
+
+            ContentResponse response = PollingRetryExecutor.pollingRetryExecutor(state, "request")
+                    .withErrorMessage("HTTP request failed")
```
komamitsu: Shouldn't we retry when receiving 408 or 429?

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
```diff
@@ -0,0 +1,389 @@
...
+            Config result = cf.create();
+            Config http = result.getNestedOrSetEmpty("http");
+            http.set("last_status", response.getStatus());
+
+            ImmutableTaskResult.Builder builder = TaskResult.defaultBuilder(request)
+                    .addResetStoreParams(ConfigKey.of("http", "last_status"));
+
+            if (storeContent) {
+                String content = response.getContentAsString();
+                if (content.length() > maxStoredResponseContentSize) {
```
komamitsu: 👍

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
danielnorberg: These are valid headers:

```
headers:
  - Foo: foo-1
  - Bar: bar
  - Foo: foo-2
```

And the http client should preserve header ordering. That is not possible with a simple object.

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
danielnorberg: 👍

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
```diff
@@ -0,0 +1,389 @@
...
+                String value = o.get(name).asText();
+                headers.put(name, value);
+            }
+
+            return headers;
+        }
+
+        private ContentResponse execute(Request req)
+        {
+            logger.debug("Sending HTTP request: {} {}", req.getMethod(), safeUri(req));
```
danielnorberg: 👍

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
```diff
@@ -0,0 +1,389 @@
...
+    private void configureEnvProxy(String scheme, List<ProxyConfiguration.Proxy> proxies)
+    {
+        Optional<ProxyConfig> proxyConfig = Proxies.proxyConfigFromEnv(scheme, env);
+        if (proxyConfig.isPresent()) {
+            ProxyConfig c = proxyConfig.get();
+            proxies.add(new SchemeProxy(scheme, new Origin.Address(c.getHost(), c.getPort()), c.useSSL()));
+        }
+    }
+
+    private class SchemeProxy
```
komamitsu: Could you tell me the intention of this extended HttpProxy?

komamitsu:
@danielnorberg  Cool. I added some comments. Can you take a look at them?

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
komamitsu: 👍  > header ordering

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
danielnorberg: The intent here is to specify separate proxy configurations for `http` and `https` requests.

The vanilla `HttpProxy` does not allow that, hence the subclassing and overriding of the `matches` method.

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
```diff
@@ -215,16 +221,36 @@ else if (HttpStatus.isRedirection(res.getStatus())) {
                 return res;
             }
             else if (HttpStatus.isClientError(res.getStatus())) {
-                // 4xx: The request is invalid for this resource. Fail hard without retrying.
-                throw new TaskExecutionException("HTTP 4XX Client Error: " + requestStatus(req, res), ConfigElement.empty());
+                switch (res.getStatus()) {
+                    case HttpStatus.REQUEST_TIMEOUT_408:
+                    case HttpStatus.TOO_MANY_REQUESTS_429:
+                        // Retry these.
+                        throw new RuntimeException("Failed HTTP request: " + requestStatus(req, res));
```
komamitsu: `ephemeralError()` is better?

digdag-standards/src/main/java/io/digdag/standards/operator/HttpOperatorFactory.java
```diff
@@ -177,13 +180,36 @@ else if (user.isPresent()) {
...
 
             return result(response, storeContent);
         }
 
+        private void configureQueryParameters(Request request)
+        {
+            // TODO: allow secret query parameters
+
+            try {
+                Config queryParameters = params.parseNestedOrGetEmpty("query");
```
komamitsu: [minor] Just `query` can sound SQL's one to me. How about `query_string`? It's more explicit, a bit long though.

danielnorberg: The URI rfc calls it `query`. https://tools.ietf.org/html/rfc3986#section-3

Is the name `query` confusing in the below snippet?

```
+fetch:
  http>: https://foo.example.com/api/baz
  query:
    foo: bar
    baz: quux
```

Is this better?

```
+fetch:
  http>: https://foo.example.com/api/baz
  query_string:
    foo: bar
    baz: quux
```

komamitsu: I see > https://tools.ietf.org/html/rfc3986#section-3

The second one seems clearer to me since I guess `query_string` is more familiar. But either might be fine.

PR #364, opened 2016-10-26
gcs_wait operator

danielnorberg:


danielnorberg:
@komamitsu PTAL

digdag-standards/src/main/java/io/digdag/standards/operator/gcp/GcsWaitOperatorFactory.java
```diff
@@ -0,0 +1,107 @@
...
+
+            return await(gcs, bucket.get(), object.get());
+        }
+
+        private TaskResult await(GcsClient gcs, String bucket, String object)
+        {
+            StorageObject metadata = pollingWaiter(state, "exists")
+                    .withWaitMessage("Object '%s/%s' does not yet exist", bucket, object)
+                    .await(pollState -> pollingRetryExecutor(pollState, "poll")
+                            .run(s -> gcs.stat(bucket, object)));
```
komamitsu: Don't need to skip retrying in case of 4xx?

digdag-tests/src/test/resources/acceptance/gcs_wait/gcs_wait.dig
```diff
@@ -0,0 +1,6 @@
+
++wait:
+  gcs_wait>: ${test_bucket}/${test_object}
```
komamitsu: We'd better add a test that uses `bucket` and `object` commands

komamitsu:
@danielnorberg Overall looks good! I added some comments. Can you check them?

PR #365, opened 2016-10-26
td: fail without retry on *most* 4xx errors

danielnorberg:
Currently we're retrying forever on 422. We could continue blacklisting status codes but that seems error prone.


danielnorberg:
@frsyuki @komamitsu PTAL

komamitsu:
👍

komamitsu:
👍

PR #366, opened 2016-10-27
do not log null operator command

danielnorberg:
Currently `operator>: null` is logged if the command is null. I.e, all of the operator configuration is in parameters.


danielnorberg:
@komamitsu PTAL

komamitsu:
👍

PR #368, opened 2016-10-27
workflow execution timeouts

danielnorberg:
Enforce TTLs for session attempts and tasks.

By default a session attempt is allowed to run for at most 7 days before it is canceled. A task is by default allowed to run for 24h and then the attempt is canceled.

Note: Blocking tasks will currently not be terminated as the attempt cancellation mechanism does not implement forceful task termination. "Cooperative" polling tasks however will be "terminated" simply by not being invoked again.


danielnorberg:
@komamitsu @frsyuki PTAL

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -288,6 +287,38 @@ public boolean isAnyNotDoneAttempts()
...
+    }
+
+    @Override
+    public List<TaskAttemptSummary> findTasksStartedBeforeWithStateAndType(TaskType type, TaskStateCode[] states, Instant startedBefore, long lastId, int limit)
+    {
+        return autoCommit((handle, dao) ->
+                handle.createQuery(
+                    "select id, attempt_id, state" +
+                    " from tasks" +
+                    " where task_type = :type " +
```
frsyuki: How about adding `(state_flags & 1) = 0` condition so that `WorkflowExecutionTimeoutEnforcer.enforceTaskTTLs` doesn't have to check same tasks again and again when an attempt includes a non-polling task which doesn't finish immediately?:

```
... +
// exclude already cancel-requested tasks
" and " + bitAnd("state_flags", Integer.toString(TaskStateFlags.CANCEL_REQUESTED)) + " = 0" +
```

digdag-tests/src/test/java/acceptance/AttemptTimeoutIT.java
```diff
@@ -0,0 +1,111 @@
...
+              "executor.task_ttl = 10s",
+              "executor.ttl_reaping_interval = 1s");
+
+        addWorkflow(projectDir, "acceptance/attempt_timeout/task_timeout.dig");
+        long attemptId = pushAndStart(server.endpoint(), projectDir, "task_timeout");
+
+        // Expect the attempt to get canceled when the task times out
+        expect(Duration.ofMinutes(1), () -> client.getSessionAttempt(attemptId).getCancelRequested());
+
+        // TODO: implement termination of blocking tasks
```
frsyuki: yes, this is a TODO...

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -288,6 +287,38 @@ public boolean isAnyNotDoneAttempts()
     }
 
     @Override
+    public List<StoredSessionAttempt> findAttemptsCreatedBeforeWithState(Instant createdBefore, int state, long lastId, int limit)
```
danielnorberg: I guess even if we use concurrent index creation, digdag will still wait for the index creation operation to complete before bringing up undertow and serving api requests?

If we worry about this index creation taking a long time, perhaps it would be better to do it manually before deploying this PR?

Perhaps we should consider separating database schema migration into a separate deployment step?

I.e.

0. Merge PR
1. Run database schema migration
2. Release & Deploy

As database schema migrations must anyway be backwards compatible in order to avoid digdag downtime, performing the schema migration before deploying the new digdag version should be safe.

Thoughts?

digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
```diff
@@ -0,0 +1,160 @@
...
+            ssm.requestCancelAttempt(attempt.getId());
+        }
+    }
+
+    private void enforceTaskTTLs()
+    {
+        Instant startDeadline = ssm.getStoreTime().minus(taskTTL);
+        long lastId = 0;
+
+        List<TaskAttemptSummary> expiredTasks = ssm.findTasksStartedBeforeWithStateAndType(TaskType.of(0), TaskStateCode.notDoneStates(), startDeadline, lastId, 100);
```
danielnorberg: That was not my intent. Perhaps we should only enforce task TTL's for _leaf_ tasks instead of for all non grouping-only tasks?

danielnorberg: I.e., tasks with no child tasks.

digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
```diff
@@ -0,0 +1,160 @@
...
+            try {
+                stateFlags = ssm.getAttemptStateFlags(task.getAttemptId());
+            }
+            catch (ResourceNotFoundException e) {
+                logger.debug("Session Attempt not found, ignoring: {}", task, e);
+                continue;
+            }
+
+            if (stateFlags.isCancelRequested()) {
+                logger.debug("Session Attempt already canceled, ignoring: {}", task);
```
danielnorberg: Sure.

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
danielnorberg: 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -1473,6 +1504,14 @@ StoredSession upsertAndLockSession(@Bind("projectId") int projectId,
                 " where sa.id = :attemptId limit 1")
         StoredSessionAttemptWithSession getAttemptWithSessionByIdInternal(@Bind("attemptId") long attemptId);
 
+        @SqlQuery("select * from session_attempts" +
+                " where state_flags = :state" +
+                " and created_at < :createdBefore" +
+                " and id > :lastId" +
+                " order by id desc" +
```
danielnorberg: 👍

digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20161110112233_AddStartedAtColumnAndIndexToTasks.java
```diff
@@ -0,0 +1,22 @@
...
+        if (context.isPostgres()) {
+            handle.update("alter table tasks" +
+                    " add column started_at timestamp with time zone");
+        }
+        else {
+            handle.update("alter table tasks" +
+                    " add column started_at timestamp");
+        }
+
+        handle.update("create index tasks_on_task_type_and_state_and_started_at on tasks (task_type, state, started_at, id desc)");
```
danielnorberg: 👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
frsyuki: `CREATE INDEX CONCURRENTLY` is a blocking operation although it doesn't lock a table. So, yes, digdag still waits. Moreover, `CREATE INDEX CONCURRENTLY` can't run in a transaction.

I agree that separation of migration is eventually necessary. I hope that it's not now.

digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
frsyuki: I see. Then, how about excluding `PLANNED` state?: `Arrays.stream(notDoneStates()).filter(state -> state != PLANNED).collect(toList())`
I actually don't know which is better. Making `loop>` operator timed out after `taskTTL` is also a possible option.

frsyuki:
LGTM 👍 
I added https://github.com/treasure-data/digdag/pull/368#discussion_r87946309 but it's not entirely necessary.

digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
danielnorberg: I'll exclude `PLANNED` state, although this is definitely starting to feel quite hacky and fragile.

PR #370, opened 2016-10-28
UI: add retry button

dzannotti:
Adds UI Retry button for failed sessions using the new api `PUT /attempts`


digdag-ui/model.js
```diff
@@ -297,6 +297,15 @@ export class Model {
     })
   }
 
+  retrySession (session: Session) {
+    const { lastAttempt } = session
+    return this.put('attempts', {
```
danielnorberg: I believe `retryAttemptName` must also be set to a unique name. There is a unique constrain on `(session_id, attempt_name)` in the `session_attempts` table.

https://github.com/treasure-data/digdag/blob/master/digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20151204221156_CreateTables.java#L120

The digdag cli generates a unique retry attempt name for the user if one has not been specified.

https://github.com/treasure-data/digdag/blob/master/digdag-cli/src/main/java/io/digdag/cli/client/Retry.java#L123

dzannotti: Yeah that was on purpose, because i wasn't sure which name to use, looking at it now probably this should be set to `wf.name + '-retry'`, what do you think?

dzannotti: in which case why not use the server generated one? (as the code does right now)

danielnorberg: The server doesn't generate a name. The cli does, client side.

frsyuki: I also think client-side should generate an unique id so that retrying submission is idempotent. Just setting a random UUID (UUID v4) to the field is good enough.

dzannotti: @frsyuki upon submission (and api response) the page redirects to `/session` so double click should not be a possible case

dzannotti:
Added uuid as `retryAttemptName`

danielnorberg:
👍 🌮 🍻 🏖

PR #373, opened 2016-11-01
client: check status code when fetching logs

danielnorberg:


danielnorberg:
@komamitsu @frsyuki  PTAL

digdag-client/src/main/java/io/digdag/client/DigdagClient.java
```diff
@@ -569,14 +583,64 @@ public InputStream getLogFile(long attemptId, RestLogFileHandle handle)
...
+
+        return invokeWithRetry(request)
+                .readEntity(InputStream.class);
+    }
+
+    private Response invokeWithRetry(Invocation request)
+    {
+        Retryer<Response> retryer = RetryerBuilder.<Response>newBuilder()
+                .retryIfException(not(DigdagClient::isDeterministicError))
+                .withWaitStrategy(exponentialWait())
```
komamitsu: This library looks cool. Do you think we'd better replace `io.digdag.util.RetryExecutor` with `com.github.rholder.retry.Retryer` in the future?

digdag-client/src/main/java/io/digdag/client/DigdagClient.java
```diff
@@ -569,14 +583,64 @@ public InputStream getLogFile(long attemptId, RestLogFileHandle handle)
...
+                Response res = request.invoke();
+                if (res.getStatusInfo().getFamily() != SUCCESSFUL) {
+                    res.close();
+                    return handleErrorStatus(res);
+                }
+                return res;
+            });
+        }
+        catch (ExecutionException | RetryException e) {
+            Throwable cause = e.getCause() != null ? e.getCause() : e;
```
komamitsu: Hmm.... It looks RetryException includes the last exception not the first one. The first exception's information would be more helpful...

komamitsu:
How about adding some tests of retry over to DigdagClientTest?

komamitsu:
LGTM 👍 
I left some minor comments on the diffs, though.

danielnorberg:
DigdagClientTest already tests retries of log fetches. Do you mean that we need to add additional tests before merging this PR?

digdag-client/src/main/java/io/digdag/client/DigdagClient.java
danielnorberg: Yeah, as far as possible I believe we should try to use existing libraries instead of writing our own.

komamitsu:
I meant I couldn't find any test to see if DigdagClient expectedly throws RetryException after 10 attempts

frsyuki:
👍  Nice improvements!
I created a PR for this PR:

frsyuki:
test was failing due to #419 but it passed once retried.

Issue #377, opened 2016-11-04
digdag-ui: logo.png is in the wrong place


tony810430:
Since digdag-ui changed the web home directory from "digdag-ui/public" to "digdag-ui", the logo.png should move from "digdag-ui/config/" to "digdag-ui/"

PR #378, opened 2016-11-04
digdag-ui: logo.png is in the wrong place

tony810430:
This PR is for #377


dzannotti:
@tony810430 thanks for this PR!, Unfortunately logo is part of customization/branding and therefore i feel that it belongs in `/config` folder, calling @danielnorberg for a second opinion

tony810430:
It's ok to replace the logo path in digdag-ui/config/config.js with "/config/logo.png" instead of just move logo.png out of "config/".

dzannotti:
@tony810430 relooking at this one, is the problem that you are having related to `config.js` referencing a file in the wrong path during development? The config.js distributed is adapt for a final build, not for a development build, could it be the issue?

logo.png is part of the configuration, and therefore i feel it belongs in the config path

tony810430:
Yes, you are right. I only ran this project in development mode.
It looks like the logo.png is in the right directory for a final build. Thanks for your comment.

PR #379, opened 2016-11-04
digdag-ui: list scheduler called incorrect digdag-client api

tony810430:


dzannotti:
hi @tony810430 thank you very much for this PR. Can you please explain a bit more context for this issue? https://github.com/treasure-data/digdag/blob/e6726c07583cf046baea3d5273f42039cb55164b/digdag-server/src/main/java/io/digdag/server/rs/ScheduleResource.java#L45 seems to agree with the current implementation

tony810430:
This API seems  use scheduler's id to query the scheduler's detail, but the function "fetchProjectWorkflowSchedule" wants to query schedule's detail by project id and workflow name. https://github.com/treasure-data/digdag/blob/e6726c07583cf046baea3d5273f42039cb55164b/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java seems to be the API that function needs. Otherwise, the function will get all schedulers back and "pause/resume" button will always trigger the first scheduler only even if the webpage is not in that workflow.

digdag-ui/model.js
```diff
@@ -309,7 +309,7 @@ export class Model {
   }
 
   fetchProjectWorkflowSchedule (projectId: number, workflowName: string): Promise<*> {
-    return this.get(`schedules?project_id=${projectId}&workflow=${workflowName}`)
+    return this.get(`projects/${projectId}/schedules?project_id=${projectId}&workflow=${workflowName}`)
```
dzannotti: `project_id` param is no longer needed as its part of the url as per
https://github.com/treasure-data/digdag/blob/e6726c07583cf046baea3d5273f42039cb55164b/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java#L131

dzannotti:
I see, you are totally right, thanks for spotting this! I left a minor note, if you could change it, would be great.

digdag-ui/model.js
tony810430: ok, i will take it out.

dzannotti:
LGTM, thanks @tony810430

PR #380, opened 2016-11-05
UI: add retry with latest revision button

tony810430:
The retry button is great now, but sometimes user will make mistake on his workflow and need to retry his workflow with latest revision not just the origin revision.
I refered to retry implementation in "digdag-cli" (https://github.com/treasure-data/digdag/blob/2f88fe1d2dcb1fc66e430094c162f2dd89ccc2b3/digdag-cli/src/main/java/io/digdag/cli/client/Retry.java).
It used two api calls to implement "retry with latest revision": the first was to get the latest workflow id by project id and workflow name and replace the failed job's workflow id to call a new attempt request.
Therefore, I implemented "retrySessionWithLatestRevision" function by copying the "retrySession" function and replacing the workflow id with latest workflow id, same as the logic in "digdag-cli"

### DIFF
--- a/digdag-ui/console.jsx
+++ b/digdag-ui/console.jsx
@@ -955,6 +955,14 @@ const SessionView = withRouter(
         .then(() => router.push('/sessions'))
     }
 
+    retrySessionWithLatestRevision () {
+      const { session, router } = this.props
+      this.setState({ loading: true })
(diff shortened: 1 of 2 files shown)


danielnorberg:
👍  Nice!

PR #381, opened 2016-11-07
for_each: index subtask names

danielnorberg:
Instead of including the entire key and value in the names of generated subtasks, use key and value indices. The key and value is still included in the subtask name, but encoded and truncated so as to make it less likely that the task name limit is hit.

This makes it possible to `for_each>` over arrays of larger records.

Another notable consequence of this change is that duplicate values in a `for_each>` parameter array now generates separate subtasks. Previously these duplicate values would generate subtasks with identical names. Thus the below would previously only generate a single subtask.

```
for_each>: {a: [1, 1]}
```

The above will now generate two subtasks, as the user likely expects.


danielnorberg:
@frsyuki @komamitsu PTAL

digdag-standards/src/test/resources/io/digdag/standards/operator/for_each/basic_expected.yml
```diff
@@ -1,6 +1,6 @@
-+for-a=1:
++for-0=a=0=1:
```
komamitsu: Why is the first `0` needed? Isn't it enough with `+for-a=0=1`?

Also, how about using `@` instead of `=` for indices? Using `=` for indices seems confusable with the next `=` for values to me.

danielnorberg: The first `0` is index of the key. The second `0` is the index of the value. Both together unambiguously identify a key/value pair.

danielnorberg: `@` does seem to be allowed in task names. https://github.com/treasure-data/digdag/blob/master/digdag-core/src/main/java/io/digdag/core/repository/ModelValidator.java#L26

Is `+for-0@a=1@2&1@b=1@4:` less confusing than `+for-0=a=1=2&1=b=1=4:` ?

komamitsu: Yeah. Using `=` as both an index and the delimiter of key and value doesn't look easy to understand to me although it might be a just matter of preference. But on second thought, the human readability of task names generated by `for_each` won't be not so important. So we can keep it as is.

komamitsu:
👍

PR #382, opened 2016-11-07
UI: fetch log file by attemptId and file.fileName when file.direct is not exist

tony810430:
The 'digdag-ui' only fetched log file when file.direct was exist.
When file.direct is not exist, the return value will be `Cannot fetch non-direct log file: ${file.fileName}`.
However, the implementation in 'digdag-cli' will fetch log by attemptId and file.fileName when there is no file.direct. (see line 556 to 580 https://github.com/treasure-data/digdag/blob/7eba3480841f0c6a272bf44de6acb201187c51b3/digdag-client/src/main/java/io/digdag/client/DigdagClient.java)
I added one more parameter(attemptId) to LogFileView and generated a new file.direct by attemptId and file.fileName when file.direct is not exist.

### DIFF
--- a/digdag-ui/model.js
+++ b/digdag-ui/model.js
@@ -252,12 +252,11 @@ export class Model {
     return this.get(`logs/${attemptId}/files?task=${encodeURIComponent(taskName)}`)
   }
 
-  fetchLogFile (file: LogFileHandle) {
+  fetchLogFile (attemptId: number, file: LogFileHandle) {
     if (!file.direct) {
-      return new Promise((resolve, reject) => {
-        reject(`Cannot fetch non-direct log file: ${file.fileName}`)
-      })
+      file.direct = this.config.url + `logs/${attemptId}/files/${file.fileName}`
     }
+
     return fetch(file.direct).then(response => {
       if (!response.ok) {
         throw new Error(response.statusText)
--- a/digdag-ui/model.js
+++ b/digdag-ui/model.js
@@ -254,15 +254,9 @@ export class Model {
 
   fetchLogFile (attemptId: number, file: LogFileHandle) {
     if (!file.direct) {
-      file.direct = this.config.url + `logs/${attemptId}/files/${file.fileName}`
+      return this.fetchArrayBuffer(`${this.config.url}logs/${attemptId}/files/${file.fileName}`)
     }
-
-    return fetch(file.direct).then(response => {
-      if (!response.ok) {
-        throw new Error(response.statusText)
-      }
-      return response.arrayBuffer()
-    })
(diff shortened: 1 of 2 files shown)


digdag-ui/model.js
```diff
@@ -252,12 +252,11 @@ export class Model {
     return this.get(`logs/${attemptId}/files?task=${encodeURIComponent(taskName)}`)
   }
 
-  fetchLogFile (file: LogFileHandle) {
+  fetchLogFile (attemptId: number, file: LogFileHandle) {
     if (!file.direct) {
-      return new Promise((resolve, reject) => {
-        reject(`Cannot fetch non-direct log file: ${file.fileName}`)
-      })
+      file.direct = this.config.url + `logs/${attemptId}/files/${file.fileName}`
```
dzannotti: I don't like the idea of overwriting `files.direct`, what do you think of something like this?

```
diff --git a/digdag-ui/model.js b/digdag-ui/model.js
index f0e14d1..0067562 100644
--- a/digdag-ui/model.js
+++ b/digdag-ui/model.js
@@ -252,18 +252,11 @@ export class Model {
     return this.get(`logs/${attemptId}/files?task=${encodeURIComponent(taskName)}`)
   }
 
-  fetchLogFile (file: LogFileHandle) {
+  fetchLogFile (file: LogFileHandle, attemptId: number) {
     if (!file.direct) {
-      return new Promise((resolve, reject) => {
-        reject(`Cannot fetch non-direct log file: ${file.fileName}`)
-      })
+      return this.fetchArrayBuffer(`${this.config.url}logs/${attemptId}/files/${file.fileName}`)
     }
-    return fetch(file.direct).then(response => {
-      if (!response.ok) {
-        throw new Error(response.statusText)
-      }
-      return response.arrayBuffer()
-    })
+    return this.fetchArrayBuffer(file.url)
   }
 
   fetchProjectArchiveLatest (projectId: number): Promise<ProjectArchive> {
@@ -387,6 +380,15 @@ export class Model {
   headers (): Headers {
     return this.config.headers({credentials: this.config.credentials})
   }
+
+  fetchArrayBuffer (url: string) {
+    return fetch(url).then(response => {
+      if (!response.ok) {
+        throw new Error(response.statusText)
+      }
+      return response.arrayBuffer()
+    })
+  }
 }
 
 var instance: ?Model = null
```

dzannotti:
This is nice! 👍
Left a nit comment on the code, thanks tony!

digdag-ui/model.js
tony810430: That's great to me. I'll refine it. Thank you.

PR #384, opened 2016-11-08
UI: sort log files by fileTime

tony810430:
When there were more than one log files, the log wouldn't be arranged in order.


dzannotti:
@tony810430 awesome!

PR #385, opened 2016-11-08
admin endpoint for getting attempt user info

danielnorberg:
* Add admin api port, for exposing administration endpoints that should not be reachable on the regular api port.

* Add `admin` user concept to `Authenticator.Result` interface, letting `Authenticator` implementations indicate whether an authenticated request is made by an admin user.

* Add `/api/admin/attempts/{id}/userinfo` endpoint, available only on the admin port and usable only by `admin` users.


danielnorberg:
@frsyuki @komamitsu PTAL

digdag-guice-rs-server/src/main/java/io/digdag/guice/rs/server/ServerBootstrap.java
```diff
@@ -2,6 +2,9 @@
 
 import org.embulk.guice.Bootstrap;
 
+import java.net.InetSocketAddress;
+import java.util.List;
```
frsyuki: these 2 lines are unnecessary?

danielnorberg:
`AdminResource` is annotated with `@AdminRestricted`, which causes all requests matched to `AdminResource` methods to be filtered by`AdminRestrictedFilter`, which in turn ensures that only requests received on the admin port are let through. Requests received on the regular port are rejected with a `404 Not Found`. 

The `AdminIT` test verifies that a client connected to the regular port is unable to reach the only admin endpoint.

digdag-guice-rs-server/src/main/java/io/digdag/guice/rs/server/ServerBootstrap.java
danielnorberg: 👍

PR #386, opened 2016-11-09
UI: Replace PrismJS with AceEditor

dzannotti:
This allows us to maintain only one version in sync with console


danielnorberg:
👍

PR #387, opened 2016-11-10
td_table_export: fix secret key access

danielnorberg:
Make the fallback from `aws.s3.*` to `aws.*` lazy in order to allow keys under `aws.s3.*` to be used.


komamitsu:
We don't need to modify the tests of this operator to use `aws.s3.*`?

danielnorberg:
Doesn't seem like there are any tests for this operator =/

komamitsu:
👍

PR #388, opened 2016-11-11
use msgpack 0.8.11

danielnorberg:
We have seen issues with reading numbers producing incorrect values when downloading job results from td api using msgpack 0.8.2.


komamitsu:
👍

PR #397, opened 2016-11-25
emr operator

danielnorberg:


digdag-docs/src/operators.rst
```diff
@@ -2031,3 +2031,176 @@ Parameters
...
+          query: queries/query.sql
+          result: s3://my-bucket/results/${session_uuid}/
+        - type: script
+          script: scripts/hello.sh
+          args: [hello, world]
+
+Secrets
+~~~~~~~
+
+:command:`aws.emr.access-key-id, aws.access-key-id`
```
komamitsu: According to the implementation of EmrOperatorFactory.EmrOperator#credentials, `access_key_id` is correct? We don't convert `-` with `_` in secrets, right?

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1739 @@
...
+
+            AmazonElasticMapReduce emr = new AmazonElasticMapReduceClient(credentials, emrClientConfiguration);
+            AmazonS3Client s3 = new AmazonS3Client(credentials, s3ClientConfiguration);
+            AWSKMSClient kms = new AWSKMSClient(credentials, kmsClientConfiguration);
+
+            // Set up file stager
+            Optional<AmazonS3URI> staging = params.getOptional("staging", String.class).transform(s -> {
+                AmazonS3URI uri = new AmazonS3URI(s);
+                if (uri.getKey() != null && !uri.getKey().endsWith("/")) {
+                    throw new ConfigException("Invalid staging uri: '" + s + "'");
```
komamitsu: Don't we need to add something like "The URI should have a tailing slash" here or on the document?

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1739 @@
...
+                    if (file.template()) {
+                        try {
+                            String content = workspace.templateFile(templateEngine, reference.filename(), UTF_8, params);
+                            byte[] bytes = content.getBytes(UTF_8);
+                            ObjectMetadata metadata = new ObjectMetadata();
+                            metadata.setContentLength(bytes.length);
+                            return new PutObjectRequest(uri.getBucket(), uri.getKey(), new ByteArrayInputStream(bytes), metadata);
+                        }
+                        catch (IOException | TemplateException e) {
+                            throw new TaskExecutionException(e, TaskExecutionException.buildExceptionErrorConfig(e));
```
komamitsu: In case of TemplateException, it sounds there is something wrong with configuration and we'd better use ConfigException instead?

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1739 @@
...
+                        }
+                    }
+                    else {
+                        return new PutObjectRequest(uri.getBucket(), uri.getKey(), workspace.getFile(reference.filename()));
+                    }
+                }
+                case RESOURCE: {
+                    byte[] bytes;
+                    try {
+                        bytes = Resources.toByteArray(new URL(reference.reference().get()));
```
komamitsu: [trivial] Calling `reference.reference().get()` without `referene.isPresent()` depends on the implementation that the types except for `DIRECT` have `Optional.of(reference)`. Maybe we can refactor `RemoteFile` later with class polymorphism.

Like this?
```
abstract class RemoteFile
{
    private final AmazonS3URI uri;

    protected RemoteFile(AmazonS3URI uri)
    {
        this.uri = uri;
    }

    public AmazonS3URI getUri()
    {
        return uri;
    }

    public abstract PutObjectRequest createUploadRequest();
}

class RemoteFileWithReference
    extends RemoteFile
{
    private final FileReference reference;

    public RemoteFileWithReference(AmazonS3URI uri, FileReference reference)
    {
        super(uri);
        this.reference = reference;
    }
}

class RemoteFileWithLocalFile
    extends RemoteFileWithReference
{
    public PutObjectRequest createUploadRequest()
    {
        :
    }
}

class RemoteFileWithDirectUpload
    extends RemoteFile
{
    private final byte[] contents;

    public RemoteFileWithDirectUpload(AmazonS3URI uri, byte[] contents)
    {
        super(uri);
        this.contents = contents;
    }

    public PutObjectRequest createUploadRequest()
    {
        :
    }
}
```

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1739 @@
...
+                    .withPrefix(staging.get().getKey());
+            do {
+                ObjectListing res = s3.listObjects(req);
+                String[] keys = res.getObjectSummaries().stream()
+                        .map(S3ObjectSummary::getKey)
+                        .toArray(String[]::new);
+                for (String key : keys) {
+                    logger.info("Removing s3://{}/{}", bucket, key);
+                }
+                s3.deleteObjects(new DeleteObjectsRequest(bucket).withKeys(keys));
```
komamitsu: When `s3.deleteObjects()` throws an exception, we should continue to delete the rest of files? Hmm, but it might cause infinite loop... Maybe we'd better leave here as is.

danielnorberg: 👍

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
danielnorberg: Actually, I'll remove the requirement for the URI to have a trailing slash.

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
danielnorberg: 👍

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1753 @@
...
+                    .withDurationSeconds(3600)
+                    .withRoleSessionName(roleSessionName));
+
+            return new BasicSessionCredentials(
+                    assumeResult.getCredentials().getAccessKeyId(),
+                    assumeResult.getCredentials().getSecretAccessKey(),
+                    assumeResult.getCredentials().getSessionToken());
+        }
+
+        private TaskResult run(String tag, AmazonElasticMapReduce emr, AmazonS3Client s3, AWSKMSClient kms, Filer filer)
```
komamitsu: [minor] `s3` isn't used

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1753 @@
...
+                    return Optional.absent();
+
+                case "RUNNING":
+                case "WAITING":
+                    logger.info("EMR cluster up: {}", cluster.id());
+                    return Optional.of(clusterState);
+
+                case "TERMINATED_WITH_ERRORS":
+                    if (createOnly) {
+                        // TODO: log more information about the errors
```
komamitsu: 👍

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1753 @@
...
+        {
+
+            return () -> {
+                // Start cluster
+                NewCluster cluster = pollingRetryExecutor(state, "submission")
+                        .withRetryInterval(DurationInterval.of(Duration.ofSeconds(30), Duration.ofMinutes(5)))
+                        // TODO: EMR requests are not idempotent, thus retrying might produce duplicate cluster submissions.
+                        .retryUnless(AmazonServiceException.class, Aws::isDeterministicException)
+                        .runOnce(NewCluster.class, s -> submitNewClusterRequest(emr, tag, stepCompiler, clusterConfig, filer, parameterCompiler));
+
```
komamitsu: Where do you store a new cluster id in a state? It seems the new cluster id should be stored here. Without that, it looks like another cluster will be launched again to me... CMIIW.

komamitsu: Oops, I read wrongly the handling `cluster` keyword in `run()` method...

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1753 @@
...
+    private static class Filer
+    {
+        private final AmazonS3Client s3;
+        private final Optional<AmazonS3URI> staging;
+        private final Workspace workspace;
+        private final TemplateEngine templateEngine;
+        private final Config params;
+
+        private final List<StagingFile> files = new ArrayList<>();
```
komamitsu: The order of files doesn't matter? If so, using Set migit be a bit better for de-duplication of files.

digdag-docs/src/operators.rst
```diff
@@ -2031,3 +2031,176 @@ Parameters
...
+        script: scripts/hello.sh
+        args: [world]
+
+      - type: command
+        command: echo
+        args: [hello, world]
+
+:command:`action_on_failure: TERMINATE_JOB_FLOW | TERMINATE_CLUSTER | CANCEL_AND_WAIT | CONTINUE`
+  The action EMR should take in response to a job step failing.
+
```
komamitsu: What about `configurations`?

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1753 @@
...
+
+        List<StepConfig> stepConfigs()
+        {
+            Preconditions.checkState(configs != null);
+            return configs;
+        }
+
+        private String localWorkingDirectory()
+        {
+            return LOCAL_STAGING_DIR + "/" + tag + "/steps/" + index + "/wd";
```
komamitsu: [minor] Maybe passing `index` as an argument and removing `index` instance variable would be better in terms of maintainability

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
danielnorberg: There's currently no deduplication implemented. It's just a collection of files. Changing this to Set would make it seem like there is a StagingFile identity and some kind of deduplication implemented even though there is none.

Perhaps deduplication would be a desirable feature, but I don't think it is necessary as staged files should not be large. I'm not sure the added complexity would be worth it.

 If it becomes necessary in the future, then we can implement it in a separate PR.

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
danielnorberg: The cluster ID is stored in the state by `pollingRetryExecutor(state, "submission")` `runOnce`: https://github.com/treasure-data/digdag/blob/74691948f7cb3cc6af25ca406a287a13bdf09aff/digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java#L466

So if `newClusterSubmitter` is re-entered due to bootstrap taking a long time (which it does), the submission is not performed again.

digdag-docs/src/operators.rst
danielnorberg: Will add 👍

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
```diff
@@ -0,0 +1,1753 @@
...
+            String secretAccessKey = emrSecrets.getSecretOptional("secret_access_key")
+                    .or(() -> awsSecrets.getSecret("secret_access_key"));
+
+            AWSCredentials credentials = new BasicAWSCredentials(accessKeyId, secretAccessKey);
+
+            Optional<String> roleArn = emrSecrets.getSecretOptional("role_arn")
+                    .or(awsSecrets.getSecretOptional("role_arn"));
+
+            if (!roleArn.isPresent()) {
+                return credentials;
```
komamitsu: Maybe we'd better use GetFederationToken to use temporary secrets unless using AssumeRole. http://docs.aws.amazon.com/AmazonS3/latest/dev/AuthUsingTempFederationTokenJava.html

digdag-standards/src/main/resources/io/digdag/standards/operator/aws/spark-sql-wrapper.py
```diff
@@ -0,0 +1,30 @@
...
+spark = SparkSession\
+    .builder\
+    .appName("spark-sql")\
+    .getOrCreate()
+
+print >> sys.stderr, 'Running query: %s' % (query_file, )
+result = spark.sql(query)
+
+print >> sys.stderr, 'Writing result: %s' % (result_uri, )
+result.write.csv(result_uri)
```
komamitsu: These codes can throw an exception? If so, it seems we'd better catch an exception and call `spark.stop()`.

digdag-tests/src/test/resources/acceptance/emr/td-www_access.scala
```diff
@@ -0,0 +1,6 @@
+import com.treasuredata.spark._
+val td = spark.td
+val d = td.df("sample_datasets.www_access")
```
komamitsu: 👍

digdag-tests/src/test/resources/acceptance/emr/emr.dig
```diff
@@ -0,0 +1,68 @@
...
+      command: cat
+      files:
+        - data.csv
+      args: [data.csv]
+    - type: spark
+      application: td-www_access.scala
+      jars:
+        - s3://td-spark/td-spark-assembly-0.1.jar
+      conf:
+        spark.td.apikey: {secret: td.apikey}
```
komamitsu: We can evaluate a secret value here?

komamitsu:
Great work! 👍

digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java
komamitsu: Okay. Using BasicAWSCredentials here would be enough although I think we can reduce security risk using a temporary credential just in case when the original credentials is leaked.

danielnorberg: How does the emr operator granting itself temporary credentials lower the security risk?

digdag-tests/src/test/resources/acceptance/emr/emr.dig
danielnorberg: The emr operator code fetches secrets for `{secret: <key>}` values, KMS-encrypts and then stores them in the step `config.json`. 

https://github.com/treasure-data/digdag/blob/7717bd144473dda32e8b6d0fb05f077f7582bb10/digdag-standards/src/main/java/io/digdag/standards/operator/aws/EmrOperatorFactory.java#L1370

PR #349, opened 2016-10-11
[WIP] Adding ConfigScope utility class

frsyuki:
This assumes #344. Diff from #344 is https://github.com/treasure-data/digdag/compare/operator-context...config-scope
suspended until #340 is ready to merge.

PR #399, opened 2016-12-02
Using _ instead of - for secret keys to be consistent with Config

frsyuki:
Many secrets can be also set as parameters because the difference of
confidential and non-confidential is not always obvious for users. Some
may think AWS access_key_id is confidential but some may not.

Therefore, users and plugin developers may set the same configuration to
secrets or parameter differently, and their naming should be consistent
as much as possible to avoid misconfiguration, to make documents easier
to understand, and to make plugins easier to write. This is a blocker of #349.


frsyuki:
@danielnorberg any thoughts on this change?

danielnorberg:
👍 LGTM

frsyuki:
thanks for review! 👍

PR #401, opened 2016-12-06
Include Digdag UI in the jar package

frsyuki:
This PR integrates Digdag UI to the HTTP server so that accessing to `http://localhost:65432` shows the UI. It is hard to use UI now because we need to run 2 servers with CORS correctly setup.

How it works: `/api/` paths match with the regular application first. If the path doesn't match, the new JAX-RS application (ServerModule.bindUiApplication) serves CSS/JS/HTML files. Files must be in a directory named `public` in the classpath.

Custom applications that extend ServerModule (including TD's hosted environment) won't serve CSS/JS/HTML files because the CSS/JS/HTML files are copied to classpath only when the jar package is built using `:digdag:shadowJar`. If someone wants to do that, we can release `digdag-ui.jar` to bintray so that those apps can include the `public/*` files in the jar files.


digdag-ui/config.sample.js
```diff
@@ -0,0 +1,59 @@
+var DIGDAG_CONFIG = {
+  url: 'http://localhost:65432/api/',
```
dzannotti: I'd put this file in `/digdag-ui/config/config.sample.js`

digdag-ui/.gitignore
```diff
@@ -2,4 +2,5 @@
 node_modules
 .DS_Store
 npm-debug.log
+npm-debug.log.*
```
dzannotti: I think here we just want `*.log` rather

frsyuki: Files are like `npm-debug.log.564447649`. So *.log doesn't work.
I prefer making it restricted so that we can avoid unexpected missing files.

digdag-ui/config.sample.js
dzannotti: Yeah makes sense, but I don't think that we should be adding a sample config to the root where the source for the app is, i guess it doesn't matter

frsyuki: hm...yes, I remove it. I hope we can add comments to the file...

danielnorberg:
👍

dzannotti:
LGTM once test passes

danielnorberg:
:shipit:

frsyuki:
👍  Merged 92d4bb4 to master.

PR #402, opened 2016-12-06
UI: Fixes files not displaying

dzannotti:
Some files were not displaying correctly, I'm not sure what caused this, but i investigated the cause and it seemed to be centered around `typeof node.constructor === 'object'` being a function


danielnorberg:
👍

PR #405, opened 2016-12-07
Download command

frsyuki:
`digdag download <project>` command lets users download a project archive to a local directory.
This is useful when we want to check the definition of workflows including query files and scripts as well as *.dig files themselves.


danielnorberg:
👍

PR #406, opened 2016-12-07
(typeof null === 'object') is true

frsyuki:
`node == null` is causing unexpected exception at `node['_type']` in `function task`.


frsyuki:
@danielnorberg can you review?

dzannotti:
:+1:

frsyuki:
👍

PR #344, opened 2016-10-08
Moving all operator arguments to OperatorContext interface

frsyuki:
#340 made it possible to move `Operator.secretSelectors` to `OperatorFactory`.

This change takes advantage it and moves all arguments of operator to `OperatorContext` interface. Operator implementation will be:

```
class MyOperator
        extends BaseOperator
{
    MyOperator(OperatorContext context)
    {
        super(context);
    }

    @Override
    public TaskResult runTask()
    {
        ...
    }
}
```

A benefit is that we can add more fields to OperatorContext in the future without breaking backward compatibility of operator API.

This PR assumes #340. Diff from #340 is https://github.com/treasure-data/digdag/compare/privileged-vars...operator-context

PR #292, opened 2016-09-19
Redshift operator

frsyuki:

PR #407, opened 2016-12-07
Collection-returning REST API and string id

frsyuki:
This PR rebased #333 and #323 from master and added changes of digdag-ui.


frsyuki:
@dzannotti could you review changes of digdag-ui? I think they're ok because `npm run flow` passed. 

@danielnorberg could you review backend changes? There're 4 kinds of changes:

* digdag-client and digdag-cli
  * Most of changes are here. They use `io.digdag.client.api.Id` instead of `long` or `int` to represent ids. I created the wrapper class because javac can do more type checking.
* digdag-server
  * Response body uses Id instead of long/int for ID but query parameters keep using int/long. RESTeasy (JAX-RS) converts request params to int/long automatically.
* digdag-tests
  * All simple replacements. For example,
     * `Long.toString(attemptId)` -> `attemptId.toString()` (where attemptId is changed from long to Id)
     * `List<RestSchedule> scheds = client.getSchedules()` -> `List<RestSchedule> scheds = client.getSchedules().getSchedules()`
* digdag-core
  * Added `io.digdag.core.database.IdName` class because DatabaseProjectStoreManager was reusing api.IdName.

I'm thinking that next version should be `0.9.0` together with other breaking changes: Operator API (#344), and secret's - to _ change (#399).
We also have some fancy changes to convince users to upgrade: digdag-ui integration (#401), emr operators (#397), and redshift operators (#292).

digdag-ui/console.jsx
```diff
@@ -478,9 +478,9 @@ class ScheduleListView extends React.Component {
 
   fetchSchedule () {
     this.setState({loading: true})
-    model().fetchProjectWorkflowSchedule(this.props.projectId, this.props.workflowName).then(schedules => {
+    model().fetchProjectWorkflowSchedule(this.props.projectId, this.props.workflowName).then(scheduleCollection => {
```
dzannotti: I think it's worth destructuring here so we can use the short-hand for object

```
then(({ schedules }) => {
  this.setState({
     schedules,
     loading: false		          loading: false
   })
})
```

digdag-ui/console.jsx
```diff
@@ -593,8 +593,8 @@ class SessionsView extends React.Component {
   };
 
   componentDidMount () {
-    model().fetchSessions().then(sessions => {
-      this.setState({sessions})
+    model().fetchSessions().then(sessionCollection => {
```
dzannotti: same as above

digdag-ui/console.jsx
```diff
@@ -571,8 +571,8 @@ class ProjectsView extends React.Component {
   };
 
   componentDidMount () {
-    model().fetchProjects().then(projects => {
-      this.setState({projects})
+    model().fetchProjects().then(projectCollection => {
```
dzannotti: same as above

digdag-ui/console.jsx
frsyuki: that's an interesting syntax. I didn't know that. Updating the code...

digdag-ui/console.jsx
```diff
@@ -1337,9 +1337,9 @@ class AttemptLogsView extends React.Component {
   }
 
   fetchLogs () {
-    model().fetchAttemptLogFileHandles(this.props.attemptId).then(files => {
+    model().fetchAttemptLogFileHandles(this.props.attemptId).then(fileCollection => {
       if (!this.ignoreLastFetch) {
-        const sortedFiles = _.sortBy(files, 'fileTime')
+        const sortedFiles = _.sortBy(fileCollection.files, 'fileTime')
```
dzannotti: I'd use destructoring here too

digdag-ui/console.jsx
frsyuki: updated 👍

dzannotti:
Left a couple of minor comments, lgtm!

digdag-ui/console.jsx
frsyuki: 👍  updated

digdag-ui/console.jsx
frsyuki: 👍  updated

digdag-ui/console.jsx
```diff
@@ -1251,9 +1251,9 @@ class AttemptTasksView extends React.Component {
   }
 
   fetchTasks () {
-    model().fetchAttemptTasks(this.props.attemptId).then(tasks => {
+    model().fetchAttemptTasks(this.props.attemptId).then(taskCollection => {
       if (!this.ignoreLastFetch) {
-        this.setState({tasks})
+        this.setState({tasks: taskCollection.tasks})
```
dzannotti: nit: destructuring here too

digdag-ui/console.jsx
```diff
@@ -1635,14 +1635,14 @@ class SessionPage extends React.Component {
...
 
   fetchSession () {
-    model().fetchSession(parseInt(this.props.params.sessionId)).then(session => {
+    model().fetchSession(this.props.params.sessionId).then(session => {
       if (!this.ignoreLastFetch) {
         this.setState({session})
       }
     })
-    model().fetchSessionAttempts(parseInt(this.props.params.sessionId)).then(attempts => {
+    model().fetchSessionAttempts(this.props.params.sessionId).then(attemptCollection => {
```
dzannotti: nit: same

digdag-server/src/main/java/io/digdag/server/rs/RestModels.java
```diff
@@ -153,6 +239,32 @@ public static RestSession session(StoredSessionWithLastAttempt session, String p
...
+                        .collect(Collectors.toList()));
+
+        List<RestSession> collection = sessions.stream()
+                .map(session -> {
+                    try {
+                        return session(session, projs.get(session.getProjectId()).getName());
+                    }
+                    catch (ResourceNotFoundException ex) {
+                        // must not happen
+                        return null;
```
danielnorberg: throw here instead of ignoring exception

danielnorberg:
👍

digdag-server/src/main/java/io/digdag/server/rs/RestModels.java
frsyuki: 👍

PR #408, opened 2016-12-07
Add attempts.index to avoid confusion between session id and attempt id

frsyuki:
Attempts have similar number with sessions but they're different. Using
attempt id and session id on clients (digdag-cli and digdag-ui) is
confusing. The idea here is to assign a sequence number to attempts in a
session (like #1, #2, ...) so that clients can implement better
interface to identify an attempt (like 2981#1, 2981#2, ...).

### DIFF
--- a/digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
+++ b/digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
@@ -14,6 +14,8 @@ public interface RestSessionAttempt
 {
     Id getId();
 
+    int getIndex();
+
     IdAndName getProject();
 
     //Optional<String> getRevision();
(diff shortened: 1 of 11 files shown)


frsyuki:
@danielnorberg any thoughts about this?

CLI can support following syntaxes:

{code}
digdag kill <session-id>#<attempt-index>
digdag retry <session-id>#<attempt-index>
digdag log <session-id>#<attempt-index>
{code}

For example: `digdag kill 329#1`. Not soon but eventually, we can make `digdag kill 329` error because it seems confusing.

Web UI also can show #1, #2, ... instead of attempt IDs. It will look similar to Circle CI's UI.

In terms of REST API, we can add followings:

* `GET /api/sessions/{id}/attempts/{index}`
* `GET /api/sessions/{id}/attempts/{index}/tasks`
* `GET /api/sessions/{id}/attempts/{index}/retries`
* `POST /api/sessions/{id}/attempts/{index}/kill`

Following REST APIs are not harmful to keep but eventually we may want to deprecate:

* `GET /api/attempts/{id}`
* `GET /api/attempts/{id}/tasks`
* `GET /api/attempts/{id}/retries`
* `POST /api/attempts/{id}/kill`

danielnorberg:
👍  SGTM

frsyuki:
I realized that we can't use `#` in CLI or REST API because it's a comment in shell and a fragment identifier in URL...

danielnorberg:
Yeah, I think dot is a good idea. 👍

PR #412, opened 2016-12-08
Reimplement user-defined secret access api to operators

frsyuki:
The first secret API design included a use case where operators use secret keys given by users in workflow definition (e.g.: `config: {spark.td.apikey: {secret: td.apikey}}}` but #344 removed it.
This PR re-adds to the API to the operator API. Use case:

* Operators must declare list of accessing secret keys in OperatorFactory.getSecretAccessList.
* But users can explicitly grant access to secrets using `_secret` directive when users give name of the secrets as a configuration parameter. Operators can get these secrets even when getSecretAccessList doesn't include them.
* Operators still have a chance to reject it by implementing Operator.testUserSecretAccess(key) method. This is useful when operators give SecretProvider to an external library and operators want to avoid the library's unexpected use of secrets.


danielnorberg:
👍

frsyuki:
👍  Merged!

PR #413, opened 2016-12-08
User-given secret parameters

frsyuki:
Some operators such as `emr>` wants to accept secrets set by users as parameters like this:

```
emr>:
...
conf:
  spark.td.apikey: {secret: td.apikey}
```

This PR implements following syntax for above use case:

```
emr>:
...
conf:
  spark.td.apikey: ${secret:td.apikey}
```

Difference is that users can use `${secret:KEY}` syntax to set a secret as a parameter when operator allows. It is NOT JavaScript. Just a syntax sugar to get a user-given secret.

Users can't set secrets into arbitrary parameters. In above case, users can use `${secret:KEY}` syntax because `emr>` operator expects secrets at `conf`.

Operators use `userSecretTemplate(config.get("conf", String.class), context.getSecrets())` method to get the secrets. Everything else including system-global access policy and user's granting format (`_secret`) is same.


danielnorberg:
The emr operator KMS-encrypts secrets in order to safely pass them into EMR using the s3 staging area. The secrets are then decrypted on the EMR master before they're used in e.g. spark-submit options etc.

For this purpose, the emr operator code needs to be able to differentiate between secrets and regular parameters.

Currently the emr operator code does this by checking if the parameter value is a string (in which case it is handled as a non-secret and not kms-encrypted) or if it is a {secret: ...} object, in which case the secret value is kms-encrypted.

For this to keep working, userSecretTemplate would need to be able to indicate whether the value returned is secret or not.

I think this will be a recurring use case. It is useful for operator code to know whether a value is sensitive and should be protected or not.

frsyuki:
I see. Do you think it would work that operators use secure ways (such as KMS) if the keys might include secrets? I mean that a operator assumes that the value may include secret (and should use KMS or secure ways) once the operator applies `userSecretTemplate` whether it actually includes secrets or not.

I'm guessing that it also makes the operator implementations simple because they don't have to have two codes for secrets and non-secrets. I'm also concerning about plugin development where if the secret/non-secret handling are quite different, operator developers don't implement `{secret: ...}` handling and users of the operators are forced to set secrets as parameters. I would rather like to let such lazy developers use a simple API to support both of them.

danielnorberg:
In the case of the emr operator it means that every step confguration parameter that could contain a secret would need to be kms-encrypted, which is unnecessary and makes debugging etc harder.

I still believe that it is useful for operator code to know whether a value is sensitive and should be protected or not.

frsyuki:
I see. Concern around debugging makes sense.
How about this API?:

```java
public class UserSecretTemplate
{
    public static UserSecretTemplate of(String input) { ... }

    // returns true if this template includes secrets
    public boolean containsSecrets() { ... }

    public String format(SecretProvider secrets) { ... }
}
```

Operators use following code to use above class:

```java
UserSecretTemplate tmpl = UserSecretTemplate.of(value);
if (tmpl.containsSecrets()) {
  // KMS
  value = tmpl.format(secrets));
  ...
}
else {
  // non-KMS
  ...
}
```

danielnorberg:
That seems useful 👍

frsyuki:
👍

PR #414, opened 2016-12-09
scheduling: add skip_on_overtime option

danielnorberg:


frsyuki:
@danielnorberg S3WaitIT is failing and it's my fault (sorry, I broke the master once). bf25c9d3e330eee11c9fa5f96675b625c963e6b8 fixed it. So please rebase from master.

digdag-core/src/main/java/io/digdag/core/workflow/AttemptBuilder.java
```diff
@@ -110,14 +111,26 @@ public AttemptRequest buildFromStoredWorkflow(
...
+                // TODO: Come up with a better names for these parameters?
+
+                // This is the last session that was processed. Normally this will be the immediately preceding session,
+                // but it can be an earlier session if some were skipped due to skip_on_overtime = true.
+                params.set("last_processed_session_time", formatSessionTime(lastExecutedSessionTime.get(), timeZone));
+
+                // This is the first session time slot that should be included if the user wants to process any preceding skipped sessions.
+                // Normally this is the same as the current session_time but can be an earlier session if some were skipped.
+                Instant firstUnprocessedSessionTime = sr.get().nextScheduleTime(lastExecutedSessionTime.get()).getTime();
+                params.set("first_unprocessed_session_time", formatSessionTime(firstUnprocessedSessionTime, timeZone));
```
frsyuki: I think that we want to have `last_successful_session_time` instead of `last_processed_session_time` or `first_unprocessed_session_time`.
We, Treasure Data, were selling a solution packaged with a BI tool called Metric Insights. When we schedule a query on it, it skips execution on overtime, and it provides a variable similar to the last successful execution time. It was quite useful. I imagine this use case with `skip_on_overtime: true` option.
A typical query was like this:

```sql
select
  TD_TIME_FORMAT(time, 'yyyy-MM-dd', 'JST') AS d,
  count(distinct user_id) as uu
from purchases
where :last_measurement_time < TD_TIME_FORMAT(time, 'yyyy-MM-dd', 'JST')
group by 1
```

`:last_successful_timestamp` is last successfully measured (calculated) timestamp (@toru-takahashi should know about this well). Users run queries every hour or every day depending on computing resource they have.

There're two benefits:

1. People don't have to care much about a single failure because the next execution will recover the results.
2. Backfilling was (relatively) easy because what they needed was to resetting :last_measurement_time to a past timestamp.

2 may not apply to Digdag but 1 is good.

We can get last successful session using this query:

```sql
select sa.*
from sessions s
join session_attempts sa on s.last_attempt_id = sa.id
where project_id = :projectId and workflow_name = :workflowName
and s.session_time < :sessionTime
and bitand(sa.state_flags, `AttemptStateFlags.SUCCESS_CODE`) != 0  -- successful
order by s.session_time desc
limit 1
```

digdag-core/src/main/java/io/digdag/core/agent/RuntimeParams.java
```diff
@@ -37,6 +37,36 @@ public static Config buildRuntimeParams(ConfigFactory cf, TaskRequest request)
...
 
+        // last_processed_session_*
+        try {
+            // last_processed_session_time is set by AttemptBuilder
+            String sat = request.getConfig().get("last_processed_session_time", String.class, null);
+            if (sat != null) {
+                Instant instant = Instant.from(TIME_FORMAT.parse(sat));
+                setTimeParameters(params, "last_processed_session_", timeZone, instant);
+            } else {
+                setEmptyTimeParameters(params, "last_processed_session_", timeZone);
```
frsyuki: setting empty time is a good idea 👍

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
```diff
@@ -123,22 +132,36 @@ private void runSchedule(ScheduleControl lockedSched)
 
         try {
             StoredWorkflowDefinitionWithProject def = rm.getWorkflowDetailsById(sched.getWorkflowDefinitionId());
+
+            SessionStore ss = sessionStoreManager.getSessionStore(def.getProject().getSiteId());
+            List<StoredSessionAttemptWithSession> activeAttempts = ss.getActiveAttemptsOfWorkflow(def.getId(), 1, Optional.absent());
```
danielnorberg: Yeah, I was also wondering if execution should only be skipped if a preceding attempt is still running. In that case it might be better to change `getActiveAttemptsOfWorkflow` to e.g. `getActiveAttemptsOfWorkflowWithSessionTimeBefore` in order to catch manually (i.e. retry, backfill, start) started session attempts.

digdag-core/src/main/java/io/digdag/core/schedule/StoredSchedule.java
```diff
@@ -23,4 +23,6 @@
     public abstract Instant getUpdatedAt();
 
     public abstract Optional<Instant> getDisabledAt();
+
+    public abstract Optional<Instant> lastSessionTime();
```
danielnorberg: 👍

digdag-core/src/main/java/io/digdag/core/workflow/AttemptBuilder.java
danielnorberg: What would the value of `last_successful_session_time` be for the first scheduled session? 

My thinking was that `first_unprocessed_session_time` would be equal to e.g. `start_time` for the first scheduled session. However, `last_successful_session_time` would not be set because there would not be a preceding "successful" session. This would make `last_successful_session_time` unusable in workflows that have a "first" scheduled execution.

What was the value of `:last_successful_timestamp` in the first scheduled query in "Metric Insights" ?

frsyuki: > My thinking was that first_unprocessed_session_time would be equal to e.g. start_time for the first scheduled session

I see. That's a very good point.

I found the document of Metric Insights: [Write a SQL Statement for a Dimensioned Metric ](http://help.metricinsights.com/m/Creating_Dimensioned_Metrics_and_Reports/l/104398-write-a-sql-statement-for-a-dimensioned-metric)

> By instructing the system to **Batch Load Data Going back no more than** 14 days, :last_measurement_time would be set to be at most 14 days before the latest 'last measurement time' across all Dimension values. In this example, this would be 14 days before 2012-10-02 which will result in :last_measurement_time being set to '2012-09-18'.

For initial execution, Metric Insights subtract at most N days from the first scheduled time to set `:last_measurement_time`. Actually this behavior sounds useful.

How about this idea?:

* **last_session_time**
  * session_time of the last schedule. It doesn't matter whether the last session actually existed, failed, or skipped.
* **last_executed_session_time**
  * similar to _last_session_time_ but it goes back to actually executed session_time at most the period of `assume_executed` option. If `assume_executed` option is not set and this is the first execution, the value becomes empty string.
  * (I changed the name from `last_processed_session_time` because "processed" sounds confusing with "successful" for me)
* **last_successful_session_time**
  * similar to _last_session_time_ but it goes back to successfully finished session_time at most the period of `assume_successful` option. If `assume_successful` option is not set and there're no successful execution before, the value becomes empty string.

With above parameters, we can cover following 3 use cases:

### sessions run in parallel (currently supported)

```sql
schedule:
  daily>: 09:00:00

+analyze:
  td>:
  query: |
    ...
    WHERE '${last_session_time}' < date AND date < '${session_time}'
```

### session depends on the last execution, skip on overtime, failed execution always needs manual fix

```sql
schedule:
  daily>: 09:00:00
  skip_on_overtime: true
  assume_executed: 14 days

+analyze:
  td>:
  query: |
    ...
    WHERE '${last_executed_session_time}' < date AND date < '${session_time}'
```

### session depends on the last execution, skip on overtime, failed execution makes the next execution running longer

```sql
schedule:
  daily>: 09:00:00
  skip_on_overtime: true
  assume_successful: 14 days

+analyze:
  td>:
  query: |
    ...
    WHERE '${last_successful_session_time}' < date AND date < '${session_time}'
```

PR #416, opened 2016-12-12
add started_at field to task api endpoint

danielnorberg:
Expose the task start time in the `/api/attempts/{id}/tasks` endpoint so that task execution timings can be shown by the cli and visualized by the UI.


komamitsu:
👍

frsyuki:
Nice 👍

PR #419, opened 2016-12-12
download command to download to temp file

frsyuki:
I found that some tests on Travis CI failing with following message

```
acceptance.DownloadIT > download FAILED
    java.lang.AssertionError: 2016-12-12 10:03:54 +0000: Digdag v0.8.23-SNAPSHOT
    error: Attempted read on closed stream. (io)

    Expected: is <0>
         but: was <1>
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.junit.Assert.assertThat(Assert.java:956)
        at acceptance.DownloadIT.download(DownloadIT.java:78)
```

They succeed when I restart it. Example: https://api.travis-ci.org/jobs/183174886/log.txt?deansi=true

One of the possible causes is that HTTP server disconnects the TCP session because of socket timeout when client's extracting process is too slow.
This PR changes the `download` command so that client downloads the archive to a temp file as fast as possible so that following extracting doesn't keep the socket idling.

This assumes #373 to be merged.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/client/Download.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/client/Download.java
@@ -1,16 +1,20 @@
 package io.digdag.cli.client;
 
 import com.beust.jcommander.Parameter;
+import com.google.common.io.ByteStreams;
 import io.digdag.cli.SystemExitException;
 import io.digdag.client.DigdagClient;
 import io.digdag.client.api.RestProject;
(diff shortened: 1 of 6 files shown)


frsyuki:
@danielnorberg PTAL

danielnorberg:
👍

danielnorberg:
Taking the liberty of merging this.

PR #421, opened 2016-12-13
td operators should not retry on invalid table name

frsyuki:
TDClient throws TDClientException (not TDClientHttpException) when
problem happens before sending request or during receiving requests.
This includes deterministic exceptions such as invalid table name
because TDClient validates table names before sending it to the server
(at least with v0.7.26).


frsyuki:
@danielnorberg PTAL~

danielnorberg:
👍

frsyuki:
👍

Issue #423, opened 2016-12-14
connecting to redshift using pg operator fails


cosmok:
I get 'Failed to connect to the database' while running the workflow at the bottom.
I can connect to redshift via psql BTW.

timezone: Australia/Sydney
 _export:
   pg: 
     host: host
     port: 5439
     database: dbname
     user: username
     password: password
     ssl: false
 
 +create_temp_interval_table:
   pg>: queries/create_temp_interval_table.sql

hiroyuki-sato:
Hello @cosmok 

I'm not familiar about the Redshift.
But you may need to write database password in `~/.config/digdag/config`.
(When you use local mode).

If you use server mode, you need to setup [secrets](http://docs.digdag.io/command_reference.html#secret-encryption-key)

For example

~/.config/digdag/config

```
secrets.pg.password = your_postgresql_password_here
```

```
digdag init -t postgresql pg_test
2016-12-14 12:52:37 +0900: Digdag v0.8.22
  Creating pg_test/queries/create_src_table.sql
  Creating pg_test/queries/insert_data_to_src_table.sql
  Creating pg_test/queries/summarize_src_table.sql
  Creating pg_test/pg_test.dig
  Creating pg_test/.gitignore
```

modify connection part.

```yaml
_export:
  pg:
    user: digdag
    host: 127.0.0.1
    database: digdag_test
```

```
digdag run pg_test.dig -a


2016-12-14 12:56:12 +0900: Digdag v0.8.22
2016-12-14 12:56:14 +0900 [WARN] (main): Reusing the last session time 2016-12-14T00:00:00+00:00.
2016-12-14 12:56:14 +0900 [INFO] (main): Using session /private/tmp/pg_test/.digdag/status/20161214T000000+0000.
2016-12-14 12:56:14 +0900 [INFO] (main): Starting a new session project id=1 workflow name=pg_test session_time=2016-12-14T00:00:00+00:00
2016-12-14 12:56:15 +0900 [INFO] (0017@+pg_test+caution): echo>: This workflow can work with PostgreSQL running in your local machine. If you already set up PostgreSQL, please create database and user described in the above configuration
This workflow can work with PostgreSQL running in your local machine. If you already set up PostgreSQL, please create database and user described in the above configuration
2016-12-14 12:56:16 +0900 [INFO] (0017@+pg_test+create_src_table): pg>: queries/create_src_table.sql
[...]

PR #428, opened 2016-12-15
Fix issue #423

komamitsu:
PostgreSQL's prepared statement is a server-side object. So if
the query creates a table and refers it later,
JdbcConnection#validateStatement fails. Replacing
Connection#prepareStatement with Connection#nativeSQL would
be a better solution. That's a local side validation.


komamitsu:
@frsyuki Can you review this change?

frsyuki:
Good catch 👍  Code also looks good.

How about adding a comment there? Like:

```
// Here uses nativeSQL instead of Connection#prepareStatement because prepareStatement
// validates the SQL by creating a server-side prepared statement and PostgreSQL
// wrongly assumes that the SQL is broken in this case:
//   * the SQL includes multiple statements
//   * a statement creates a table and a later statement uses it
```

PR #429, opened 2016-12-15
ui: background refresh [ci skip]

danielnorberg:
No more manual refresh! \o/


komamitsu:
Nice! The change itself looks good. But I have a bit concern that this change might increase API calls and cause heavier load on server side. But it looks we can control the interval of refresh on our side changing the constant, so it's okay for the first release.
👍

Issue #427, opened 2016-12-15
link to Digdag server document is broken


cosmok:
here:
http://docs.digdag.io/command_reference.html#server

broken link:
under: -c, --config PATH
[Digdag server] http://docs.digdag.io/digdag_server.html

frsyuki:
Fixed at #430

cosmok:
Hi,
Thanks for updating the links, but, where can I read about all the server properties? I am trying to have only instance of a particular scheduled workflow/project to run (so no overlaps if the workflow takes longer than the schedule frequency), I can see io.digdag.limits.maxAttempts  in the code/tests, 1. is that the right parameter to set 2. where can I see all the parameters and description for servers and clients?

hiroyuki-sato:
@cosmok 
You can check server parameters below of the `In the config file, following parameters are available` on http://docs.digdag.io/command_reference.html#server

client parameters are 
http://docs.digdag.io/command_reference.html#client-mode-commands

Maybe, It seems that the `io.digdag.limits.maxAttempts` can't configure yet. 
That's why you can't find the document about it. 
https://github.com/treasure-data/digdag/blob/master/digdag-core/src/main/java/io/digdag/core/Limits.java#L11

cosmok:
@hiroyuki-sato thank you!
Do you know of a way to just one instance of a workflow running?

hiroyuki-sato:
@cosmok 

I'm not sure, but have you ever tried `digdag server --max-task-threads N`? 
http://docs.digdag.io/command_reference.html#server-mode-commands


I tested `digdag server -m --max-task-threads 1`

```yaml
timezone: Asia/Tokyo

schedule:
  minutes_interval>: 1

+setup:
  echo>: start ${session_time}

+disp_current_date:
  echo>: ${moment(session_time).utc().format('YYYY-MM-DD HH:mm:ss Z')}

+repeat:
  for_each>:
    order: [first, second, third]
    animal: [dog, cat]
  _do:
    echo>: ${order} ${animal}
  _parallel: true

+teardown:
  echo>: finish ${session_time}
```


```
2016-12-23 10:06:03 +0900 [INFO] (0030@+hoge+repeat): for_each>: {order=[first, second, third], animal=[dog, cat]}
2016-12-23 10:06:03 +0900 [INFO] (0030@+hoge+repeat^sub+for-0=order=0=first&1=animal=0=dog): echo>: first dog
first dog
[...]

PR #430, opened 2016-12-15
Fix dead link #427.

hiroyuki-sato:
I copied from . [Client-mode commands](http://docs.digdag.io/command_reference.html#client-mode-commands)


frsyuki:
👍

hiroyuki-sato:
Thanks!

PR #434, opened 2016-12-16
Fix selfupdate on Windows

frsyuki:
On Windows, a process can't change or delete a file if it is executing.
`selfupdate` command was failing due to this limitation. This change
uses a background process on Windows so that it repeats moving the file
until the execution finishes. The background process deletes itself at
the end using an exceptional behavior of Windows where a bat script can
delete itself.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/SelfUpdate.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/SelfUpdate.java
@@ -2,6 +2,7 @@ package io.digdag.cli;
 
 import com.beust.jcommander.Parameter;
 import com.google.common.io.ByteStreams;
+import com.google.common.io.Resources;
 import org.jboss.resteasy.client.jaxrs.ResteasyClientBuilder;
 
 import javax.ws.rs.client.Client;
@@ -13,6 +14,8 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.URISyntaxException;
 import java.nio.file.AccessDeniedException;
+import java.nio.file.AtomicMoveNotSupportedException;
+import java.nio.file.FileSystemException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -20,6 +23,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import static io.digdag.cli.SystemExitException.systemExit;
+import static java.nio.file.StandardCopyOption.ATOMIC_MOVE;
 import static java.nio.file.StandardCopyOption.REPLACE_EXISTING;
 import static java.util.Locale.ENGLISH;
 import static javax.ws.rs.core.UriBuilder.fromUri;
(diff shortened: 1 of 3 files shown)


frsyuki:
@komamitsu can I ask you to take a look?

digdag-cli/src/main/java/io/digdag/cli/SelfUpdate.java
```diff
@@ -100,28 +104,79 @@ private void selfUpdate(String version)
...
+            }
+            path.toFile().setExecutable(true, false);
+            path.toFile().setReadable(true, false);
+
+            // try to move the file to make sure that the file is on the same file system
+            // with the destination so that enable move more likely works. This is also
+            // helpful if /tmp file system is mounted with noexec option on Linux.
+            Path near = dest.getParent().resolve(".digdag.selfupdate.bat");
+            try {
+                Files.move(path, near, REPLACE_EXISTING);
```
komamitsu: I didn't get why this code would improve the operation. What problems do you want to resolve with the change?

digdag-cli/src/main/resources/digdag/cli/selfcopy.bat
```diff
@@ -0,0 +1,11 @@
+@echo off
+set i=0
+:repeat
+  if %i% leq 30 (goto move) else goto :done
+:move
+  ping 1.1.1.1 -n 1 -w 500 >NUL 2>NUL
```
komamitsu: 😿  👍

komamitsu:
I left some comments. But totally LGTM 👍

digdag-cli/src/main/resources/digdag/cli/selfcopy.bat
```diff
@@ -0,0 +1,11 @@
...
+set i=0
+:repeat
+  if %i% leq 30 (goto move) else goto :done
+:move
+  ping 1.1.1.1 -n 1 -w 500 >NUL 2>NUL
+  move /Y %1 %2 >NUL 2>NUL
+  set /a "i = i + 1"
+  if exist %1 goto repeat
+:done
+del %0
```
frsyuki: That's a good point. Let me check

digdag-cli/src/main/java/io/digdag/cli/SelfUpdate.java
frsyuki: In fact, this change is not for Windows. On UNIX, updating a file using atomic operation is necessary to not make a running java process crashed. Java process opens a jar file and keeps its file descriptor opened so that it reads class files lazily. Overriding the file durning running the process impacts this lazy loading.

digdag-cli/src/main/resources/digdag/cli/selfcopy.bat
frsyuki: Putting `del %0` at the head didn't work.
What I found instead is that we should use `%~f0` instead of `%0`: http://stackoverflow.com/questions/2888976/how-to-make-bat-file-delete-it-self-after-completion

PR #435, opened 2016-12-19
td: Add support for ${secret:...} syntax in result_url option

frsyuki:
This lets users to set password without writing it in plain text.

### DIFF
--- a/digdag-tests/src/test/resources/acceptance/td/td/td_result_url.dig
+++ b/digdag-tests/src/test/resources/acceptance/td/td/td_result_url.dig
@@ -1,0 +1,8 @@
+timezone: UTC
+
++run:
+  td>: query.sql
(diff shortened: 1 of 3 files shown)


frsyuki:
@komamitsu PTAL

digdag-tests/src/test/resources/acceptance/td/td/td_result_url.dig
```diff
@@ -0,0 +1,8 @@
+timezone: UTC
+
++run:
+  td>: query.sql
+  result_url: td://${secret:td.apikey}@/${td.database}/result_url
```
komamitsu: If the placeholder isn't converted at all or is replaced with empty string, this job should fail. So this test can check if the changes work well.
👍

komamitsu:
👍

frsyuki:
👍

PR #436, opened 2016-12-21
Selfupdate should not always use system's default java command

frsyuki:
This fixes `selfupdate` command so that it uses the same `java` command used to execute this digdag command. This is useful when the system doesn't have `java` in PATH and users use a custom java installation to run digdag.
This assumes #435.


komamitsu:
👍

PR #438, opened 2016-12-26
Update RedshiftPersistentTransactionHelper to not leave garbage tables

frsyuki:
This fixes following issues:

* doesn't leave garbage status tables even when action fails
  after prepare() (which commits CREATE TABLE)
  before updateStatusRowAndCommit() (which commits INSERT INTO).
* fixes help message that suggests schema name in status_table option.
  because status table name is escaped using escapeIdent, the option
  can't include schema name.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/redshift/RedshiftConnection.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/redshift/RedshiftConnection.java
@@ -197,26 +197,127 @@ public class RedshiftConnection
         return sb.toString();
     }
 
+    // Redshift doesn't support row-level locks or FOR UPDATE NOWAIT. Therefore, unlike
+    // PgPersistentTransactionHelper, here uses following strategy:
+    //
+    //   1. BEGIN transaction
+    //   2. CREATE TABLE ${status_table}_${queryId} (query_id, created_at, completed_at)
+    //      AS SELECT '${queryId}'::text, CURRENT_TIMESTAMP::timestamptz, NULL::timestamptz";
+    //   3. If CREATE TABLE succeeded, this transaction is locking the table. Run the action,
+    //      and COMMIT the transaction.
+    //   4. If CREATE TABLE failed with SQL state 23505, another thread or node succeeded to
+    //      commit CREATE TABLE statement before. Thus skip the action.
+    //
     private class RedshiftPersistentTransactionHelper
-            extends PgPersistentTransactionHelper
+            implements TransactionHelper
     {
-        RedshiftPersistentTransactionHelper(String statusTableName, Duration cleanupDuration)
+        private String statusTableNamePrefix;
+        private final Duration cleanupDuration;
+
+        RedshiftPersistentTransactionHelper(String statusTableNamePrefix, Duration cleanupDuration)
         {
-            super(statusTableName, cleanupDuration);
+            this.statusTableNamePrefix = statusTableNamePrefix;
+            this.cleanupDuration = cleanupDuration;
+        }
+
+        private String statusTableName(UUID queryId)
+        {
+            return String.format("%s_%s", statusTableNamePrefix, queryId);
         }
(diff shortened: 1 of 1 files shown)


digdag-standards/src/main/java/io/digdag/standards/operator/redshift/RedshiftConnection.java
```diff
@@ -197,26 +197,127 @@ String buildUnloadStatement(UnloadConfig unloadConfig)
...
+                    " (query_id, created_at, completed_at)" +
+                    " AS SELECT '%s'::text, CURRENT_TIMESTAMP::timestamptz, NULL::timestamptz",
+                    escapeIdent(statusTableName(queryId)),
+                    queryId.toString());
+            try {
+                execute(sql);
+                return true;
+            }
+            catch (SQLException ex) {
+                if (isConflictException(ex)) {
```
komamitsu: Where is this table locked?

Also, `CREATE TABLE IF NOT EXISTS` itself throws SQLException that has `23505` error code?

komamitsu: Using the existence of a status table itself as something like a lock would work to avoid duplicated adding records, but it sounds that retries will keep failing if the first connection is disconnected after creating a status table...

komamitsu: T1
```
digdag_test=# begin; create table locktbl (i int);
BEGIN
CREATE TABLE
```
```
digdag_test=# select * from stv_locks;
 table_id |     last_commit     |        last_update         | lock_owner | lock_owner_pid |    lock_owner_start_ts     |  lock_owner_end_ts  |      lock_status
----------+---------------------+----------------------------+------------+----------------+----------------------------+---------------------+------------------------
   599721 | 2000-01-01 00:00:00 | 2017-01-04 02:51:36.01517  |    5941070 |           1597 | 2017-01-04 02:51:36.015169 | 2000-01-01 00:00:00 | Holding write lock
   599721 | 2000-01-01 00:00:00 | 2017-01-04 02:51:36.015173 |    5941070 |           1597 | 2017-01-04 02:51:36.015169 | 2000-01-01 00:00:00 | Holding  delete lock
   599721 | 2000-01-01 00:00:00 | 2017-01-04 02:51:36.015173 |    5941070 |           1597 | 2017-01-04 02:51:36.015169 | 2000-01-01 00:00:00 | Holding  insert lock
(3 rows)
```
T2
```
digdag_test=# begin; create table locktbl (i int);
BEGIN
(waiting....)
```
T1
```
digdag_test=# rollback;
ROLLBACK
```
T2
```
CREATE TABLE
digdag_test=#
```
T1
```
digdag_test=# begin; create table locktbl (i int);
BEGIN
(waiting...)
```
T2
```
digdag_test=# commit;
COMMIT
```
T1
```
ERROR:  duplicate key violates unique constraint "pg_class_relname_nsp_index" (possibly caused by concurrent transaction conflict)
digdag_test=#
```
`CREATE TABLE` with the same table name works as an exclusive lock 👍

komamitsu:
👍

digdag-standards/src/main/java/io/digdag/standards/operator/redshift/RedshiftConnection.java
komamitsu: When an integration test ran with this change, it ran into this error.
```
Function "timestamptz(timestamp with time zone,integer)" not supported.
```

I guess it's related to that CURRENT_TIMESTAMP is a leader-node only function (http://docs.aws.amazon.com/redshift/latest/dg/Date_functions_header.html#date-functions-deprecated). I replaced CURRENT_TIMESTAMP with SYSDATE and confirmed it worked. I'll replace all CURRENT_TIMESTAMP used by Redshift operators with SYSDATE.

PR #439, opened 2016-12-26
Fix user secret filter of Operator

frsyuki:
Operator should be able to drop granted access to a secret key. On the
other hand, Operator must declare all keys if they are not granted by
users.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/DefaultSecretProvider.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/DefaultSecretProvider.java
@@ -16,19 +16,25 @@ import io.digdag.spi.SecretScopes;
 import io.digdag.spi.SecretStore;
 
 import java.util.List;
-import java.util.function.Predicate;
+
+import static java.util.Locale.ENGLISH;
 
 class DefaultSecretProvider
         implements SecretProvider
 {
+    static interface OperatorSecretFilter
+    {
+        boolean test(String key, boolean userGranted);
+    }
+
     private final SecretAccessContext context;
     private final SecretAccessPolicy secretAccessPolicy;
     private final Config grants;
-    private final Predicate<String> operatorSecretFilter;
+    private final OperatorSecretFilter operatorSecretFilter;
     private final SecretStore secretStore;
 
     DefaultSecretProvider(
-            SecretAccessContext context, SecretAccessPolicy secretAccessPolicy, Config grants, Predicate<String> operatorSecretFilter, SecretStore secretStore)
+            SecretAccessContext context, SecretAccessPolicy secretAccessPolicy, Config grants, OperatorSecretFilter operatorSecretFilter, SecretStore secretStore)
     {
         this.context = context;
         this.secretAccessPolicy = secretAccessPolicy;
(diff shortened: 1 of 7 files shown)


digdag-core/src/main/java/io/digdag/core/agent/DefaultSecretProvider.java
```diff
@@ -89,20 +93,32 @@ else if (node.isTextual()) {
                         .from(base)
                         .append(remainder)
                         .join(Joiner.on('.'));
+                if (!operatorSecretFilter.test(key, true)) {
+                    throw new SecretAccessFilteredException(key, String.format(ENGLISH,
+                            "Unexpected access to a secret key '%s' aliased from '%s'", key, remounted));
+                }
                 return fetchSecret(remounted);
             }
             else if (node.isBoolean() && node.asBoolean()) {
```
komamitsu: We don't take care of Integer?

frsyuki: I think we don't have to because previous code also doesn't.

digdag-core/src/main/java/io/digdag/core/agent/DefaultSecretProvider.java
```diff
@@ -16,19 +16,25 @@
...
 
 import java.util.List;
-import java.util.function.Predicate;
+
+import static java.util.Locale.ENGLISH;
 
 class DefaultSecretProvider
         implements SecretProvider
 {
+    static interface OperatorSecretFilter
```
komamitsu: `static` isn't needed

digdag-core/src/main/java/io/digdag/core/agent/DefaultSecretProvider.java
```diff
@@ -46,22 +52,20 @@
...
         //// Secret access control:
-        // 1. Reject if the operator doesn't need the access to the secret (operatorSecretFilter).
-        // 2. Allow if users explicitly grant access using _secret directive (Config grants).
-        // 3. Allow if system access policy (SecretAccessPolicy) allows.
-        // 4. Reject.
+        // 1. If users explicitly grant access using _secret directive in workflow definition (Config grants):
+        //    1-a. Allow if the operator wants to use it (operatorSecretFilter)
+        //    1-b. Reject.
+        // 2. If the system access policy file (SecretAccessPolicy) allows:
+        //    2-a. Allow if the operator wants to use it (operatorSecretFilter)
```
komamitsu: 1-a and 2-a is actually different in terms of existence of user grant. Could you add a bit detailed explanations?

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -324,27 +324,34 @@ protected TaskResult callExecutor(Path projectPath, String type, TaskRequest mer
         return operator.run();
     }
 
-    private static class OperatorSecretFilter
-            implements Predicate<String>
+    private static class AgentOperatorSecretFilter
+            implements DefaultSecretProvider.OperatorSecretFilter
```
komamitsu: Does this class need to have `allowUserSecretAccess` method? It seems there is no actual usage of this class without calling the method. If we can remove the method, this class can be simpler like this:
```
    private static class AgentOperatorSecretFilter
            implements DefaultSecretProvider.OperatorSecretFilter
    {
        private final Set<String> predeclaredSecretKeys;
        private final Predicate<String> dynamicFilter;

        AgentOperatorSecretFilter(OperatorFactory factory, Operator operator)
        {
            this.predeclaredSecretKeys = ImmutableSet.copyOf(
                    factory.getSecretAccessList().getSecretKeys());
            this.dynamicFilter = (key) -> operator.testUserSecretAccess(key);
        }

        @Override
        public boolean test(String key, boolean userGranted)
        {
            if (userGranted || predeclaredSecretKeys.contains(key)) {
                return dynamicFilter.test(key);
            }
            return false;
        }
    }
```

komamitsu:
LGTM 👍 
I left some comments, though.

But maybe we'd better get feedbacks from @danielnorberg if possible

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
frsyuki: it is not possible because an Operator instance is not available until OperatorManager calls `OperatorFactory.newOperator(OperatorContext)` but a DefaultSecretProvider.OperatorSecretFilter instance is required by `OperatorFactory.newOperator(OperatorContext)` because OperatorContext contains a SecretProvider instance that contains OperatorSecretFilter.

digdag-core/src/main/java/io/digdag/core/agent/DefaultSecretProvider.java
frsyuki: 👍

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
komamitsu: Ah, I see. I overlooked it...

frsyuki: I updated the comment at cc0ddd0.

komamitsu:
👍

PR #443, opened 2017-01-11
client: include User-Agent header with version

danielnorberg:


digdag-client/src/main/java/io/digdag/client/Version.java
```diff
@@ -1,4 +1,4 @@
-package io.digdag.core;
+package io.digdag.client;
```
frsyuki: I think you need to change build.gradle.

danielnorberg: 👍

PR #445, opened 2017-01-12
Add workflow edit UI

frsyuki:
This PR adds a simple workflow editing UI.

But changing workflow definition doesn't take effect at this moment. #322 is necessary first.


digdag-ui/console.jsx
```diff
@@ -20,6 +21,7 @@ import Duration from 'duration'
 import uuid from 'node-uuid'
 import jQuery from 'jquery'
 import ReactInterval from 'react-interval'
+import {Buffer} from 'buffer/'
```
dzannotti: / is unnecessary

digdag-ui/console.jsx
```diff
@@ -124,6 +127,14 @@ class CodeViewer extends React.Component {
     Ace.acequire('ace/ext/linking')
   }
 
+  getValue () {
+    if (this._editor) {
+      return this._editor.getValue()
+    } else {
```
dzannotti: nitpicking: i think the else condition is unnecessary, but i think its a stylistical chioce

digdag-ui/console.jsx
```diff
@@ -1920,6 +1950,313 @@ class WorkflowRevisionPage extends React.Component {
...
+    if (this.editor) {
+      text = this.editor.getValue()
+    } else {
+      text = this.props.file ? fileString(this.props.file.name, this.props.projectArchive) : ''
+    }
+
+    const buffer = new Buffer(text)
+    const ab = new ArrayBuffer(buffer.length);
+    const view = new Uint8Array(ab);
+    for (var i = 0; i < buffer.length; ++i) {
```
dzannotti: does `view.set(buffer)` works in this case? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set
I'm not a great fan of for loops or copying over like this

digdag-ui/console.jsx
```diff
@@ -1920,6 +1950,313 @@ class WorkflowRevisionPage extends React.Component {
...
+        </h3>
+        <pre>
+          <Measure>
+            { ({ width }) =>
+                <CodeEditor
+                  className='editor'
+                  language='yaml'  // TODO how to let ace guess language?
+                  value={file ? fileString(file.name, this.props.projectArchive) : ''}
+                  style={{ width }}
+                  ref={(vlaue) => { this.editor = vlaue }}
```
dzannotti: i think you meant value

digdag-ui/console.jsx
```diff
@@ -1920,6 +1950,313 @@ class WorkflowRevisionPage extends React.Component {
...
+
+  handleDelete (key) {
+    _.remove(this.state.entries, (file) => file.key === key)
+    this.setState({entries: this.state.entries})
+  }
+
+  handleAddFile () {
+    const key = uuid.v4()
+    this.state.entries.unshift({
+      newFile: Boolean(true),
```
dzannotti: No need to use the class constructor here

digdag-ui/console.jsx
```diff
@@ -1920,6 +1950,313 @@ class WorkflowRevisionPage extends React.Component {
...
+      key: key,
```
dzannotti: `key,` should be sufficient (es6 shorthand properties)

dzannotti: Also is key really relevant (should we not generate it at render time rather than keeping it as state?

digdag-ui/console.jsx
```diff
@@ -1920,6 +1950,313 @@ class WorkflowRevisionPage extends React.Component {
...
+      return
+    }
+    this.setInitialEntries()
+  }
+
+  render () {
+    const editors = this.state.entries.map(entry =>
+        <FileEditor
+          newFile={entry.newFile}
+          key={entry.key}
```
dzannotti: the idx for the array here is sufficient as key

digdag-ui/console.jsx
```diff
@@ -1920,6 +1950,313 @@ class WorkflowRevisionPage extends React.Component {
...
+  state = {
+    projectId: null,
+    project: null,
+    projectName: "new-project",
+    revisionName: uuid.v4(),
+    projectArchive: null,
+    saveMessage: "",
+  };
+
+  constructor (props) {
```
dzannotti: this is not needed (as we don't really use the constructor for anything here)

dzannotti:
I left some thoughts, most of them are style related, there are some linting and flow related errors

digdag-ui/console.jsx
```diff
@@ -191,6 +202,23 @@ class CodeViewer extends React.Component {
   }
 }
 
+class CodeEditor extends CodeViewer {
```
dzannotti: Extending React components can lead to unexpected behaviours (when children never call super), i think a better approach for this would be to make a component that passes down an option props to CodeViewer

What do you think of something like this?
```
diff --git a/digdag-ui/console.jsx b/digdag-ui/console.jsx
index d686ab14..b8b87074 100644
--- a/digdag-ui/console.jsx
+++ b/digdag-ui/console.jsx
@@ -106,6 +106,20 @@ class CodeViewer extends React.Component {
     className: ?string;
     language: string;
     value: string;
+    editorOptions: Object;
+  }
+
+  static defaultProps = {
+    editorOptions: {
+      enableLinking: true,
+      highlightActiveLine: false,
+      readOnly: true,
+      showLineNumbers: false,
+      showGutter: false,
+      showPrintMargin: false,
+      tabSize: 2,
+      useSoftTabs: true
+    }
   }
 
   /* eslint-disable */
@@ -139,16 +153,7 @@ class CodeViewer extends React.Component {
     const Ace = require('brace')
     require('./ace-digdag')
     this._editor = Ace.edit(this.editor)
-    this._editor.setOptions({
-      enableLinking: true,
-      highlightActiveLine: false,
-      readOnly: true,
-      showLineNumbers: false,
-      showGutter: false,
-      showPrintMargin: false,
-      tabSize: 2,
-      useSoftTabs: true
-    })
+    this._editor.setOptions(this.props.editorOptions)
     this._editor.on('click', this.editorClick.bind(this))
     this._updateEditor(this.props)
   }
@@ -173,6 +178,9 @@ class CodeViewer extends React.Component {
     if (nextProps.value !== this.props.value) {
       this._updateEditor(nextProps)
     }
+    if (this._editor && !_.isEqual(nextProps.editorOptions, this.props.editorOptions)) {
+      this._editor.setOptions(this.props.editorOptions)
+    }
   }
 
   componentWillUnmount () {
```

digdag-ui/console.jsx
frsyuki: key needs to be consistent (needs to be in state) to implement `handleDelete`

PR #446, opened 2017-01-12
automatically grant access to ${secret:...} keys

danielnorberg:


digdag-plugin-utils/src/main/java/io/digdag/util/UserSecretTemplate.java
```diff
@@ -26,6 +32,13 @@ public boolean containsSecrets()
         return TEMPLATE_PATTERN.matcher(source).find();
     }
 
+    public void forEachKey(Consumer<String> action) {
```
frsyuki: How about adding `List<String> getKeys()` instead of this? Because `Consumer` restricts use of checked exception. Users still can use `getKeys().stream().forEach(...)`.

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -324,6 +334,34 @@ protected TaskResult callExecutor(Path projectPath, String type, TaskRequest mer
         return operator.run();
     }
 
+    private static void setPath(Config config, String path, Object value)
+    {
+        Preconditions.checkArgument(!path.isEmpty(), "path");
+        List<String> parts = Splitter.on('.').splitToList(path);
```
frsyuki: You can use `io.digdag.client.config.ConfigKey`:

```java
ConfigKey key = ConfigKey.parse(path);
for (String nestName : key.getNestNames()) {
    config = config.getNestedOrSetEmpty(nestName);
}
config.set(parts.get(key.getLastName(), value);
```

digdag-plugin-utils/src/main/java/io/digdag/util/UserSecretTemplate.java
danielnorberg: Ok, I wanted to avoid unnecessary garbage.

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
danielnorberg: 👍

PR #448, opened 2017-01-13
Upload and download project archive files to/from S3

komamitsu:
If `archive.type` in system configuration is set to `s3`,
- digdag-server stores project archive files on S3
- digdag-cli directly downloads project archive files from S3 when `archive.s3.direct_download` is set to true on server side

The following configurations are needed to be set
- archive.s3.bucket
- archive.s3.credentials.access-key-id
- archive.s3.credentials.secret-access-key

### DIFF
--- a/digdag-client/src/main/java/io/digdag/client/DigdagClient.java
+++ b/digdag-client/src/main/java/io/digdag/client/DigdagClient.java
@@ -440,7 +440,9 @@ public class DigdagClient implements AutoCloseable
                 return request.invoke(webTarget);
             }
             catch (WebApplicationException e) {
-                firstRedirectException = e;
+                if (firstRedirectException == null) {
+                    firstRedirectException = e;
+                }
                 Response response = checkNotNull(e.getResponse());
                 int status = response.getStatus();
                 if (status % 100 == 3 && response.getLocation() != null) {
--- a/digdag-client/src/main/java/io/digdag/client/DigdagClient.java
+++ b/digdag-client/src/main/java/io/digdag/client/DigdagClient.java
@@ -22,7 +22,6 @@ import io.digdag.client.api.RestLogFileHandle;
 import io.digdag.client.api.RestLogFileHandleCollection;
 import io.digdag.client.api.RestProject;
 import io.digdag.client.api.RestProjectCollection;
-import io.digdag.client.api.RestRevision;
 import io.digdag.client.api.RestRevisionCollection;
 import io.digdag.client.api.RestSchedule;
 import io.digdag.client.api.RestScheduleCollection;
@@ -36,7 +35,6 @@ import io.digdag.client.api.RestSessionAttempt;
 import io.digdag.client.api.RestSessionAttemptCollection;
 import io.digdag.client.api.RestSessionAttemptRequest;
 import io.digdag.client.api.RestSetSecretRequest;
-import io.digdag.client.api.RestTask;
 import io.digdag.client.api.RestTaskCollection;
 import io.digdag.client.api.RestWorkflowDefinition;
 import io.digdag.client.api.RestWorkflowDefinitionCollection;
@@ -72,10 +70,10 @@ import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
 import java.util.function.Supplier;
-import java.util.stream.Stream;
 
 import static com.github.rholder.retry.StopStrategies.stopAfterAttempt;
 import static com.github.rholder.retry.WaitStrategies.exponentialWait;
+import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Predicates.not;
 import static java.util.Locale.ENGLISH;
 import static javax.ws.rs.core.HttpHeaders.USER_AGENT;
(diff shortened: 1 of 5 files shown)


komamitsu:
@frsyuki Can you review this p-r when you get a chance?

digdag-client/src/main/java/io/digdag/client/DigdagClient.java
```diff
@@ -650,7 +684,8 @@ private static boolean isDeterministicError(Response res)
 
     private static boolean isDeterministicError(int status)
     {
-        if (status >= 400 && status < 500) {
+        // Retrying against 3xx doesn't make sense
+        if (status >= 300 && status < 500) {
```
frsyuki: 👍

digdag-core/src/main/java/io/digdag/core/storage/ArchiveManager.java
```diff
@@ -99,9 +114,17 @@ private String formatFilePath(int siteId, String projectName, String revisionNam
     {
         Preconditions.checkArgument(!Strings.isNullOrEmpty(projectName), "projectName");
         Preconditions.checkArgument(!Strings.isNullOrEmpty(revisionName), "revisionName");
+        // It's possible the storage used to store a project archive in doesn't accept some characters.
+        // For instance, AWS S3 says the following characters are generally safe
+        //   - Alphanumeric characters [0-9a-zA-Z]
+        //   - Special characters !, -, _, ., *, ', (, and )
+        //
+        // So we'd better encode a project name into the characters above
+        String encodedProjectName = Base64.getEncoder().encodeToString(projectName.getBytes(UTF_8)).replace("=", "_");
```
frsyuki: I think what we want to use here is base64url without padding (defined in RFC 4648).
General base64 uses `/` character. `=` in base64 is padding which is OK to remove.
You can build this variant of Base64 by using Guava:

```java
import com.google.common.io.BaseEncoding;
BaseEncoding.base64Url().omitPadding().encode(projectName.getBytes(UTF_8));
```

digdag-core/src/main/java/io/digdag/core/storage/ArchiveManager.java
```diff
@@ -99,9 +114,17 @@ private String formatFilePath(int siteId, String projectName, String revisionNam
...
         return String.format(ENGLISH,
-                "%d/%s/%s.%s.tar.gz",
-                siteId, projectName, revisionName, DATE_TIME_SUFFIX_FORMAT.format(Instant.now()));
+                "%s%d/%s/%s.%s.tar.gz",
+                pathPrefix, siteId, encodedProjectName, revisionName,
+                DATE_TIME_SUFFIX_FORMAT.format(Instant.now()));
```
frsyuki: I found a minor problem...: `ProjectResource.putProject` uploads a file to a storage, and then inserts a row to RDBMS. This INSERT may fail if the projectName & revision is already used. But the path on storage uses projectName & revision. So, if two users upload the same revision name at the same time, one fails due to unique index but the file could be overridden. `Instant.now` is added to the path to prevent this issue but its precision is second. Conflict may still happen.

digdag-core/src/main/java/io/digdag/core/storage/ArchiveManager.java
```diff
@@ -66,13 +65,29 @@ public String getPath()
...
+    private final boolean directDownloadEnabled;
 
     @Inject
     public ArchiveManager(StorageManager storageManager, Config systemConfig)
     {
         this.storageManager = storageManager;
         this.systemConfig = systemConfig;
         this.uploadArchiveType = systemConfig.get("archive.type", ArchiveType.class, ArchiveType.DB);
+        this.pathPrefix = getArchivePathPrefix(systemConfig, uploadArchiveType);
+        this.directDownloadEnabled = systemConfig.get("archive." + uploadArchiveType + ".direct_download", Boolean.class, false);
```
frsyuki: 👍

digdag-client/src/main/java/io/digdag/client/DigdagClient.java
```diff
@@ -425,17 +424,52 @@ public RestProject putProjectRevision(String projName, String revision, File bod
...
+    {
+        WebApplicationException firstRedirectException = null;
+        WebTarget webTarget = initialWebTarget;
+
+        for (int i = 0; i < MAX_REDIRECT; i++) {
+            try {
+                return request.invoke(webTarget);
+            }
+            catch (WebApplicationException e) {
+                firstRedirectException = e;
```
komamitsu: Oops. Good catch 👍

digdag-core/src/main/java/io/digdag/core/storage/ArchiveManager.java
komamitsu: @frsyuki How about just using UUID instead of `Instant.now`?
It seems the key of S3 file doesn't need to include a timestamp in the name since we can remove expired S3 files using `revisions` table in digdag's database or S3's lifecycle feature.

frsyuki: Using UUID sounds good. S3 anyways has created-at time as a metadata of a file.

frsyuki:
@komamitsu I created #450 so that we can describe S3 storage in docs.

frsyuki:
LGTM 👍

komamitsu:
@frsyuki Thanks!

komamitsu:
Thanks!

komamitsu:
Rebased

PR #449, opened 2017-01-16
Fix a race issue in RedshiftIT

komamitsu:


komamitsu:
Confirmed RedshiftIT itself works fine

PR #452, opened 2017-01-16
remove dead secret access limit code

danielnorberg:


frsyuki:
:+1: Less code = better

PR #455, opened 2017-01-18
Redirect w/o headers when downloading project archive

komamitsu:


komamitsu:
@frsyuki Can you review this change?

digdag-client/src/main/java/io/digdag/client/DigdagClient.java
```diff
@@ -427,17 +427,18 @@ public RestProject putProjectRevision(String projName, String revision, File bod
     @FunctionalInterface
     interface RequestWithFollowingRedirect<T>
     {
-        T invoke(WebTarget webTarget);
+        T invoke(WebTarget webTarget, Optional<WebTarget> previousWetTarget);
```
frsyuki: This is minor thing but how about using `Optional<Response> redirectResponse` instead? `Response` provides more information to the caller because previousWebTarget is anyways available.

frsyuki:
👍

PR #460, opened 2017-01-19
Fix search page of sphinx

frsyuki:
search is broken: http://docs.digdag.io/search.html?q=test&check_keywords=yes&area=default
It seems that sphinx_rtd_theme with Sphinx 1.5 has an issue. Downgrading to 1.4.9 fixed the issue.


danielnorberg:
👍

frsyuki:
👍

PR #461, opened 2017-01-19
ui: fix schedule pause/resume button

danielnorberg:
* Only show the button if there is a schedule for the workflow

* Remove the "loading" button state to eliminate flicker due to periodic background update.


frsyuki:
:+1: please merge!

PR #463, opened 2017-01-19
ui: refresh on retry

danielnorberg:
Instead of jumping out to the sessions view, stay in the single session view and refresh it.


digdag-ui/config/config.js
```diff
@@ -1,5 +1,5 @@
 var DIGDAG_CONFIG = {
-  url: '/api/',
+  url: 'http://localhost:65432/api/',
```
frsyuki: Ahhh! I also do this often but this shouldn't be here!

danielnorberg: Ooops

PR #465, opened 2017-01-20
1.0 should be newer than 1.0-SNAPSHOT

frsyuki:
`io.digdag.client.Version` class is used to implement cli version checking.
1.0 should be newer than 1.0-SNAPSHOT.
We usually don't release -SNAPSHOT versions but it's during development.

This is same with Maven's ArtifactVersion.compareTo implementation: https://github.com/apache/maven/blob/bce33aa2662a51d18cb00347cf2fb174dc195fb1/maven-artifact/src/test/java/org/apache/maven/artifact/versioning/ComparableVersionTest.java#L166

### DIFF
--- a/digdag-client/src/main/java/io/digdag/client/Version.java
+++ b/digdag-client/src/main/java/io/digdag/client/Version.java
@@ -50,6 +50,8 @@ public class Version
     @Override
     public int compareTo(Version another)
     {
+        // 1.0 < 1.1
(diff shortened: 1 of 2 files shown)


frsyuki:
@komamitsu PTAL

danielnorberg:
👍

komamitsu:
👍

frsyuki:
:+1:

PR #467, opened 2017-01-24
Fix "required attributes are not set [retryCount]" error

frsyuki:
This happens if REST API tried to get tasks of an very old attempt that
did not have retryCount field. This change solves it by setting a
default value.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -1,6 +1,8 @@
 package io.digdag.core.database;
 
+import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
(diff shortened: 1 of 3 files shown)


digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -149,7 +151,9 @@ public DatabaseSessionStoreManager(DBI dbi, ConfigFactory cf, ConfigMapper cfm,
         dbi.registerMapper(new InstantMapper());
         dbi.registerArgumentFactory(cfm.getArgumentFactory());
 
-        this.mapper = mapper;
+        this.taskArchiveMapper = mapper.copy();
+        this.taskArchiveMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
komamitsu: 👍

komamitsu:
👍

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
frsyuki: @komamitsu In fact, this line is not the main fix. This line is for this: https://github.com/treasure-data/digdag/pull/467/files#diff-b8b63c99b1a34f8b65c161e6885fa07cR477
The main fix (fix for adding columns) is this: https://github.com/treasure-data/digdag/pull/467/files#diff-a6a683385d89bdbb7743550326a8645eR38

frsyuki:
👍

PR #470, opened 2017-01-27
Add autoCommit for test

frsyuki:


digdag-core/src/main/java/io/digdag/core/database/ThreadLocalTransactionManager.java
```diff
@@ -82,21 +140,37 @@ public void abort()
...
     public ThreadLocalTransactionManager(DataSource ds)
     {
+        this(ds, false);
+    }
+
+    ThreadLocalTransactionManager(DataSource ds, boolean autoAutoCommit)
+    {
         this.ds = checkNotNull(ds);
+        if (autoAutoCommit) {
+            LazyTransaction transaction = new LazyTransaction(ds, true);
```
komamitsu: Creating auto-commit style transaction and storing it in Thread local should be in `begin()` method lazily?

frsyuki: A purpose of autoAutoCommit is that threads don't have to call `begin()` or even `autoCommit()`. Instead, threads directly call `getHandle`. And it starts autoCommit transaction (which is not actually transaction though).
And `getHandle` is still lazy operation. So I think this is fine.

komamitsu:
@frsyuki I noticed `WorkflowExecutor#submitTasks` tries to use aborted `BasicHandle` when `ResourceConflictException` occurs.

```
        catch (ResourceConflictException sessionAlreadyExists) {
            StoredSessionAttemptWithSession conflicted;
            if (ar.getRetryAttemptName().isPresent()) {
                conflicted = sm.getSessionStore(siteId)
                    .getAttemptByName(session.getProjectId(), session.getWorkflowName(), session.getSessionTime(), ar.getRetryAttemptName().get());
            }
            else {
                conflicted = sm.getSessionStore(siteId)
                    .getLastAttemptByName(session.getProjectId(), session.getWorkflowName(), session.getSessionTime());
            }
            throw new SessionAttemptConflictException("Session already exists", sessionAlreadyExists, conflicted);
        }
```

https://github.com/treasure-data/digdag/blob/eb83a68cc9f4ca79992980dd565ca3cb5b31fe34/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java#L317-L324

`ThreadLocalTransactionManager#getHandle` should return a new handle if underlying BasicHandle is already aborted?

frsyuki:
@komamitsu that's a good point...
`ResourceConflictException` happens when `insertAttempt` throws the exception (`putAndLockSession` doesn't throw ResourceConflictException).
How about doing this?: https://github.com/treasure-data/digdag/pull/472

frsyuki:
(I noticed that Transaction interface is unnecessary...?)

komamitsu:
@frsyuki 
> Transaction interface
Interesting. Instantiating `BasicHandle` is just passing information to it, so we may be able to directly use it for lazy instantiation without `Transaction` interface.

PR #472, opened 2017-01-30
Add TransactionManager#reset to allow exception handlers to use db

frsyuki:


komamitsu:
Hmm. It doesn't seem to work well... https://travis-ci.org/treasure-data/digdag/jobs/197880670

komamitsu:
@frsyuki 
How about modifying `io.digdag.core.database.BasicDatabaseStoreManager#transaction(io.digdag.core.database.BasicDatabaseStoreManager.TransactionActionWithExceptions<T,D,E1,E2,E3>, java.lang.Class<E1>, java.lang.Class<E2>, java.lang.Class<E3>)` like this? I'm assuming E1/E2/E3 other than RuntimeException are expected exceptions.

```
        Handle handle = transactionManager.getHandle(configMapper);
        try {
            return action.call(handle, handle.attach(daoIface));
        }
        catch (Exception e) {
            // TODO: Revisit around here
            if (e instanceof RuntimeException) {
                throw e;
            }
            if (exClass1.isInstance(e) || exClass2.isInstance(e) || exClass3.isInstance(e)) {
                transactionManager.reset();
            }
            throw e;
        }
```

frsyuki:
@komamitsu test is not working because `run` is not working. LocalSite needs to commit so that runner threads can read the commit.

frsyuki:
@komamitsu I think that resetting transaction automatically causes unexpected troubles. For example, in this case:

```
doABC() {
  begin(() -> {
    doA();  // change some records in DB
    doB();  // failed to change some records in DB
    doC();  // change some records in DB
  });
}
```

This `doABC` assumes that doA and doB run transactionally but actually only doC takes effect.

PR #473, opened 2017-01-30
Add session_id to the runtime parameter

### LINKED ISSUE #469: digdag  query session log by uuid
Hi,

Since session_id is no longer available, is it possible to add support for querying logs by uuid?

The idea is to to mail logs when a workflow fails.

frsyuki:
`session_id` was removed once because there was a unclear design decision
around session and attempt. Now, there is an idea to add attempt index
to replace attempt_id. Therefore, session_id is now becoming a stable
concept. `session_id` is used both in REST API and CLI interface.

Fixes #469.


cosmok:
Remeved test will also need to be added, removed here: https://github.com/treasure-data/digdag/commit/1db165b9ca78478ac7579363c0342cb5605f25df

Would be great if this was part of the next release!

frsyuki:
@cosmok that's a good point. thank you for pointing it out

Issue #356, opened 2016-10-19
SMTP auth always fails


Yoshihiko-Miyaichi:
I want to use Google Apps's SMPT server for sending mails.
For this i have an own domain and it is hosted with Google Apps
e.g.: example.com
When using now a user with the right password and a username like admin@example.com, the sending fails with Google Apps reporting an Authentication Error with username or password wrong.

This was tested with:
1)mail.dig
_export:
    mail.host: smtp.gmail.com
    mail.port: 587
    mail.from: admin@example.com
    mail.username: admin@example.com
    mail.password: PASSWORD
    mail.debug: true
+digdag:
  mail>: mail.txt
  subject: test mail
  to: admin@example.com

2.) digdag run mail.dig
Here on debug the error was given:

DEBUG SMTP: protocolConnect login, host=smtp.gmail.com, user=admin@example.com, password=<non-null>
DEBUG SMTP: Attempt to authenticate using mechanisms: LOGIN PLAIN DIGEST-MD5 NTLM XOAUTH2 
DEBUG SMTP: Using mechanism LOGIN
DEBUG SMTP: AUTH LOGIN command trace suppressed
DEBUG SMTP: AUTH LOGIN failed

hiroyuki-sato:
I'm a Digdag user (not a developer). I tested send mail in my environment. 

I can send e-mail via `smtp.gmail.com` with the following step.

## Summary.

* I used `digdag server.` 
* I'm not sure  `digdag run` support  `mail` operator with SMTP auth.
  * I suppose It does not support yet.

## Gmail SMTP server setting

http://www.digdag.io/operators.html#mail-sending-email

I choose b) in this case. 

## Server configuration

### Create server configuration. 

First, create `server.conf`.

```
server.bind = xxx.xxx.xx.xxx # server IP ADDRESS
#
digdag.secret-encryption-key = dGVzdDEyMzR0ZXN0MTIzNA==
```

You an create your secret with the following command

`echo -n "test1234test1234" | openssl base64`

The secret must be 16 characters (128byte).

### Run digdag server

```
digdag server -c server.conf -m
```
**This is the on-memory server. Please use test only.**

## client setting

```
.
|-- body.txt
|-- client.conf
`-- test_mail.dig
```

### client config

```
client.http.endpoint = http://address:port
```

## create workflow

```yaml
_export:
  mail:
    host: smtp.gmail.com
    from: your_gmail_address
    port: 587
    tls: true
    debug: true
[...]

PR #479, opened 2017-02-05
Delete password config on examples/mail.dig

shigemk2:
#356 

Fix examples/mail.dig because we already use mail password not in dig file but with config `secrets.mail.password`.


frsyuki:
Thank you!

PR #480, opened 2017-02-08
Loading workflow should always override _workdir

frsyuki:
Call operator passes its all exported/stored parameters to project
archive loader. And project archive loader respects _workdir if it's
already set. But it causes unexpected behavior: there're 3 workflows A,
B, and C. A calls B, B calls C. When A calls B, it sets _workdir. When B
calls C, it doesn't set _workdir because it's already set. Therefore, C
uses B's _workdir.
This change fixes this issue by ignoring _workdir when it's already set.

### DIFF
--- a/digdag-tests/src/test/resources/acceptance/call/nested1.dig
+++ b/digdag-tests/src/test/resources/acceptance/call/nested1.dig
@@ -1,0 +1,5 @@
++nested1_out:
+  sh>: echo "${typeof _workdir}" > ${outdir}/nested1.out
+
++nested1:
+  call>: nested2/nested2.dig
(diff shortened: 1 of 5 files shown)


digdag-tests/src/test/resources/acceptance/call/nested1.dig
```diff
@@ -0,0 +1,5 @@
++nested1_out:
+  sh>: echo "${typeof _workdir}" > ${outdir}/nested1.out
```
komamitsu: Just `"${_workdir}` would be better?

komamitsu:
LGTM 👍

frsyuki:
thanks!

PR #481, opened 2017-02-09
Add normalized timezone name to RestWorkflowDefinition

frsyuki:


digdag-client/src/main/java/io/digdag/client/api/RestWorkflowDefinition.java
```diff
@@ -19,6 +21,9 @@
 
     String getRevision();
 
+    @JsonProperty("timezone")
+    ZoneId getTimeZone();
```
komamitsu: The attribute name is `timezone`, so the getter should be `getTimezone()`?

komamitsu:
I added one comment, but overall it looks good 👍

digdag-client/src/main/java/io/digdag/client/api/RestWorkflowDefinition.java
frsyuki: I took the same code from https://github.com/treasure-data/digdag/blob/082f72b3b1adcbad570950544b60fc360835dfd3/digdag-client/src/main/java/io/digdag/client/api/RestWorkflowSessionTime.java#L19-L20

frsyuki: I think using `getTimeZone` is OK for consistency with others: https://github.com/treasure-data/digdag/search?utf8=%E2%9C%93&q=getTimeZone

frsyuki:
👍

PR #482, opened 2017-02-09
Fixing tests and enhance handling of auto commit in ThreadLocalTransactionManager

komamitsu:


frsyuki:
you don't have to create PR to my PR that is PR to your PR :-) I created a PR to your PR because the original code is owned by you. You don't need to create PR to prepare your code.

PR #490, opened 2017-02-23
Recommend using Result output to S3

toru-takahashi:
table:export feature has some limitations.
For general use-case, we recommend using Result Output to S3 using td operator with result_url.


komamitsu:
👍

PR #492, opened 2017-02-24
Fix digdag-docs version and release

szyn:
Fix digdag-docs version and release because there are **not incremented**.

```
Update version from "0.8" to "0.9"
Update release from "0.8"  to "0.9.5"
```


frsyuki:
👍  Thank you!

PR #495, opened 2017-03-02
Add userInfo field to revision

frsyuki:
userInfo is information of an user who pushed the revision. UserInfo is
an opaque JSON object that is set by Authenticator module. Authenticator
module creates userInfo when it authorizes requests comming from a
client to push a new revision.
Authenticator module is supposed to be customized by system integrators.
The default Authenticator (authenticator that doesn't authenticate)
always sets an empty object.


komamitsu:
👍

PR #497, opened 2017-03-03
Fix require> operator to fail if dependent session fails

frsyuki:
`require>` operator was ignoring failure of dependent session. This is
inconsistent from `call>` operator which seems unexpected. With this
change, `require>` operator fails if dependent session is done with
success=false flag. This happens if the latest attempt of the session fails,
or is canceled.

This change also adds `ignore_failure` option to require operator so
that users can still make the workflow succeeded even when dependent
workflow fails.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
@@ -20,7 +20,7 @@ import io.digdag.core.workflow.AttemptBuilder;
 import io.digdag.core.workflow.WorkflowExecutor;
 import io.digdag.core.workflow.SessionAttemptConflictException;
 import io.digdag.core.session.SessionStoreManager;
-import io.digdag.core.session.AttemptStateFlags;
+import io.digdag.core.session.StoredSessionAttempt;
 import io.digdag.core.session.StoredSessionAttemptWithSession;
 import io.digdag.core.repository.StoredRevision;
 import io.digdag.core.repository.ArchiveType;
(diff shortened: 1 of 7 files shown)


digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
```diff
@@ -184,12 +184,13 @@ public AttemptStateFlags startSession(
                 Optional.absent());
 
         // TODO FIXME SessionMonitor monitors is not set
+        StoredSessionAttemptWithSession attempt;
         try {
-            StoredSessionAttemptWithSession attempt = exec.submitWorkflow(siteId, ar, def);
-            return attempt.getStateFlags();
+            attempt = exec.submitWorkflow(siteId, ar, def);
         }
         catch (SessionAttemptConflictException ex) {
```
komamitsu: Hmm... It's a bit weird that `SessionAttemptConflictException` is thrown when dependent workflow fails.

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -97,4 +100,34 @@ public void testRequire()
         // Verify that the file created by the child workflow is there
         assertThat(Files.exists(childOutFile), is(true));
     }
+
+    @Test
```
komamitsu: 👍

komamitsu:
👍

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
frsyuki: no, submitWorkflow doesn't throw `SessionAttemptConflictException` when a workflow fails. It throws `SessionAttemptConflictException` when attempt name conflicts.

PR #498, opened 2017-03-03
executor.task_ttl should not kill tasks BLOCKED, PLANNED or READY for long time

frsyuki:
BLOCKED and PLANNED states mean that there're other tasks running in the
attempt (BLOCKED: running dependent tasks, PLANNED: running child tasks).
Purpose of task_ttl is killing tasks without any progress for long time,
killing BLOCKED / PLANNED tasks don't make sense. Instead, it should
only kill leaf tasks.

READY is also excluded because having task in READY state for long time means
that the workflow is progressing expectedly but simply number of threads is
insufficient.

To kill long-running attempt, attempt_ttl should be used instead.

### DIFF
--- a/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
+++ b/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
@@ -45,6 +45,14 @@ public class WorkflowExecutionTimeoutEnforcer
     private static final Duration DEFAULT_TASK_TTL = Duration.ofDays(1);
     private static final Duration DEFAULT_REAPING_INTERVAL = Duration.ofSeconds(5);
 
+    // This is similar to TaskStateCode.notDoneStates() but BLOCKED and PLANNED are excluded
+    private static final TaskStateCode[] TaskTTLEnforcedTaskStates = new TaskStateCode[] {
+        TaskStateCode.READY,
+        TaskStateCode.RETRY_WAITING,
+        TaskStateCode.GROUP_RETRY_WAITING,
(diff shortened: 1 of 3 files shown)


digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
```diff
@@ -45,6 +45,15 @@
     private static final Duration DEFAULT_TASK_TTL = Duration.ofDays(1);
     private static final Duration DEFAULT_REAPING_INTERVAL = Duration.ofSeconds(5);
 
+    // This is similar to TaskStateCode.notDoneStates() but BLOCKED, PLANNED, and READY are excluded.
+    // BLOCKED and PLANNED are excluded because there're other running tasks that should be enforced instead.
+    // READY is excluded the workflow itself is working correctly and number of threads is insufficient.
+    private static final TaskStateCode[] TaskTTLEnforcedTaskStates = new TaskStateCode[] {
```
komamitsu: This variable name should start with lower case `t` instead of `T`

komamitsu: Ah, static variable... So all upper cases would be better

komamitsu:
👍 , although I left a minor comment

digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
frsyuki: that's true

PR #503, opened 2017-03-07
[WIP] Reuse ScriptEngine to avoid leak of compiled expressions in Nashorn

frsyuki:
Evaluating moment.js using a new ScriptEngine is causing memory leak.
Number of idk.nashorn.internal.ir.BinaryNode instance keeps increasing.
Sharing one ScriptEngine solves it but different workflows should not
share context for security reasons.
This change shares make it possible by freezing global context of
ScriptEngine.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
@@ -1,30 +1,39 @@
 package io.digdag.core.agent;
 
-import java.util.List;
-import java.util.Map;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.charset.Charset;
-import javax.script.ScriptEngine;
-import javax.script.ScriptEngineManager;
-import javax.script.ScriptException;
-import javax.script.Invocable;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.node.ObjectNode;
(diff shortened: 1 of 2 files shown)


frsyuki:
After many trial & errors, I found that it is difficult to make ScriptEngine instance immutable...
The reason is that there're no approaches to freeze global context so that a script evaluation can't have side effect (each evaluations don't affect each other).
For example, this works unexpected with this patch:

```
+set_global_a:
  echo>: ${a=1}

+show_global_a:
  echo>: ${typeof a}

+set_prototype_b:
  echo>: ${moment.b=2}

+show_prototype_b:
  echo>: ${typeof moment.b}
```

Expected result is that both `+show_global_a` and `+show_moment_b` show "undefined". However, they both show "number". It means that task evaluation can have side effects.

This reply on nashorn ML mentions the same thing. But freezing prototype of all standard built-in constructors seems unrealistic approach.
http://mail.openjdk.java.net/pipermail/nashorn-dev/2014-October/003753.html

frsyuki:
Other approaches we could take are:

* (A) Making creation of ScriptEngine lazy so that ScriptEngine instance is created only when a variable contains `${...}` syntax. This is a neat optimization but not perfect solution because memory still bloats if tasks in a loop contains `${...}` syntax. 99% of tasks in loops should have `${...}` because loop usually needs a variable.
* (B) Add a command-line option to enable ScriptEngine sharing. But this is not good idea because we can't enable it by default so that users don't depend on mutable global variables unexpectedly on local mode before pushing a workflow to a server which should disable ScriptEngine sharing.
* (C) Fix Nashorn not to cause memory leak.

frsyuki:
https://github.com/treasure-data/digdag/pull/1256 is going to solve this problem fundamentally by changing JavaScript engine to Graal.js.

PR #505, opened 2017-03-08
Add result_connection and result_settings to td> operator

frsyuki:
This depends on https://github.com/treasure-data/td-client-java/pull/83


digdag-standards/src/main/java/io/digdag/standards/operator/td/IdOrNameParam.java
```diff
@@ -0,0 +1,112 @@
...
+import java.util.Objects;
+import java.util.regex.Pattern;
+import static com.google.common.base.Preconditions.checkState;
+
+public class IdOrNameParam
+{
+    private static final Pattern ID_PATTERN = Pattern.compile("0|[1-9][0-9]*");
+
+    private final String name;
+    private final long id;
```
komamitsu: Using `Long` instead might be better since it accepts `null`

digdag-standards/src/main/java/io/digdag/standards/operator/td/IdOrNameParam.java
```diff
@@ -0,0 +1,112 @@
...
+
+    @JsonCreator
+    public static IdOrNameParam of(JsonNode expr)
+    {
+        if (expr.isIntegralNumber()) {
+            return new IdOrNameParam(null, expr.asLong());
+        }
+        else if (expr.isTextual()) {
+            String text = expr.asText();
+            if (ID_PATTERN.matcher(text).matches()) {
```
komamitsu: If the connection name (not id) is `"1234"`, `td` operator tries to handle it as a job id and fails?

digdag-standards/src/test/java/io/digdag/standards/operator/td/TdOperatorFactoryTest.java
```diff
@@ -124,4 +124,38 @@ public void rejectQueryFileOutsideOfProjectPath()
...
+            throws Exception
+    {
+        Path projectPath = Paths.get("").normalize().toAbsolutePath();
+
+        Config config = newConfig()
+            .set("database", "testdb")
+            .set("query", "select 1")
+            .set("result_connection", "18446744073709551616");
+
+        exception.expectMessage("for key 'result_connection' but got \"18446744073709551616\"");
```
komamitsu: This message seems ambiguous to me although I know the changes of this pull request don't include the error message...

komamitsu:
LGTM overall 👍 

But I have these two concerns:
- https://github.com/treasure-data/digdag/pull/505/files#r104834838
- There is no test that actually sees if `result_connection` works expectedly

digdag-standards/src/main/java/io/digdag/standards/operator/td/IdOrNameParam.java
frsyuki: `"1234"` will be handled as an connection id.
There're 3 reasons to do this:

* `${...}` syntax always returns string. If here uses `"1234"` as an name of connection, there're no ways to parameterize it as an id.
* when web console or external programs generate .dig files, they may not care about types strictly (e.g. text input box of web console is always text even though users put integer there). Using different behavior depending on type causes troubles with it.
* `ObjectMapper.readValue(it, long.class)` successfully converts `"1234"` into long.

digdag-standards/src/test/java/io/digdag/standards/operator/td/TdOperatorFactoryTest.java
frsyuki: right, I had a problem with testing in fact...
In actual use case, digdag command shows more messages like this:

```
2017-03-07 17:11:42 -0800 [ERROR] (0018@+ab+a): Configuration error at task +ab+a: Expected [simple type, class com.google.common.base.Optional<io.digdag.standards.operator.td.IdOrNameParam>] for key 'result_connection' but got 18446744073709551616 (number) (config)
> Instantiation of [simple type, class io.digdag.standards.operator.td.IdOrNameParam] value failed: Must be id or name: 18446744073709551616 (json mapping)
2017-03-07 17:11:43 -0800 [INFO] (0018@+ab^failure-alert): type: notify
error:
  * +ab+a:
    Expected [simple type, class com.google.common.base.Optional<io.digdag.standards.operator.td.IdOrNameParam>] for key 'result_connection' but got 18446744073709551616 (number) (config)
```

It is great if we can easily tell the cause of type mapping failure to Jackson but it's complicated.

digdag-standards/src/main/java/io/digdag/standards/operator/td/IdOrNameParam.java
frsyuki: maybe we should change this as following:

`result_connection: name` always uses `name` as name. `result_connection: {id: N}` or `result_connection_id: N` uses `N` as id.

frsyuki:
👍  Added a commit to address.

PR #511, opened 2017-03-14
Fix `digdag retry --resume` option in document

akirakw:
I think `digdag retry --resume` cannot specify the task name option.


komamitsu:
Good catch! Thanks

https://github.com/treasure-data/digdag/blob/master/digdag-cli/src/main/java/io/digdag/cli/client/Retry.java#L43

PR #515, opened 2017-03-16
Use ThreadLocal to store a database transaction handle

komamitsu:


komamitsu:
@frsyuki Can you review this pull request when you get a chance? I added some calls of `TransactionManager#reset` in `catch` blocks. I want you to check them especially. Thanks

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -110,24 +118,27 @@ public void run(TaskRequest request)
     {
         long taskId = request.getTaskId();
 
-        // set task name to thread name so that logger shows it
-        try (SetThreadName threadName = new SetThreadName(request.getTaskName())) {
-            try (TaskLogger taskLogger = callback.newTaskLogger(request)) {
-                TaskContextLogging.enter(LogLevel.DEBUG, taskLogger);
-                try {
-                    runningTaskMap.put(taskId, request);
+        tm.begin(() -> {
```
frsyuki: I think that starting transaction here is not good idea because the task runs in a transaction. All locks acquired in this transaction won't be released until the user task finishes.
`TaskCallbackApi` (and `SecretProvider`) are the only interface for agents to access database. Therefore, you can put `tm.begin` in TaskCallbackApi` to make all DB access from tasks transactional.

frsyuki: maybe like this: https://gist.github.com/frsyuki/7c62437dd55cefa0b343dcf593d8630e

digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
```diff
@@ -482,46 +484,62 @@ public void runWhile(BooleanSupplier cond)
...
-                    propagateSessionArchive();
+            //IncrementalStatusPropagator prop = new IncrementalStatusPropagator(date);  // TODO doesn't work yet
+            final AtomicInteger waitMsec = new AtomicInteger(INITIAL_INTERVAL);
+            while (true) {
+                if (tm.<Boolean>begin(() -> !cond.getAsBoolean())) {
+                    break;
                 }
-                else {
+
+                tm.begin(() -> {
```
frsyuki: I think here uses smaller transactions to avoid potential deadlocks.
Smaller transaction means almost one transaction per database operation.
For example,

* propagateBlockedChildrenToReady: one transaction for sm.findDirectParentsOfBlockedTasks, separated transactions for each sm.lockTaskIfExists calls.
* propagateAllPlannedToDone: one transaction for sm.findTasksByState, separated transactions for each sm.lockTaskIfExists calls.
* enqueueReadyTasks: one transaction per task

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
komamitsu: 👍 
BTW, I don't get this existing comment `use the HTTP request time as the runTime`. I have no idea what `the HTTP request` exactly means...
https://gist.github.com/frsyuki/7c62437dd55cefa0b343dcf593d8630e#file-task-callback-transaction-diff-L201

komamitsu:
@frsyuki Addressed your comments on OperatorManager.

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
frsyuki: I think the intention there is that `ScheduleTime.runNow(instant)` should be replaced by something because it is not deterministic. It's not very important....

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
```diff
@@ -96,23 +100,24 @@ public TaskLogger newTaskLogger(TaskRequest request)
...
         String revision = request.getRevision().get();
 
         try {
-            return archiveManager.openArchive(
-                    pm.getProjectStore(request.getSiteId()),
-                    request.getProjectId(),
-                    revision);
+            return tm.begin((TransactionManager.SupplierInTransaction<
+                            Optional<StorageObject>, ResourceNotFoundException,
+                            StorageFileNotFoundException, RuntimeException>) () ->
```
frsyuki: This is very minor but Instead of casting the lambda type, you can use this syntax:

```java
tm.<Optional<StorageObject>, ResourceNotFoundException, StorageFileNotFoundException>begin(() -> {
  ...
},
ResourceNotFoundException.class, StorageFileNotFoundException.class);
```

I wish that javac becomes smart enough to guess the types automatically...

komamitsu: 👍 👍 👍

PR #519, opened 2017-03-22
Modify td_table_export operator explanation

kamikaseda:
I found not matched sentence. It seems copied from td_load operator. Could you review and merge?


frsyuki:
👍

PR #520, opened 2017-03-22
Fix old path

hiroyukim:
I'd like to fix it because the path of the management screen is old.


komamitsu:
@dzannotti Can you take a look at this pull request? Thanks

dzannotti:
I think that both paths are valid, but sure 👍  lgtm

PR #522, opened 2017-03-23
Avoid unexpected validate because of using dynamic parameters

szyn:
I found a bug that digdag-cli's `--params-file` option broken.

According to [JCommander - 22. Dynamic parameters](http://jcommander.org/#_dynamic_parameters),
  
> JCommander allows you to specify parameters that are not known at compile time, such as "-Da=b -Dc=d". 

So that, this issue is caused by a `--params-file` option that is to be included in dynamic parameter (`--params`).

e.g. When I want to start digdag server (using params file)...

```
### I want to use these params 
$ cat params/develop.yml
---
ENV: develop
DEBUG: true

### OK
$ digdag server --bind 0.0.0.0 --memory -P params/develop.yml
2017-03-23 02:43:00 +0000: Digdag v0.9.6
2017-03-23 02:43:06 +0000 [INFO] (main): secret encryption engine: disabled
2017-03-23 02:43:07 +0000 [INFO] (main): XNIO version 3.3.6.Final
2017-03-23 02:43:07 +0000 [INFO] (main): XNIO NIO Implementation Version 3.3.6.Final
2017-03-23 02:43:07 +0000 [INFO] (main): Starting server on 0.0.0.0:65432
2017-03-23 02:43:07 +0000 [INFO] (main): Bound api on /0:0:0:0:0:0:0:0:65432

### NG
$ digdag server --bind 0.0.0.0 --memory --params-file params/develop.yml
2017-03-23 02:43:33 +0000: Digdag v0.9.6
error: Dynamic parameter expected a value of the form a=b but got:s-file # Dynamic perameter's validation error
```

So, I created ParameterValidator class and using `@Parameter` instead of `@DynamicParameter` and List covert to Map.   
Also, these params validate key=value.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/Check.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/Check.java
@@ -1,6 +1,5 @@
 package io.digdag.cli;
 
-import com.beust.jcommander.DynamicParameter;
 import com.beust.jcommander.Parameter;
 import com.google.common.base.Optional;
 import com.google.inject.Injector;
@@ -28,8 +27,10 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.time.Instant;
 import java.time.ZoneId;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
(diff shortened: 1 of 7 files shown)


komamitsu:
@frsyuki PTAL~

frsyuki:
LGTM 👍  Thank you!

PR #526, opened 2017-03-28
Add SecretControlStore.lockProjectSecret

frsyuki:
Some secret operations such as refreshing OAuth token need inter-sever
mutual exclusion. This new method provides locking without depending on
another external components.


digdag-core/src/main/java/io/digdag/core/database/DatabaseSecretControlStore.java
```diff
@@ -50,15 +51,81 @@ public void setProjectSecret(int projectId, String scope, String key, String val
...
+
+            // TODO this logic is copied from DatabaseSecretStore.getSecret.
+            Optional<String> value;
+            if (secret == null) {
+                value = Optional.absent();
+            }
+            else {
+                // TODO: look up crypto engine using name
+                if (!crypto.getName().equals(secret.engine)) {
+                    throw new AssertionError("Crypto engine mismatch");
```
komamitsu: Adding the values of `crypt.getName()` and `secret.engine` into the message would be better.

digdag-core/src/test/java/io/digdag/core/database/DatabaseSecretStoreTest.java
```diff
@@ -82,6 +82,18 @@ public void singleSecret()
...
 
     @Test
+    public void lockSecret()
+            throws Exception
+    {
+        Optional<String> absent = factory.begin(() -> secretControlStore.lockProjectSecret(projectId, SecretScopes.PROJECT, KEY1, (control, value) -> value));
+        assertThat(absent, is(Optional.absent()));
+
+        factory.begin(() -> secretControlStore.setProjectSecret(projectId, SecretScopes.PROJECT, KEY1, VALUE1));
+        Optional<String> present = factory.begin(() -> secretControlStore.lockProjectSecret(projectId, SecretScopes.PROJECT, KEY1, (control, value) -> value));
```
komamitsu: We'd better have a test that confirms if another thread is blocked with the method

komamitsu:
@frsyuki I left a comment https://github.com/treasure-data/digdag/pull/526/files#r108588806 and another minor comment. Can you look at the first one?

digdag-core/src/test/java/io/digdag/core/database/DatabaseSecretStoreTest.java
frsyuki: Good point. I added a test.

digdag-core/src/main/java/io/digdag/core/database/DatabaseSecretControlStore.java
frsyuki: yes, true. I don't know why here uses AssertionError though...

komamitsu:
👍

PR #527, opened 2017-03-28
Do more retry to be more reliable on undeterministic errors

frsyuki:
Some components contain unreliable behavior. For example, S3Storage
essentially fails and 3.5 seconds (3 retrying with 0.5 seconds initial
wait) isn't long enough. 3.5 seconds seems too short for most of
undeterministic errors.
This change makes it 121 seconds (5 retrying with 1 second initial wait
and 3x grow rate).


komamitsu:
LGTM 👍

PR #529, opened 2017-03-30
Add { credentials: 'include' } to fetch in JS

morizyun:
I added `{ credentials: 'include' }` to fetch(url) option in JavaScript, due to using Digdag-UI with basic authentication.
https://developers.google.com/web/updates/2015/03/introduction-to-fetch#sending_credentials_with_a_fetch_request

Currently, we catched the following errors:

```
Uncaught (in promise) Error: Unauthorized at console.jsx:1742
Error: Unauthorized at model.js:486
```

Thank you for your cooperation, in advance!

### DIFF
--- a/digdag-ui/console.jsx
+++ b/digdag-ui/console.jsx
@@ -1737,7 +1737,7 @@ class VersionView extends React.Component {
 
   componentDidMount () {
     const url = DIGDAG_CONFIG.url + 'version'
-    fetch(url).then(response => {
+    fetch(url, { credentials: 'include' }).then(response => {
       if (!response.ok) {
         throw new Error(response.statusText)
       }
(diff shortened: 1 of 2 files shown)


dzannotti:
Hi @morizyun thanks for this PR!
I think that it would be better (and simplier) if we moved the fetch from being called directly in console.jsx to inside model.jsx, using the method `http` from model which already includes `credentials: true`

Let me know if you wish to repurpose this PR for that change, or if you wish for me to make the change in a separate PR

morizyun:
@dzannotti Thank you very much for your kind proposal. 
Sounds great! That would be better. Please make new PR to solve this problem.

I am going to close this proposal.

PR #531, opened 2017-03-30
UI: Fix archive display

dzannotti:
This adds support to correctly display py files referenced from spark and script operators and sql files referenced in emr operators


frsyuki:
👍 
spark script is not always python but might be spark script (shell script): http://docs.digdag.io/operators/emr.html. But maybe it's not a big issue if it doesn't guess type file type accurately.

PR #532, opened 2017-03-31
UI: Use model for versionview

dzannotti:
This is tightly related to https://github.com/treasure-data/digdag/pull/529 and @morizyun requirement/efforts.  Looking at the PR i noticed that `VersionView` was calling fetch directly (when it should call it thru `model` instead


frsyuki:
👍  Nice improvements.

PR #533, opened 2017-03-31
/api/attempts?project=&workflow= should lookup workflow by name

frsyuki:
A workflow in a project is identified by projectId and name because
workflow ID changes every revision. Therefore, listing up attempts by
workflow name should not use workflow ID when querying on database.
Otherwise, before this fix, the REST API was listing up attempts that
were exected by the latest revision only.

Scheduler's `skip_on_overtime` option also had the same problem. It
should list up workflows by name so that workflows with the same name
shouldn't run in parallel. It's unexpected if a workflow doesn't get
skipped when a new revision is pushed even though the workflow name or
definition is not changed.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -1145,20 +1145,20 @@ public class DatabaseSessionStoreManager
         }
 
         @Override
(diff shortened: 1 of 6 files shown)


komamitsu:
Let me know the background more deeply.

For the first paragraph, what are actual cases you're are concerned? `/api/attempts` can return an empty result? If so, how can we reproduce it?

komamitsu:
Ah, got it. It can be easily happened after pushing a new revision of the project even the project has attempts before.

komamitsu:
LGTM 👍

frsyuki:
👍  Thank you!

PR #535, opened 2017-04-05
Fix admin port restriction when server listens on 0.0.0.0

frsyuki:
`control.getListenAddresses` was returning listen addresses taken from
`Undertow.ListenerInfo`. On the other hand,
`HttpServletRequest.getAddress()` returns socket's local address.
They become a same address if server listens on a specific IP such as
`192.168.0.1`. However, if server listens on `0.0.0.0`, listen address
is `0.0.0.0` but socket's local address is `127.0.0.1`. Therefore,
AdminRestrictedFilter rejects all requests if server listens on 0.0.0.0.

To solve this issue,, this change removes use of listen address or
socket address. Instead, it works as following:

1. When Undertow handles a request, it sets listen address name on
   `HttpServerExchange` instance. This is done by
   `UndertowServer.SetListenAddressNameHandler` class.

2. `HttpServerExchange` instance is not visible from Servlet interface
   (or RESTEasy which is running on top of Servlet) because it's
   Undertow's interface. Therefore, a servlet filter
   `UndertowServer.SetListenAddressNameServletFilter` taskes listen
   address name from `HttpServerExchange`, and sets it to
   `ServletRequest`.

3. Servlet applications can get the name using
  `Object ServletRequest.getAttribute(String)` method.
  JAX-RS applications can use the name using
  `Object ContainerRequestContext.getProperty(String)` method
  which internally calls `ServletRequest.getAttribute(String)`.

### DIFF
--- a/digdag-guice-rs-server-undertow/src/main/java/io/digdag/guice/rs/server/undertow/UndertowServer.java
+++ b/digdag-guice-rs-server-undertow/src/main/java/io/digdag/guice/rs/server/undertow/UndertowServer.java
@@ -5,6 +5,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import io.digdag.guice.rs.GuiceRsBootstrap;
+import io.digdag.guice.rs.GuiceRsServerRuntimeInfo;
 import io.digdag.guice.rs.GuiceRsServletContainerInitializer;
 import io.digdag.guice.rs.server.ServerBootstrap;
 import io.digdag.guice.rs.server.undertow.UndertowServerConfig.ListenAddress;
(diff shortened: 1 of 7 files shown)


digdag-guice-rs-server-undertow/src/main/java/io/digdag/guice/rs/server/undertow/UndertowServer.java
```diff
@@ -86,6 +96,55 @@ ServerBootstrap getServerBootstrap()
...
+        public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+                throws IOException, ServletException
+        {
+            if (request instanceof HttpServletRequestImpl) {
+                HttpServerExchange exchange = ((HttpServletRequestImpl) request).getExchange();
+                String name = exchange.getAttachment(LISTEN_ADDRESS_NAME_ATTACHMENT);
+                if (name != null) {
+                    request.setAttribute(GuiceRsServerRuntimeInfo.LISTEN_ADDRESS_NAME_ATTRIBUTE, name);
+                }
+                chain.doFilter(request, response);
```
komamitsu: This line should be outside of the `if` block?

komamitsu:
Added one comment https://github.com/treasure-data/digdag/pull/535/files#r109865693.
But overall LGTM 👍

frsyuki:
👍

digdag-guice-rs-server-undertow/src/main/java/io/digdag/guice/rs/server/undertow/UndertowServer.java
frsyuki: According to JavaDoc, here needs to call `chain.doFilter` if here doesn't want to block request processing.

https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/Filter.html#doFilter(javax.servlet.ServletRequest, javax.servlet.ServletResponse, javax.servlet.FilterChain)

PR #537, opened 2017-04-06
Add `task_name` to built-in variables

komamitsu:
With this feature, users can use a task name in workflow like this:
```
+task1:
    td>: queries/${task_name.replace(/\+/g, '_')}.sql    # >>> "_myworkflow_task1.sql"
    insert_into: hourly_summary
```

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/RuntimeParams.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/RuntimeParams.java
@@ -4,10 +4,6 @@ import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
-import java.util.Arrays;
-import java.util.LinkedList;
-import java.util.List;
-
 import io.digdag.client.config.Config;
 import io.digdag.client.config.ConfigException;
 import io.digdag.client.config.ConfigFactory;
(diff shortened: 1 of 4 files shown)


komamitsu:
@frsyuki Can you take look at this pull request when you get a chance?

frsyuki:
Very nice :+1:
I wondered whether the task name should include the root task name, or not.
In `+built_in_variables+get_variables+task_name`, `+built_in_variables` is not precisely task name because it's workflow name. However, internally, workflow name is used as the "root" task name. So it's a sort of task name.
I honestly don't have strong opinion on it. How do you think?
Code itself looks very good.

komamitsu:
Yeah. It's probably better to remove a workflow name from it since users don't know a workflow name is internally included in a task name. I'll modify the pull request later.

komamitsu:
Separated the workflow name from a task name like this.
```
+parent:
  +task0:
      td>: queries/${workflow_name}/${task_name.join('__')}.sql     # >>> "queries/mydig/parent__task0.sql"
      insert_into: hourly_summary
```
@frsyuki Could you look over the updated one?

frsyuki:
@komamitsu yes, LGTM 👍

frsyuki:
Oh Mitsu, I recalled one problem. Task name delimiter is not always `+`. Dynamically generated tasks use `^` as a delimiter (e.g. `^sub`, `^failure-alert`). So, removing `+` actually loses some information...

frsyuki:
How about just using the first idea? It seems most straightforward way.
Calling `task_name.replace(/\W/g, '_')}` or `tsak_name.split(/\W/).join('__')` still works.

komamitsu:
Ah. Good catch!

> How about just using the first idea? It seems most straightforward way.

👍

PR #542, opened 2017-04-19
Add site ID and project name to a thread name

komamitsu:
so that we can easily find which site ID and project is affected by
an issue searching log files


komamitsu:
@frsyuki PTAL

frsyuki:
@komamitsu very minor thing but what does `orig` in the variable name mean here..? It's not "original" (like saved original thread name), right? Because it's a new name of the thread.

komamitsu:
@frsyuki As you know, `SetThreadName` concatenates a thread id and a passed name, in this case it's a task name, and set the concatenated one as a thread name. So `origThreadName` is actually used as a part of a thread name, but it's not the same as a thread name created in `SetThreadName` and I used prefix `orig`. But I'm not sure it's the best name. Any suggestion is welcome :)

PR #543, opened 2017-04-19
Print the stacktrace of TransactionHandle

komamitsu:
Which is stored when ThreadLocalTransactionManager#begin is called first.
It's useful to investigate "Nested transaction is not allowed" error.


komamitsu:
@frsyuki Can you take a look at this pull request?

PR #545, opened 2017-04-24
Modify triple hyphen to double hyphen for project option in Document

kamikaseda:
Document for command option has triple hyphen "---" in project option like "---project". Correct option is double hyphen like "--project".


frsyuki:
Thank you! 👍

PR #546, opened 2017-04-24
Documentation: quote example options

kamikaseda:
Example of Command Options shows single hyphen in spite of that source code has double hyphen. It should be quoted by double back quotation as command. It is able to show correct option format.


komamitsu:
@kamikaseda Thanks!

PR #547, opened 2017-04-24
Fix character encoding of email subject

frsyuki:
`mail>` operator was not setting character encoding to the subject which
means that it was using platform's default encoding. If the platform's
default encoding is UTF-8, it's not a problem. However, especially if
digdag is running in a minimal docker container, characters that can't
be encoded using the default encoding are replaced by `?`.


komamitsu:
👍

PR #549, opened 2017-04-26
Fixed mark() implementation of ResumableInputStream

frsyuki:
Actually removed because it is completely optional to implement.

### DIFF
--- a/digdag-plugin-utils/src/test/java/io/digdag/util/ResumableInputStreamTest.java
+++ b/digdag-plugin-utils/src/test/java/io/digdag/util/ResumableInputStreamTest.java
@@ -1,0 +1,291 @@
+package io.digdag.util;
+
+import com.google.common.base.Optional;
+import io.digdag.client.config.ConfigException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.ArrayDeque;
+import java.util.Arrays;
+import java.util.Deque;
+import org.junit.Test;
+
+import static org.hamcrest.Matchers.is;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.junit.Assert.assertThat;
+import static org.junit.Assert.fail;
+
+public class ResumableInputStreamTest
+{
+    static final int IO_EXCEPTION = -2;
+    static final int RUNTIME_EXCEPTION = -3;
+
+    static class TestingInputStream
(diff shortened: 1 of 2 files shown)


digdag-plugin-utils/src/test/java/io/digdag/util/ResumableInputStreamTest.java
```diff
@@ -0,0 +1,291 @@
...
+        @Override
+        public int read() throws IOException
+        {
+            return throwIfErrorCommand(returnValues.removeFirst());
+        }
+
+        @Override
+        public int read(byte[] b, int off, int len) throws IOException
+        {
+            return throwIfErrorCommand(returnValues.removeFirst());
```
komamitsu: Returning a number more than `len` can confuse me a bit. `skip()` is as well.

digdag-plugin-utils/src/test/java/io/digdag/util/ResumableInputStreamTest.java
```diff
@@ -0,0 +1,291 @@
...
+        SimpleReopener reopener = new SimpleReopener(next);
+        ResumableInputStream in = new ResumableInputStream(next, reopener);
+        assertThat(in.read(), is(1));
+        assertThat(reopener.getLastOffset(), is(0L));
+        assertThat(reopener.getReopenCount(), is(5));
+        assertThat(in.read(), is(2));
+        assertThat(reopener.getLastOffset(), is(1L));
+        assertThat(reopener.getReopenCount(), is(6));
+        assertThat(in.read(), is(-1));
+        assertThat(reopener.getReopenCount(), is(6));
```
komamitsu: It seems you can check the last offset here.

komamitsu:
Added some minor comments on ResumableInputStreamTest, but totally LGTM 👍 

BTW, `in` of ResumableInputStream can be null without synchronization at `reopen()`. So other methods (e.g. `close()`) may run into NPE in multithreads environment. Maybe we'd better synchronize those methods in ResumableInputStream later to avoid potential problems.

frsyuki:
@komamitsu  Good points! I added a follow-up commit: https://github.com/treasure-data/digdag/commit/18d9012aa27806efafcee37f648ca78d72d4b1de

PR #554, opened 2017-05-08
td_table_export> supports some inefficients format but they're obsoleted

frsyuki:
* `json-line.gz`: simply wrong. `line-json.gz` is the correct name.
* `line-json.gz`: obsoleted implementation. `jsonl.gz` is much faster alternative.
* `json.gz`: obsoleted implementation.


frsyuki:
depends on https://github.com/treasure-data/td-client-java/pull/87

komamitsu:
👍

PR #560, opened 2017-05-12
Release digdag-ui package

frsyuki:
This lets system integrators serve digdag-ui beside REST API while they
inject custom modules.


komamitsu:
👍

PR #563, opened 2017-05-20
Fix typo on document

shio-phys:
fix typo


dzannotti:
@shio-phys thanks for contributing!

PR #564, opened 2017-05-23
Add tips for mail.password parameter

toru-takahashi:
Sometimes, we get an inquiry about this warning. > `Unsecure 'password' parameter is deprecated.`

For example, https://github.com/treasure-data/digdag/issues/356


frsyuki:
👍

PR #566, opened 2017-05-24
fix config doc

akirakoyasu:
fix server.http.io-threads default

ref: https://github.com/treasure-data/digdag/blob/7c8c695/digdag-guice-rs-server-undertow/src/main/java/io/digdag/guice/rs/server/undertow/UndertowServer.java#L202


frsyuki:
:+1: Thank you

PR #574, opened 2017-06-04
Add timeout parameter for http operator

duck8823:
This PR add timeout parameter for http operator.
Any API server response slowly, but http operator set timeout 30 seconds.
I want to set timeout in http operator.

The way to set timeout is...
```yaml
http>: http://example.com
timeout: 60
```
in .dig file.

regards.

### DIFF
--- a/digdag-standards/src/test/java/io/digdag/standards/operator/HttpOperatorFactoryTest.java
+++ b/digdag-standards/src/test/java/io/digdag/standards/operator/HttpOperatorFactoryTest.java
@@ -1,6 +1,7 @@
 package io.digdag.standards.operator;
 
 import com.github.tomakehurst.wiremock.junit.WireMockRule;
+import io.digdag.client.config.Config;
 import io.digdag.spi.Operator;
 import io.digdag.spi.TaskExecutionException;
 import org.junit.Before;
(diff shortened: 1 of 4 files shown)


komamitsu:
Thanks.

It seems this pull request ran into an error https://travis-ci.org/treasure-data/digdag/jobs/239227299#L2509.

I sent a pull request to fix that https://github.com/duck8823/digdag/pull/1. Can you take a look at it?

digdag-standards/src/test/java/io/digdag/standards/operator/HttpOperatorFactoryTest.java
```diff
@@ -0,0 +1,77 @@
...
+import static io.digdag.core.workflow.WorkflowTestingUtils.loadYamlResource;
+import static org.junit.Assert.fail;
+
+public class HttpOperatorFactoryTest
+{
+	@Rule
+	public TemporaryFolder folder = new TemporaryFolder();
+
+	@Rule
+	public WireMockRule wireMockRule = new WireMockRule(8089);
```
komamitsu: If another process is using this port number, this test fails, right? How about using a random port number instead? Maybe passing 0 would work.

duck8823: Yes. this test fails if use conflicted port.
I changed to use random port in this test.

duck8823:
Thank you for your review.
I fixed port number in test cases.

komamitsu:
LGTM! Thanks

PR #578, opened 2017-06-13
Add note on whitespace between !include and :

yoichi:
I think it is convenient to add a note, because I also had a problem described in:
Https://github.com/treasure-data/digdag/issues/548

Best Regards,


frsyuki:
Looks good 👍

PR #590, opened 2017-07-08
Remove unsupported "Accept-Encoding: deflate" from client requests

frsyuki:
DigdagClient uses apache httpclient and it adds
`Accept-Encoding: gzip, deflate` autoamtically:

Server:

```
$ nc -l 8765
GET /api/version HTTP/1.1
Accept-Encoding: gzip, deflate
User-Agent: DigdagClient/0.9.12
Host: localhost:8765
Connection: Keep-Alive
```

Client command: `digdag -e http://localhost:8765 version`

However, if server encodes using deflate and returns
`Content-Encoding: deflate` header, client causes an error:

```
        at org.jboss.resteasy.client.jaxrs.internal.ClientResponse.readFrom(ClientResponse.java:285)
        at org.jboss.resteasy.client.jaxrs.internal.ClientResponse.readEntity(ClientResponse.java:181)
        at org.jboss.resteasy.specimpl.BuiltResponse.readEntity(BuiltResponse.java:219)
        at io.digdag.client.DigdagClient.doGet(DigdagClient.java:885)
        at io.digdag.client.DigdagClient.getVersion(DigdagClient.java:809)
        at utils.TemporaryDigdagServer.start(TemporaryDigdagServer.java:374)
...(snip)...
    Caused by: com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'ªV': was expecting ('true', 'false' or 'null')
     at [Source: org.apache.http.conn.EofSensorInputStream@74bed930; line: 1, column: 4]
        at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1581)
        at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:533)
        at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._reportInvalidToken(UTF8StreamJsonParser.java:3451)
        at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._handleUnexpectedValue(UTF8StreamJsonParser.java:2610)
        at com.fasterxml.jackson.core.json.UTF8StreamJsonParser._nextTokenNotInObject(UTF8StreamJsonParser.java:841)
        at com.fasterxml.jackson.core.json.UTF8StreamJsonParser.nextToken(UTF8StreamJsonParser.java:737)
        at com.fasterxml.jackson.jaxrs.base.ProviderBase.readFrom(ProviderBase.java:770)
[...]

PR #589, opened 2017-07-08
Handle "Content-Encoding: gzip | deflate" in request body in Undertow

frsyuki:
This change lets Underetow HTTP server decompress request body
compressed by the cilents.
This functionality was added to Undertow since 1.4.7.Final by
[UNDERTOW-920](https://issues.jboss.org/browse/UNDERTOW-920). Some fixes
were added by [UNDERTOW-935](https://issues.jboss.org/browse/UNDERTOW-935).

PR #591, opened 2017-07-08
Add support for response body compression

frsyuki:
Add support for response body compression …This change lets Undertow compress response body using gzip or deflate
if client sends requests with `Accept-Encoding: gzip` or
`Accept-Encoding: deflate` header.

This change assumes #589.

When client sends `Accept-Encoding: gzip, deflate`, the server uses `gzip` (gzip has higher priority). This is because the client has a bug with deflate. See #590.

### DIFF
--- a/digdag-client/build.gradle
+++ b/digdag-client/build.gradle
@@ -19,7 +19,7 @@ dependencies {
     compile 'io.jsonwebtoken:jjwt:0.6.0'
 
     // this dependency is here only to override dependency of resteasy 3.0.13 -> jboss-logging 3.1.4.GA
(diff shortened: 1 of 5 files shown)


yoyama:
@komamitsu @muga Is there any reason this PR is not merged even though approved same as #589 ?

yoyama:
Just in case, I will postpone to 0.9.41.

yoyama:
As of discussion in person, we will merge to v0_10 after v0_11 has been created.

szyn:
This change assumes #589 as the description. I will merge this PR and close it.

PR #595, opened 2017-07-11
Add note on scheduling workflow doc

shio-phys:
Some people seem to be confused when using schedule directive to run tasks every minute.
http://recruit.gmo.jp/engineer/jisedai/blog/introduction-to-digdag/
(Japanese)

The reason is that starting field with `*` does not meet valid YAML syntax.
So I added document about that.


frsyuki:
👍

PR #597, opened 2017-07-14
Let users configure default Treasure Data API endpoint

frsyuki:
This change let users configure default td.endpoint of all td-related
operators using system config. This is helpful especially for system
integrators who provides a digdag server for customers with a
pre-configured td.endpoint.

It doesn't use JVM system property because system property affects globally. It may impact badly when we want to support remote agents. On the other hand, with system config, remote agents will be able to get the config from the central server when it starts. This approach guarantees that all remote agents use consistent configuration.

There are some inconsistency around the word "endpoint" because some configuration assumes that "endpoint" contains "http" or "https", but td.endpoint doesn't. But it's ignored for now.

### DIFF
--- a/digdag-standards/src/test/java/io/digdag/standards/operator/td/TDOperatorTest.java
+++ b/digdag-standards/src/test/java/io/digdag/standards/operator/td/TDOperatorTest.java
@@ -410,8 +410,7 @@ public class TDOperatorTest
         assertThat(defaultDefaults.getEndpoint(), is("api.treasuredata.com"));
 
         Config overrideConfig = configFactory.create()
-                .set("config.td.default_endpoint", "api.treasuredata.co.jp")
-                ;
+                .set("config.td.default_endpoint", "api.treasuredata.co.jp");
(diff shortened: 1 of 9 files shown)


digdag-standards/src/test/java/io/digdag/standards/operator/td/TDOperatorTest.java
```diff
@@ -393,6 +403,19 @@ public void testRunJobMigrateState()
...
 
+    @Test
+    public void testSystemDefaultConfig()
+    {
+        SystemDefaultConfig defaultDefaults = TDOperator.systemDefaultConfig(configFactory.create());
+        assertThat(defaultDefaults.getEndpoint(), is("api.treasuredata.com"));
+
+        Config overrideConfig = configFactory.create()
+                .set("config.td.default_endpoint", "api.treasuredata.co.jp")
+                ;
```
komamitsu: Should this semicolon be put at the tail of the above line?

komamitsu:
LGTM 👍 
I left a minor comment, though

PR #598, opened 2017-07-18
Fix display format of past time by CLI

akirakw:
This PR fixes  display format of past time difference by CLI.

In the current implementation,  the difference in past time only seconds are shown.

Before fix:
```
  disabled at: 2017-07-18 16:17:11 +0900 (-32s ago)
```

After fix:

```
  disabled at: 2017-07-18 16:17:11 +0900 (16m 36s ago)
```


frsyuki:
👍  Good point. disabledAt is always past. Thank you.

PR #600, opened 2017-07-19
Add Release notes

toru-takahashi:
http://docs.digdag.io/releases.html is not updated sometimes.
So, on Github, it might be better to link to https://github.com/treasure-data/digdag/tree/master/digdag-docs/src/releases instead of the doc.


komamitsu:
Thanks!

PR #601, opened 2017-07-21
Add pull-always option to docker configuration

alu:
I added `pull-always` option to docker configuration in *.dig file.
If the option configured, `docker pull` comannd will always  execute before `docker run` command.

Docker images that are using on digdag will updated sometimes so we want to use latest image always.

Thanks.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/DockerCommandExecutor.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/DockerCommandExecutor.java
@@ -70,7 +70,7 @@ public class DockerCommandExecutor
         }
         else {
             imageName = baseImageName;
-            if (dockerConfig.has("pull-always")) {
+            if (dockerConfig.get("pull-always", Boolean.class, false)) {
                 pullImage(imageName);
             }
         }
--- a/digdag-standards/src/main/java/io/digdag/standards/command/DockerCommandExecutor.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/DockerCommandExecutor.java
@@ -70,6 +70,9 @@ public class DockerCommandExecutor
         }
         else {
             imageName = baseImageName;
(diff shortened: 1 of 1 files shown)


digdag-standards/src/main/java/io/digdag/standards/command/DockerCommandExecutor.java
```diff
@@ -70,6 +70,9 @@ private Process startWithDocker(Path projectPath, TaskRequest request, ProcessBu
         }
         else {
             imageName = baseImageName;
+            if (dockerConfig.has("pull-always")) {
```
komamitsu: It seems this option should work as boolean so that users can set `false` to disable this feature. How about this one?
```
            if (dockerConfig.get("pull-always", Boolean.class, false)) {
                pullImage(imageName);
            }
```

alu: Thank you for review.
 
I think this options similar to `build` option.
If a user unnecessary pull image before run, the user just does not define the option.

But this is just my opinion so i would' like to change implementation along your advice if digdag team want.

alu: @komamitsu 

I fixed it (I was misunderstanding about `build` config so please ignore my previous comment.).

Then, i have one more suggestion.
i think better of `pull-always`'s default is true because in most cases it is convenient to be enabled.

Hou do you think it?

komamitsu: Thanks!

Regarding the default value, IMO `false` is safer and more expected behaviour. If the behaviour of Docker image changes implicitly with `true` by default, it may surprise users and cause a trouble. Also,  the backward compatibility is kept for now with `false` by default.

We can discuss that again and change the default behaviour on another ticket if needed in the future.

PR #604, opened 2017-07-26
fix all logos link

shio-phys:
the link of 'full list of files' goes 404.
http://docs.digdag.io/logo.html

so i change it to github's absolute url.


komamitsu:
LGTM 👍 
Thanks!

PR #607, opened 2017-07-31
Keep state params when task failed

akirakw:
This PR changes to keep state params when task failed.

In the current implementation, state params are keep stored when the task is successful, but is deleted when the task failed.

The state params is useful for tracking and troubleshooting of the failed task by CLI or WebUI, so I think the session should keep stored state params even if task failed.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -1826,7 +1826,7 @@ public class DatabaseSessionStoreManager
         long setStartedState(@Bind("id") long taskId, @Bind("oldState") short oldState, @Bind("newState") short newState);
 
         @SqlUpdate("update tasks" +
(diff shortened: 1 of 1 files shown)


frsyuki:
With this change, will state params be reset when the session is resumed?

frsyuki:
I think that overall idea is very good. I'm just wondering there're unexpected side effects.

akirakw:
I think this change is not affected to resuming session and related tasks for the following reasons:
- ResumingTask class has no state param, so when resuming tasks from task_archives, the state param always has initial (empty) value.
- Task resuming is executed only for the tasks finished with SUCCESS.

ref. https://github.com/treasure-data/digdag/blob/v0.9.13/digdag-core/src/main/java/io/digdag/core/workflow/TaskControl.java#L200

frsyuki:
I also double checked. Looks good!

frsyuki:
Added a comment as a following-up commit: https://github.com/treasure-data/digdag/commit/63f765147f91c170cab04f3ea934163ac99cc5b6

Issue #609, opened 2017-08-01
Not set the environment variable of the host to the Docker Container as it is


lulichn:
digdag version
```
$ digdag
2017-08-01 01:28:24 +0000: Digdag v0.9.13
```

Currently, when tasks run in a Docker container, environment variables passed from host to container.
```YAML
_export:
  docker:
    image: docker:latest

+step0:
  sh>: env
```

```
$ digdag run _docker.dig
2017-08-01 00:28:10 +0000 [WARN] (main) io.digdag.cli.Run: Using a new session time 2017-08-01T00:00:00+00:00.
2017-08-01 00:28:10 +0000 [INFO] (main) io.digdag.cli.Run: Using session /home/ec2-user/digdag/.digdag/status/20170801T000000+0000.
2017-08-01 00:28:10 +0000 [INFO] (main) io.digdag.core.workflow.WorkflowExecutor: Starting a new session project id=1 workflow name=_docker session_time=2017-08-01T00:00:00+00:00
2017-08-01 00:28:12 +0000 [INFO] (0016@[0:default]+_docker+step0) io.digdag.core.agent.OperatorManager: sh>: env
_command=env
_type=sh
AWS_AUTO_SCALING_HOME=/opt/aws/apitools/as
AWS_CLOUDWATCH_HOME=/opt/aws/apitools/mon
AWS_ELB_HOME=/opt/aws/apitools/elb
AWS_PATH=/opt/aws
DOCKER_CHANNEL=stable
DOCKER_VERSION=17.06.0-ce
docker={"image":"docker:latest"}
EC2_AMITOOL_HOME=/opt/aws/amitools/ec2
EC2_HOME=/opt/aws/apitools/ec2
HISTCONTROL=ignoredups
HISTSIZE=1000
HOME=/home/ec2-user
HOSTNAME=ip-xxx
JAVA_HOME=/usr/lib/jvm/java
LANG=en_US.UTF-8
last_executed_session_date_compact=
last_executed_session_date=
last_executed_session_local_time=
last_executed_session_time=
last_executed_session_tz_offset=+0000
last_executed_session_unixtime=
LESS_TERMCAP_mb=
LESS_TERMCAP_md=
LESS_TERMCAP_me=
LESS_TERMCAP_se=
LESS_TERMCAP_ue=
LESS_TERMCAP_us=
LESSOPEN=||/usr/bin/lesspipe.sh %s
LOGNAME=ec2-user
[...]

PR #611, opened 2017-08-02
[WIP] set environment variables

lulichn:
#609


frsyuki:
Passing environment variables using command line argument has security problems because sensitive information might leak through process list.

frsyuki:
I don't know why CommandExecutor interface needs a change.

yoyama:
If there is no action, I will close it.

PR #612, opened 2017-08-03
Make CI tests more robust

komamitsu:

PR #614, opened 2017-08-04
Fix race condition between cancel-request and subtask generation

frsyuki:
This fixes a possible race condition between cancel-request of an attempt and task-generating tasks (_error, loop>, etc.) as commented at https://github.com/treasure-data/digdag/compare/fix-cancel-request-race?expand=1#diff-d2d86ab5df03d4398a9ebc299b9183c5.
This PR assumes #612.

This also includes improvement of WorkflowExecutionTimeoutEnforcer:

* WorkflowExecutionTimeoutEnforcer should use smaller transactions so
  that it unlikely causes rollback. sendTimeoutNotification is a
  possible cause of rollback.

* Once sendTimeoutNotification completes successfully, cancel-request of
  the attempt should not rollback.

* `requestCancelAttempt` locks many records. WorkflowExecutionTimeoutEnforcer should unlock the records as soon as possible to avoid possible deadlock or performance degradation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ before_cache:
   - sudo chown `id -u`:`id -g` -R $HOME/.gradle $HOME/.m2  # make them cachable
 
 script:
-  - ci/run_test.sh
+  - travis_wait ci/run_test.sh
 
 after_success:
   - ci/travis_report_coverage.sh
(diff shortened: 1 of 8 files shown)


komamitsu:
LGTM! 👍

PR #617, opened 2017-08-09
Feature enhance if operator (_else_do)

yoyama:
I would like to implement  if task1 else task2 in if operator.
So I implemented '_else_do' task in if operator in this PR.


frsyuki:
ConfigEvalEngine (`${...}` parameter evaluator) has special handling for `_do` key at https://github.com/treasure-data/digdag/blob/b68b4dd2fd9a16b3751decf074c5bdc4b1d2c86b/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java#L150-L152. This is because parameters in `_do` key must not be evaluated. For example, in following case,

```
+loop_task:
  loop>: 3
  _do:
    echo>: ${i}
```

When `+loop_task` runs, variable `${i}` is not defined yet. Therefore, evaluation of `${i}` needs to be delayed until the inner task runs.

I think that the same thing applies to `if`:

```
+if_task:
  if>: ${typeof "lastTime" === "undefined"}
  _do:
    echo>: lastTime is not defined. Starting from the beginning.
  _else_do:
    echo>: Scanning from ${lastTime}
```

This workflow fails if `_else_do` is evaluated when `+if_task` runs due to `ReferenceError: "lastTime" is not defined` error. `_do` is the only key handled specially by ConfigEvalEngine.

yoyama:
Thank you four your comment.
I confirmed the issue and fixed it by adding "_else_do" as specially handled keys at ConfigEvalEngine.

digdag-core/src/test/java/io/digdag/core/workflow/IfOperatorFactory.java
```diff
@@ -0,0 +1,68 @@
+package io.digdag.core.workflow;
```
komamitsu: Why did you add this class instead of only modifying `io.digdag.standards.operator.IfOperatorFactory`?

yoyama: I am sorry for my late response.
The reason why I added the test class is that some other tests also add its test classes  in digdag-core/src/test/java/io/digdag/core/workflow such as EchoOperatorFactory, FailOperatorFactory and so on.
I adjusted to such manner.
I guess if I use the class in digdag-standards to test digdag-core, digdag-standarads is forced to depend on digdag-standards. And it may not be preferred.
But I don't know detail, so if I should use digdag-standards, I change the code.
Best regards,
  You Yamagata

muga:
@yoyama @komamitsu It looks like the rest of this PR may be the test improvement but, please let us know if need any help. Not rush at all but, I want to merge this feature:-)

yoyama:
Thank you for your comment.
I referred to other operators test classes and then changed the way to the tests.
If there are any issues please let me know.

PR #620, opened 2017-08-10
Change docker's pull-aways option to pull_always for consistency

frsyuki:
Operators use underscore as the separator.
This change also adds document for the option.
pull-always was added by #601 and it's not released yet. This change doesn't affect backward compatibility.

### DIFF
--- a/digdag-docs/src/architecture.md
+++ b/digdag-docs/src/architecture.md
@@ -53,6 +53,16 @@ You can use [Docker](https://www.docker.com/) to run tasks in a container.  If `
     +step1:
       py>: tasks.MyWorkflow.step1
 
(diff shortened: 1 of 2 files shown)


digdag-docs/src/architecture.md
```diff
@@ -53,6 +53,16 @@ You can use [Docker](https://www.docker.com/) to run tasks in a container.  If `
     +step1:
       py>: tasks.MyWorkflow.step1
 
+Digdag caches pulled image and reuses it. By default, currently, Digdag uses the cached image consistently even if there is an update. You can set `pull_always: true` option so that Digdag checks update of the image every time when a task starts.
```
komamitsu: > You can set `pull_always: true` option so that Digdag checks update of the image every time when a task starts

Strictly speaking, this doesn't say that Digdag pulls an update image after checking that if it exists. Adding that might be better.

frsyuki: ah, nice point \o/
I'll improve

PR #622, opened 2017-08-11
Fix order of generated tasks on UI

frsyuki:
Timeline view of tasks shows tasks in tree. It was not showing tasks in
correct order if some tasks are generated dynamically. This PR
fixes the problem by sorting the tasks before display.

Example:

```yaml
# a.dig
+loop1:
  for_each>:
    k: [1, 2]
  _do:
    echo>: ${k}

+loop2:
  for_each>:
    v: ['a', 'b']
  _do:
    echo>: ${v}
```

Before fix:

8a11aa5bbc8278ac2d22ae167a36abba

After fix:

cfa9d2143578568bad29bc32d5f6a2ed

### DIFF
--- a/digdag-ui/console.jsx
+++ b/digdag-ui/console.jsx
@@ -1302,8 +1302,8 @@ const JobLink = ({storeParams, stateParams}:{storeParams: Object, stateParams: O
 }
 
 function sortTasksForTreeView (tasks: Array<Task>): Array<Task> {
-  let collectChildrenRecursively = function (result: Array<Task>, taskGroups: Map<string, Array<Task>>, parentTask: Task) {
-    let group: ?Array<Task> = taskGroups.get(parentTask.id)
+  function collectChildrenRecursively (result: Array<Task>, taskGroups: Map<string, Array<Task>>, parentTask: Task) {
+    const group: ?Array<Task> = taskGroups.get(parentTask.id)
     if (group != null) {
       taskGroups.delete(parentTask.id)
       group.forEach(t => {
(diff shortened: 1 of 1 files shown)


digdag-ui/console.jsx
```diff
@@ -1301,6 +1301,52 @@ const JobLink = ({storeParams, stateParams}:{storeParams: Object, stateParams: O
   return <a href={link} target='_blank'>{jobId}</a>
 }
 
+function sortTasksForTreeView (tasks: Array<Task>): Array<Task> {
+  let collectChildrenRecursively = function (result: Array<Task>, taskGroups: Map<string, Array<Task>>, parentTask: Task) {
```
dzannotti: this can be a `const`

digdag-ui/console.jsx
```diff
@@ -1301,6 +1301,52 @@ const JobLink = ({storeParams, stateParams}:{storeParams: Object, stateParams: O
...
+    let group: ?Array<Task> = taskGroups.get(parentTask.id)
```
dzannotti: this one too

digdag-ui/console.jsx
```diff
@@ -1326,7 +1372,7 @@ class TaskListView extends React.Component {
           </thead>
           <tbody>
             {
-              Array.from(this.props.tasks.values()).map(task =>
+              sortTasksForTreeView(Array.from(this.props.tasks.values())).map(task =>
```
dzannotti: I'm not 100% sure but, I think `this.props.tasks.values()` already returns an array, why casting it to array again?

digdag-ui/console.jsx
frsyuki: :+1: Thank you!

digdag-ui/console.jsx
frsyuki: According to MDN, [Map.prototype.values()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/values) returns Iterator. And thus flow shows this error without the cast:

```
1375:               sortTasksForTreeView(this.props.tasks.values()).map(task =>
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^ $Iterator. This type is incompatible with the expected param type of
1304: function sortTasksForTreeView (tasks: Array<Task>): Array<Task> {
                                            ^^^^^^^^^^^ array type
```

PR #625, opened 2017-08-11
Faster startup and less memory consumption in integration tests

frsyuki:
Adding -XX:TieredStopAtLevel=1 has similar effect with obsoleted -client
option. -Xverify:none simply skips validation of jar files and makes
startup of JVM faster.

This might mitigate the problem described at #624.


frsyuki:
1 hour 07 minutes became 56 minutes. 17% improved
https://circleci.com/gh/treasure-data/digdag

komamitsu:
Interesting. Let's try it and see what happens.

Issue #628, opened 2017-08-23
`download_file` in `pg>` operator raise an exception if download data contains null.


hiroyuki-sato:
Hello maintainers.

`download_file` in `pg>` operator raise an exception if download data contains null.

* Digdag: 0.9.1.5
* OS: MacOSX 10.12.6
* PostgreSQL: 9.6.2

### prepare data

```sql
drop table if exists digdag_test;
create table digdag_test (
  id int not null,
  name varchar(255) null,
  birthday varchar(255) null
);

insert into digdag_test(id) values(1) ;
insert into digdag_test(id) values(2) ;
```

```
select * from digdag_test;
 id | name | birthday
----+------+----------
  1 |      |
  2 |      |
(2 rows)
```

### test.dig & dump.sql

```yaml
_export:
  pg:
    host: localhost
    user: user
    database: digdag_support

+task1:
  pg>: dump.sql
  download_file: out.txt
```

dump.sql

```sql
select * from digdag_test;
```

### digdag run -a test result

```
2017-08-23 19:52:08 +0900: Digdag v0.9.15
2017-08-23 19:52:09 +0900 [WARN] (main): Reusing the last session time 2017-08-23T00:00:00+00:00.
2017-08-23 19:52:09 +0900 [INFO] (main): Using session/digdag-support/test_20170823/.digdag/status/20170823T000000+0000.
2017-08-23 19:52:09 +0900 [INFO] (main): Starting a new session project id=1 workflow name=test session_time=2017-08-23T00:00:00+00:00
2017-08-23 19:52:10 +0900 [INFO] (0017@[0:default]+test+task1): pg>: dump.sql
2017-08-23 19:52:10 +0900 [ERROR] (0017@[0:default]+test+task1): Task failed with unexpected error: at index 1
java.lang.NullPointerException: at index 1
	at com.google.common.collect.ObjectArrays.checkElementNotNull(ObjectArrays.java:239)
	at com.google.common.collect.ObjectArrays.checkElementsNotNull(ObjectArrays.java:230)
	at com.google.common.collect.ObjectArrays.checkElementsNotNull(ObjectArrays.java:225)
	at com.google.common.collect.ImmutableList.copyOf(ImmutableList.java:273)
	at io.digdag.standards.operator.jdbc.AbstractJdbcResultSet.getObjects(AbstractJdbcResultSet.java:60)
[...]

PR #630, opened 2017-08-23
Support NULL-able values on `pg` operator's `download_file` option

komamitsu:
`io.digdag.standards.operator.jdbc.AbstractJdbcResultSet#getObjects` wrapped result values with `com.google.common.collect.ImmutableList`, but it doesn't accept null values. So this change uses just `ArrayList` instead of `ImmutableList`.

This pull request is to fix https://github.com/treasure-data/digdag/issues/628.


komamitsu:
@frsyuki Can you take a look at this change?

komamitsu:
Thanks!

PR #631, opened 2017-08-25
Show "Canceled" instead of "Failure" on attempt status view

frsyuki:
Status of a session shows "Failure" if the session was canceled. But because cancelation usually means intentional cancel, it seems better to show different color from failure which is usually unintentional.

c04e9c2be21deeb9fe699200857dfa18


komamitsu:
@frsyuki LGTM

PR #633, opened 2017-08-25
Add a test that confirms DigdagClient sends `Accept-Encoding: gzip`

komamitsu:
For https://github.com/treasure-data/digdag/pull/590


frsyuki:
Nice!!!

PR #634, opened 2017-08-28
Refactoring of use of Throwables.propagate.

frsyuki:
As guava's document says, adding throw to Throwables.propagate can let
compilers know that following statements after the call are unreachable.
Behavior doesn't change because Throwable.propagate never returns.


komamitsu:
👍

PR #635, opened 2017-08-28
TransactionManager.getOrReuse for submodules that may or may not nested

frsyuki:
Some modules such as NotificationSender may or may not in transaction
depending on its context. Accessing database from those modules have an
issue where it causes deadlock due to lack of connections in connection
pool, or causes "Nested transaction is not allowed" error.

TransactionManager.getOrReuse lets those modules share the same
transaction context with the external caller such as
WorkflowExecutionTimeoutEnforcer, while it can start a new transaction
if the external caller is not in transaction such as
NotifyOperatorFactory.

In fact, WorkflowExecutionTimeoutEnforcer should not call
NotificationSender synchronously. Instead, it should create a
notification task and send notification asynchronously. But here it's
postponed.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/ThreadLocalTransactionManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/ThreadLocalTransactionManager.java
@@ -273,6 +273,47 @@ public class ThreadLocalTransactionManager
     }
 
     @Override
+    public <T> T beginOrReuse(SupplierInTransaction<T, RuntimeException, RuntimeException, RuntimeException> func)
+    {
+        return beginOrReuse(func, RuntimeException.class, RuntimeException.class, RuntimeException.class);
+    }
+
+    @Override
+    public <T, E1 extends Exception> T beginOrReuse(SupplierInTransaction<T, E1, RuntimeException, RuntimeException> func, Class<E1> e1)
+            throws E1
+    {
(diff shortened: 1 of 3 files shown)


komamitsu:
@frsyuki 
It sounds dangerous to me to share a transaction between `MultiThreadAgent#transactionManager` and other modules such as NotificationSender if we can't guarantee those modules use only read-only actions in the shared transaction. If those modules issue COMMIT in the shared transaction, it can cause unexpected wrong task states.

I think the root cause is that those other modules try to use `MultiThreadAgent#transactionManager`. Sharing a transaction may reduce the number of connections. But it's safer to pass another `ThreadLocalTransactionManager` to a custom module's class, in this case it's `TdOauthTokenManager`, I think. What do you think?

frsyuki:
@komamitsu 

> it's safer to pass another ThreadLocalTransactionManager to a custom module's class

I see, I didn't have that idea! It makes more sense. I think that it's even better if custom modules have separated connection pool (== another `DataSource` instance) as well as separated transaction manager.

Then...the customer module injects DatabaseConfig instead of TransactionManager, overrides some parameters as necessary (e.g. decrease max connections), and creates a new instance of DataSourceProvider and ThreadLocalTransactionManager?

frsyuki:
#713 takes over this PR.

PR #642, opened 2017-09-05
Fix wrong "_background|_after is not used" message

### LINKED ISSUE #641: [Q] Parallel background
Following is my dig:

```
+prepare:
  +data1:
    echo>: data1

  # +data1 and +data2 run in parallel.
  +data2:
    _background: true
    echo>: data2

  # +data3 runs after +data1 and +data2.
  +data3:
    echo>: data3
```

and I get following error message, looks like the dig did running parall.
```
[ERROR] (0019@[0:default]+fail+prepare+data2): Parameter '_background' is not used at task +fail+prepare+data2.
```
I am confused it's a false alarm or my dig is wrong. Thanks

frsyuki:
Setting _background or _after option is working but it was showing
`Parameter '_background' is not used` or `Parameter '_after' is not used`
warning messages wrongly.

Fixes #641.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowCompiler.java
+++ b/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowCompiler.java
@@ -236,7 +236,7 @@ public class WorkflowCompiler
                     Map<String, TaskBuilder> names = new HashMap<>();
                     for (TaskBuilder subtask : subtasks) {
                         if (subtask.getConfig().get("_background", boolean.class, false)) {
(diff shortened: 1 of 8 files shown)


komamitsu:
LGTM.

BTW, I'm wondering why we don't make `CheckedConfig#get` mark a key with `used` like this:
```

    CheckedConfig(Config config, UsedKeysSet usedKeys)
    {
        super(config);
        this.usedKeys = usedKeys;
    }

    @Override
    public <E> E get(String key, Class<E> type, E defaultValue)
    {
        this.usedKeys.setAllUsed(true);
        return super.get(key, type, defaultValue);
    }
```
Doesn't this work?

PR #650, opened 2017-09-24
Support 'store_result' option in the JDBC operator family

komamitsu:


komamitsu:
@frsyuki Can you review this pull request?

digdag-tests/src/test/java/acceptance/PgIT.java
```diff
@@ -199,6 +199,73 @@ public void selectAndDownloadWithNullValues()
...
+        setupSourceTable();
+
+        CommandStatus status = TestUtils.main(
+                "run", "-o", root().toString(),
+                "--project", root().toString(),
+                "-p", "pg_database=" + tempDatabase,
+                "-p", "outfile=out",
+                "-X", "config.pg.max_store_result_rows=2",
+                "pg.dig");
+        assertThat(status.code(), is(1));
```
frsyuki: It's better to check error message as well as exit code to confirm that tests don't wrongly success when pg operator failed to parse `config.pg.max_store_result_rows` option.

digdag-tests/src/test/java/acceptance/PgIT.java
```diff
@@ -199,6 +199,73 @@ public void selectAndDownloadWithNullValues()
...
+
+        setupSourceTable();
+
+        CommandStatus status = TestUtils.main(
+                "run", "-o", root().toString(),
+                "--project", root().toString(),
+                "-p", "pg_database=" + tempDatabase,
+                "-p", "outfile=out",
+                "pg.dig");
+        assertThat(status.code(), is(0));
```
frsyuki: Adding first message argument to assertThat is very helpful for debugging:

```
assertThat(status.errUtf8(), status.code(), is(0));
```

This shows `status.errrUtf8()` when the assertion fails. It informs reason of the failure which is helpful especially on CI.

digdag-tests/src/test/java/acceptance/PgIT.java
komamitsu: 👍 
Depending on an error message is a bit naive, but it's better than never. Maybe we'd better add error codes more reliable in the future.

komamitsu:
> I think it's better to consider consistency with td> operator.

Sounds reasonable.
Addressed your feedbacks.

komamitsu:
@frsyuki Updated. Can you take a look when you get a chance?

frsyuki:
Nice 👍  LGTM

komamitsu:
Thanks!

frsyuki:
Following-up: #663.

PR #652, opened 2017-09-27
Fix #651. use encode file_name parameter except '^'

### LINKED ISSUE #651: digdag-ui can't show tasklog when giving hash parameter to for_each>:
When giving hash parameter:
```yaml
+step1:
  echo>: start
+step2:
  +repeat:
    for_each>:
      param:
        - key: p1
        - key: p2
        - key: p3
    _do:
      +substep:
        echo>: ${param.key}
+step3:
  echo>: end
```

Tasks success but digdag-ui fails showing tasklog:

image

hiroyuki-sato:
@frsyuki @komamitsu 

Please take a look when you can.

Fix #651. 

What do you think this PR? especially `fileName.replaceFirst("%5E", "^")` part.
(I think It's better to create separate class/method)

* digdag-ui client request `/api/logs/#{attempt_id}/files/#{file_name}` with following `file_name` parameter.

`+test+step2+repeat%5Esub+for-0=param=2=%7B%22key%22%3A%22+substep@59cb79c5163b8740.19948@hostname.local.log.gz`

* Current implementation decodes % encoding like `+test+step2+repeat^sub+for-0=param=2={"key":"`.
So I Add `@Encode` annotation. 

* But `%5E` character needs decode as `^`. because file name begin with `+test+step2+repeat^sub+for-`

digdag

### DIFF
--- a/digdag-server/src/main/java/io/digdag/server/rs/LogResource.java
+++ b/digdag-server/src/main/java/io/digdag/server/rs/LogResource.java
@@ -5,6 +5,7 @@ import java.time.Instant;
 import java.io.InputStream;
 import java.io.IOException;
 import javax.ws.rs.Consumes;
+import javax.ws.rs.Encoded;
 import javax.ws.rs.Produces;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
@@ -119,6 +120,7 @@ public class LogResource
     @GET
     @Produces("application/gzip")
     @Path("/api/logs/{attempt_id}/files/{file_name}")
+    @Encoded
     public byte[] getFile(
             @PathParam("attempt_id") long attemptId,
             @PathParam("file_name") String fileName)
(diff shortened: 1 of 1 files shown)


komamitsu:
Thanks! It seems nice to me.

But I'm using S3StorageFactory (`log-server.type = s3`) and it works well with the workflow on https://github.com/treasure-data/digdag/issues/651.

So what we need to modify might be `io.digdag.core.log.LocalFileLogServerFactory.LocalFileLogServer` or something.

@frsyuki What do you think?

frsyuki:
According to https://github.com/treasure-data/digdag/issues/651#issuecomment-332420646 (Thank you very much for this investigation), the root cause is that someone decodes file name. Re-encoding the file name at the server seems wrong solution.
Shouldn't we fix the root cause instead? Someone (probably UI) decoded file names while it shouldn't.

frsyuki:
Isn't this?: #661

hiroyuki-sato:
This issue fixed by #661.

https://github.com/treasure-data/digdag/pull/661#issuecomment-334624676

Issue #654, opened 2017-09-28
Proposal for for_range operator


frsyuki:
We have a ETL workflow that needs to process approximately 300 million rows at a time. It is too big to process using one statement. Therefore, we want to divide it into multiple tasks using workflow.

Our current approach is this:

```yaml
# Fetch statistics of the source table
+count:
  td>: {data: select min(time) as min_time, max(time) as max_time, count(*) as count from src}
  store_last_results: true

# Divide it into 1 million rows
+loop:
  loop>: ${Math.ceil(td.last_results.count / 1000000.0)}
  _do:
    td>: etl_with_time_range.sql
    time_width: ${Math.max((max_time - min_time) / Math.ceil(td.last_results.count / 1000000.0), 1)}
    time_cond: |
      ${td.last_results.min_time + i * parseInt(time_width)} <= time
      AND time <${td.last_results.min_time + (i + 1) * parseInt(time_width)}
```

There are several problems with this approach.

* Workflow becomes complicated. It's hard to read.
* Calculation of ceil and max are tricky and easily lead bugs.
* If count is big, this loop creates too many tasks. It should add limitation on number of loop but adding such limitation makes the workflow even more complicated.

Proposal here is adding another operator to make it easy.
It can make the workflow much simpler:

```yaml
# Fetch statistics of the source table
+count:
  td>: {data: select min(time) as min_time, max(time) as max_time, count(*) as count from src}
  store_last_results: true

# Divide it into 1 million rows
+loop:
  # for the range of [min_time, max_time)
  for_range>:
    from: ${td.last_results.min_time}
    until: ${td.last_results.max_time + 1}
    # slice the range into this many chunks:
    count: ${Math.min(td.last_results.count / 1000000, 200)}  # at most 200 tasks
  # it loops for each chunk with `range.from` and `range.until` parameters set
  _do:
    td>: etl_with_time_range.sql
    time_cond: |
      ${range.from} <= time AND ${range.until} < time
```

frsyuki:
Sample implementation is like this:

```java
        public TaskResult run()
        {
            Config params = request.getConfig();

            Config doConfig = request.getConfig().getNested("_do");
            boolean parallel = params.get("_parallel", boolean.class, false);

            Config rangeConfig = params.parseNested("_command");

            long from = rangeConfig.get("from", long.class);
            long until = rangeConfig.get("until", long.class);
            Optional<Long> stepConfig = rangeConfig.getOptional("step", Long.class);
[...]

PR #655, opened 2017-09-29
Adding for_range operator

frsyuki:
See #654 for discussion.


digdag-standards/src/main/java/io/digdag/standards/operator/ForRangeOperatorFactory.java
```diff
@@ -0,0 +1,113 @@
...
+
+        private static String buildTaskName(long rangeFrom, long rangeUntil)
+        {
+            return String.format("+range-from=%d&until=%d", rangeFrom, rangeUntil);
+        }
+
+        private static void enforceTaskCountLimit(int size)
+        {
+            if (size > Limits.maxWorkflowTasks()) {
+                throw new ConfigException("Too many for_range subtasks. Limit: " + Limits.maxWorkflowTasks());
```
komamitsu: Showing `size` here would help users to understand what's wrong since it's likely users don't know an actual number of tasks with `step`.

digdag-standards/src/main/java/io/digdag/standards/operator/ForRangeOperatorFactory.java
```diff
@@ -0,0 +1,113 @@
...
+            Config params = request.getConfig();
+
+            Config doConfig = request.getConfig().getNested("_do");
+            boolean parallel = params.get("_parallel", boolean.class, false);
+
+            Config rangeConfig = params.parseNested("_command");
+            long from = rangeConfig.get("from", long.class);
+            long until = rangeConfig.get("until", long.class);
+
+            Optional<Long> stepConfig = request.getConfig().getOptional("step", Long.class);
```
komamitsu: How about getting these parameters from `rangeConfig`? I think getting them from the same config would be more consistent.

digdag-standards/src/main/java/io/digdag/standards/operator/ForRangeOperatorFactory.java
frsyuki: That's very good point.

digdag-standards/src/main/java/io/digdag/standards/operator/ForRangeOperatorFactory.java
frsyuki: hmm...two  things: I want to avoid nested objects as much as possible because I'm seeing many people confused with YAML's indents. Another thing is that if we have `width` in nested config, it's misleading because people may think it's the width of the total range (not width of a slice).

frsyuki: To avoid all syntax misunderstanding, this is the best:

```
for_range>: [min, max]  # inclusive min max
step:
slices:  # step or slices
```

frsyuki: But with `step` instead of `width`, nesting seems OK.

frsyuki:
I changed the syntax to be like this:


```yaml
+loop:
  for_range>:
    from: 0
    to: 100
    step: 10
  _parallel: true
  _do:
    +nesting:
      echo>: "doing something for range [${range.from}, ${range.to})"
```

digdag-standards/src/main/java/io/digdag/standards/operator/ForRangeOperatorFactory.java
```diff
@@ -0,0 +1,113 @@
...
+            Optional<Long> countConfig = request.getConfig().getOptional("count", Long.class);
```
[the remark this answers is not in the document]
frsyuki: I now think `count` is confusing. I picked `slices`.

frsyuki:
I also added `range.index` variable so that tasks can use it as an unique identifier of the task.
This is useful for this use case: for_range operator creates temporary tables (tmp_0, tmp_1, tmp_2). A following task concatenates them (tmp_0 UNION ALL tmp_1 UNION ALL tmp_2)

PR #659, opened 2017-10-02
Fix issue #658

### LINKED ISSUE #658: digdag-ui can't show task log when use of HTTP Basic Authentication
Hi. digdag-ui can't show task log when connected through a proxy server.   
Using Nginx and restricting access with HTTP Basic Authentication.   
In that situation, we always get a `401  (Unauthorized)` Error.

See below. (Left: Normal, Right: Through Proxy)

szyn:
To fix #658 issue.
Include credentials when fetchArrayBuffer.

It looks like below. (Left: Normal, Right: Through Proxy)

### DIFF
--- a/digdag-ui/model.js
+++ b/digdag-ui/model.js
@@ -485,7 +485,9 @@ export class Model {
   }
 
   fetchArrayBuffer (url: string) {
-    return fetch(url).then(response => {
+    return fetch(url, {
+      credentials: 'include'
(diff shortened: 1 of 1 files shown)


frsyuki:
Very good catch!

frsyuki:
Merged. Thank you.

szyn:
Thank you, too!

PR #662, opened 2017-10-06
Added path-style-access opiton in server mode configures

hinata:
This option require in using aws s3 like object storage (e.g. ceph/radosgw, minio and riak-cs).

### DIFF
--- a/digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3StorageFactory.java
+++ b/digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3StorageFactory.java
@@ -33,9 +33,9 @@ public class S3StorageFactory
 
         if (config.has("path-style-access")) {
             client.setS3ClientOptions(
-              new S3ClientOptions().withPathStyleAccess(
-                config.get("path-style-access", Boolean.class)
-              ));
+              S3ClientOptions.builder().setPathStyleAccess(
+                config.get("path-style-access", Boolean.class, false)
+              ).build());
         }
 
         String bucket = config.get("bucket", String.class);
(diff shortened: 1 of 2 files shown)


frsyuki:
I think we can merge this. Any concerns, @komamitsu?

digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3StorageFactory.java
```diff
@@ -30,6 +31,13 @@ public Storage newStorage(Config config)
             client.setEndpoint(config.get("endpoint", String.class));
         }
 
+        if (config.has("path-style-access")) {
```
frsyuki: this should be `config.get("path-style-access", boolean.class, false)` otherwise `archive.s3.path-style-access = false` enables path-style-access unexpectedly.

digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3StorageFactory.java
```diff
@@ -30,6 +31,13 @@ public Storage newStorage(Config config)
...
+            client.setS3ClientOptions(
+              new S3ClientOptions().withPathStyleAccess(
+                config.get("path-style-access", Boolean.class)
```
frsyuki: this should be config.get("path-style-access", boolean.class, false)

frsyuki: `withPathStyleAccess` is a deprecated method. According to documents, here should use S3ClientOptions.Builder.

komamitsu:
LGTM!

frsyuki:
:+1:

frsyuki:
@hinata Thank you for your contribution!

hinata:
@frsyuki Thank you, too.

PR #663, opened 2017-10-06
Making "store_last_results: true" consistent between td> and pg>

frsyuki:
`td>` and `td_run>` operators support `store_last_results: boolean`
option. It stores the first row to `${td.last_results}` parameter. But
`pg>` and `redshift>` operators store all rows and throw an exception
if number of rows exceeds limit.

This change makes it possible to achieve the consistent behavior to be
less surprising. `pg>` and `redshift>` operators support `all`, `first`
and `true` options.  `all` stores all rows and causes exception if rows
exceeds limit. `first` sets the first row. `true` is an alias of `first`
to be consistent with `td>` operator.

`td>` operator should also support `all` and `first` but not implemented
yet.

### DIFF
--- a/digdag-docs/src/operators/pg.md
+++ b/digdag-docs/src/operators/pg.md
@@ -28,7 +28,7 @@ _export:
 ```
 +select_members:
   pg>: select_members.sql
-  store_last_results: true
+  store_last_results: first
 
 +send_email:
   for_each>:
@@ -90,14 +90,22 @@ _export:
   download_file: output.csv
   ```
 
-* **store_last_results**: BOOLEAN
+* **store_last_results**: first | all
 
-  Whether to store the query result. *Default:* `false`.
+  Whether to store the query results to ``pg.last_results`` parameter. *Default:* `false`.
+
+  Setting ``first`` stores the first row to the parameter as an object (e.g. ``${pg.last_results.count}``).
+
+  Setting ``all`` stores all rows to the parameter as an array of objects (e.g. ``${pg.last_results[0].name}``). If number of rows exceeds limit, task fails.
 
   Examples:
 
   ```
-  store_last_results: true
+  store_last_results: first
+  ```
+
+  ```
+  store_last_results: all
   ```
 
 * **database**: NAME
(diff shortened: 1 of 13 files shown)


frsyuki:
@komamitsu I understand your concern but what are possible other options...?

komamitsu:
Just `store_result`or `store_last_result` (only one object), and `store_results` or `store_result_set` (array of objects) seem clear to me. But `td` operator already uses `store_last_results` to store only one object, so it's not easy to go with the above naming keeping backward compatibility.

One reasonable idea is to use other names for the new JDBC operators family like `save_last_result` (object) and `save_result_set` (array of object) and add the new ones to `td` operator as well. And then make the current `store_last_results` (object) deprecated. But it takes time and we need to maintain old `store_last_result` and new `save_xxxx` options for a while. So it might be not bad to go with the current way `store_last_results: true|first|all`.

frsyuki:
@komamitsu right...I guessed that having both of `save_last_result` and `save_result_set` also confuses users because option name is different even though they control the same thing -- saving result set in a variable (so setting both of save_last_result and save_result_set should be an error).

But on the other hand, `save_result_set: first | all` is also confusing because using `${pg.result_set}` both for first row and all rows causes the same issue with `store_last_results`. However, still, using different variable names (such as `${pg.result_set_first.count}` for `save_result_set: first` and `${pg.result_set_all[0].name}` for `save_result_set: all`) seems also confusing because variable name doesn't match with the option name...

frsyuki:
I'm thinking that `store_last_results: true|first|all` is not too bad idea even without consideration of backward compatibility.

frsyuki:
Another possible option is that option name is ``save_result_set: first | all`` and variable name is `${pg.result_set.first}` for `first` or `${pg.result_set.all}` for `all`.
Well, but still current behavior doesn't seem worse than this.

komamitsu:
> `${pg.result_set.first}` for `first` or `${pg.result_set.all}` for `all`

I prefer this one because the variables `pg.result_set.first` and `pg.result_set.all` have each single type (object / array). In other word, users can know the type of a variable based on its name.

My root concern is that `last_results` can be either type object or array and I think this may confuse users since users can't know the type of a variable based on its name.

I think we could add this feature as an alternative of `last_results` later.

frsyuki:
> I think we could add this feature as an alternative of last_results later.

Sounds like a reasonable plan.

PR #664, opened 2017-10-09
UI: Add Kill Attempt Button

szyn:
It might be a good idea to kill attempt from digdag-ui.   
So, I added `kill` button to attempt view.

### DIFF
--- a/digdag-ui/console.jsx
+++ b/digdag-ui/console.jsx
@@ -355,6 +355,13 @@ function attemptCanResume (attempt) {
   return attempt.done && !attempt.success
 }
 
+function attemptCanKill (attempt) {
+  if (!attempt) {
+    return false
+  }
+  return !attempt.done && !attempt.success
+}
+
 const SessionStatusView = ({session}:{session: Session}) => {
   const attempt = session.lastAttempt
   return attempt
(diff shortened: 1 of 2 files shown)


frsyuki:
definitely it is a good idea!

digdag-ui/console.jsx
```diff
@@ -355,6 +355,13 @@ function attemptCanResume (attempt) {
   return attempt.done && !attempt.success
 }
 
+function attemptCanKill (attempt) {
+  if (!attempt) {
+    return false
+  }
+  return !attempt.done && !attempt.success
```
frsyuki: An attempt can be killed only when attempt is not done. `&& !attempt.success` is unnecessary. Instead, we need to check `attempt.cancelRequested` flag because having `cancelRequested` flag means that it's already kill-requested.

frsyuki:
Follow-up commit: 7f5f8366c467eccd5d14016cbe0122f7ef386435

digdag-ui/console.jsx
szyn: I have understood the indicated matter described in your comment.  
Thank you very much for the review and your follow-up.

PR #640, opened 2017-09-05
Add contributors to release notes

komamitsu:

PR #665, opened 2017-10-10
Create release note template with contributors list automatically

frsyuki:
With this change, ``./gradlew setVersion -Pto=...`` command
creates release note file with a template which comes with version,
release date, and contributors list.

This automates #640.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -438,8 +438,21 @@ task setVersion << {
         releases.write(releases.getText().replaceAll("# add new version here", "# add new version here\n    releases/release-${to}"))
 
         // release-<versionr>.rst
-        file("digdag-docs/src/releases/release-${to}.rst").append("")
-        "git add digdag-docs/src/releases/release-${to}.rst".execute().waitFor()
+        String notePath = "digdag-docs/src/releases/release-${to}.rst"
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -438,8 +438,21 @@ task setVersion << {
...
+                "\n" +
+                "...\n" +
+                "\n" +
+                "Release Date\n" +
+                "------------\n" +
+                "${new Date().format('yyyy-MM-dd')}\n" +
+                "\n" +
+                "Contributors\n" +
+                "------------\n")
+        ["bash", "-c", "git shortlog -s \$(git tag --sort=-creatordate | head -n 1)..HEAD | awk '{print \"* \" \$2}' >> ${notePath}"].execute().waitFor()
```
komamitsu: > `$2`

If one of rows from `git shortlog` is `   1234   Foo Bar`, this only captures `Foo`?

komamitsu: `awk '{$1 = ""; print "* " $0}'`
Like this?

frsyuki: Good catch...! Sadayuki Furuhashi was becoming Sadayuki...

PR #666, opened 2017-10-11
Expose embulk stderr

civitaspo:
Hi maintainers,

Currently, embulk operator does not expose embulk stderr, so it's difficult to know why embulk execution is failed. This p-r is to expose the stderr.


frsyuki:
ok, EmbulkOperatorFactory is very old code...
We should rather use CommandLogger to put log to task log as well as logger so that the message comes with workflow/task name and server-mode users can see the logs.
Created #667.

civitaspo:
Thank you for following up! So, let this close.

PR #668, opened 2017-10-11
Fix CORS issue with direct log file download

frsyuki:
PR #659 introduced a problem where browser rejects a request to get a
log file from a S3's pre-signed URL when S3 sends
`Access-Control-Allow-Origin: '*'` header.  This happens only when the
server generates a direct-download URL for log files.

An example message from Chrome is:
```
The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.
```

Credentials should not be included in the request to accept
`Access-Control-Allow-Origin: '*'` header when the server generates
pre-signed URLs for log files.

### DIFF
--- a/digdag-ui/model.js
+++ b/digdag-ui/model.js
@@ -298,10 +298,11 @@ export class Model {
   }
 
   fetchLogFile (attemptId: string, file: LogFileHandle) {
-    if (!file.direct) {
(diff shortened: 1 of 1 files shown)


szyn:
I’m so sorry for my overlooking and appreciate that you fixed quickly.

PR #673, opened 2017-10-16
Fix argument of parallel parameter on redshift_unload document

hiroyukim:
Correct the parameter argument of parallel in the redshift_unload document. I checked the code and the test, and it was true that it should be described as ON.

## Code

https://github.com/treasure-data/digdag/blob/master/digdag-standards/src/main/java/io/digdag/standards/operator/redshift/RedshiftUnloadOperatorFactory.java#L144

## Test

https://github.com/treasure-data/digdag/blob/master/digdag-standards/src/test/java/io/digdag/standards/operator/redshift/RedshiftUnloadOperatorFactoryTest.java#L150

### DIFF
--- a/digdag-docs/src/operators/redshift_unload.md
+++ b/digdag-docs/src/operators/redshift_unload.md
@@ -25,7 +25,7 @@
         null_as: nULl
         escape: false
         addquotes: true
-        parallel: true
+        parallel: ON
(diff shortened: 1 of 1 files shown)


frsyuki:
Good catch. Thank you 👍

PR #682, opened 2017-11-10
Fix TimeZone format for td_partial_delete

kamikaseda:
The TimeZone format is wrong.


komamitsu:
Indeed, `io.digdag.client.api.LocalTimeOrInstant#fromString` uses `java.time.format.DateTimeFormatter#ISO_OFFSET_DATE_TIME`. Thanks!

PR #690, opened 2017-11-27
[DOC] Add missing executor parameters.

hiroyuki-sato:
Add missing `executor.task_ttl ` and `executor.attempt_ttl ` documentation.

Ref: https://twitter.com/toyama0919/status/933979861382610949

### DIFF
--- a/digdag-docs/src/command_reference.rst
+++ b/digdag-docs/src/command_reference.rst
@@ -350,7 +350,7 @@ In the config file, following parameters are available
 * log-server.s3.credentials.secret-access-key (string. default: instance profile)
 * log-server.s3.path-style-access (boolean. default: false)
 * digdag.secret-encryption-key = (base64 encoded 128-bit AES encryption key)
-* executor.task_ttl (string. default: 1d. A task is killed if a single task is running longer than this period.)
+* executor.task_ttl (string. default: 1d. A task is killed if it is running longer than this period.)
 * executor.attempt_ttl (string. default: 7d. An attempt is killed if it is running longer than this period.)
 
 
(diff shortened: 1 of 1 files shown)


digdag-docs/src/command_reference.rst
```diff
@@ -350,6 +350,8 @@ In the config file, following parameters are available
 * log-server.s3.credentials.secret-access-key (string. default: instance profile)
 * log-server.s3.path-style-access (boolean. default: false)
 * digdag.secret-encryption-key = (base64 encoded 128-bit AES encryption key)
+* executor.task_ttl (A task kills if a single task is running longer than this period. string. default: 1d)
```
komamitsu: Could you change `kills` to `is killed`? 

Also, it might be better to put a type first considering other items above. How about this?
```
(string. default: 1d. A task is killed if it runs more than this period)
```

hiroyuki-sato: @komamitsu Thank you for the reply. Fixed.

hiroyuki-sato:
Thank you for the reply. Please take a look again when you have a time.

digdag-docs/src/command_reference.rst
```diff
@@ -350,6 +350,8 @@ In the config file, following parameters are available
 * log-server.s3.credentials.secret-access-key (string. default: instance profile)
 * log-server.s3.path-style-access (boolean. default: false)
 * digdag.secret-encryption-key = (base64 encoded 128-bit AES encryption key)
+* executor.task_ttl (string. default: 1d. A task is killed if a single task is running longer than this period.)
```
komamitsu: It might be better to use the same expression (`s/a single task/it/`) with the below one?

hiroyuki-sato: Thank you for the reply. fixed it.

hiroyuki-sato:
Thank you for the reply. Please take a look again when you have a time.

komamitsu:
Thanks!

Issue #687, opened 2017-11-22
setting secrets can't contain unescaped '$' on cli


69guitar1015:
Hi
I found that I set secrets like
```sh
digdag secrets --local --set password=i_want_$50000
# or 
digdag secrets --local --set password # enter
password: i_want_$50000
```

and use it in dig file then these error occurs
```
java.lang.IllegalArgumentException: Illegal group reference
	at java.util.regex.Matcher.appendReplacement(Matcher.java:857)
	at io.digdag.util.UserSecretTemplate.replaceAll(UserSecretTemplate.java:64)
	at io.digdag.util.UserSecretTemplate.format(UserSecretTemplate.java:55)
	at io.digdag.core.agent.GrantedPrivilegedVariables.lambda$buildAccessor$1(GrantedPrivilegedVariables.java:62)
	at io.digdag.core.agent.GrantedPrivilegedVariables.get(GrantedPrivilegedVariables.java:81)
	at io.digdag.standards.operator.ShOperatorFactory.collectEnvironmentVariables(ShOperatorFactory.java:156)
	at io.digdag.standards.operator.ShOperatorFactory$ShOperator.runTask(ShOperatorFactory.java:111)
	at io.digdag.util.BaseOperator.run(BaseOperator.java:35)
	at io.digdag.core.agent.OperatorManager.callExecutor(OperatorManager.java:312)
	at io.digdag.cli.Run$OperatorManagerWithSkip.callExecutor(Run.java:694)
	at io.digdag.core.agent.OperatorManager.runWithWorkspace(OperatorManager.java:254)
	at io.digdag.core.agent.OperatorManager.lambda$runWithHeartbeat$2(OperatorManager.java:137)
	at io.digdag.core.agent.LocalWorkspaceManager.withExtractedArchive(LocalWorkspaceManager.java:25)
	at io.digdag.core.agent.OperatorManager.runWithHeartbeat(OperatorManager.java:135)
	at io.digdag.core.agent.OperatorManager.run(OperatorManager.java:119)
	at io.digdag.cli.Run$OperatorManagerWithSkip.run(Run.java:676)
	at io.digdag.core.agent.MultiThreadAgent.lambda$null$0(MultiThreadAgent.java:127)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[...]

PR #694, opened 2017-12-02
Take care of dollar signs in secrets values

komamitsu:
https://docs.oracle.com/javase/8/docs/api/java/util/regex/Matcher.html#appendReplacement-java.lang.StringBuffer-java.lang.String-

> A dollar sign ($) may be included as a literal in the replacement string by preceding it with a backslash (\$).

This change addresses it and also adds a few cases that an user is already using an escaped secrets value (e.g. `value\$1234`) and this change works well with the workaround.

Fix https://github.com/treasure-data/digdag/issues/687


komamitsu:
@frsyuki Can you review this change when you get a chance?

komamitsu:
@frsyuki Kindly ping

frsyuki:
omg. LGTM!!

komamitsu:
Thanks

PR #695, opened 2017-12-04
Add docs for params option in server mode

szyn:
Hello.  
I added docs for params option in server mode. This is the useful options for users because we can set parameters that change for each environment.

### DIFF
--- a/digdag-docs/src/command_reference.rst
+++ b/digdag-docs/src/command_reference.rst
@@ -299,6 +299,16 @@ Runs a digdag server. --memory or --database option is required. Examples:
 
   Example: ``--disable-executor-loop``
 
+:command:`-p, --param KEY=VALUE`
+  Add a session parameter (use multiple times to set many parameters) in KEY=VALUE syntax. This parameter is availabe using ``${...}`` syntax in the YAML file, or using language API.
+
+  Example: ``-p environment=staging``
+
+:command:`-P, --params-file PATH`
+  Read parameters from a YAML file. Nested parameter (like {mysql: {user: me}}) are accessible using "." syntax (like \${mysql.user}).
+
+  Example: ``-P params.dig``
+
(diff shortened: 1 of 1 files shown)


digdag-docs/src/command_reference.rst
```diff
@@ -299,6 +299,16 @@ Runs a digdag server. --memory or --database option is required. Examples:
...
 
+:command:`-p, --param KEY=VALUE`
+  Add a session parameter (use multiple times to set many parameters) in KEY=VALUE syntax. This parameter is availabe using ``${...}`` syntax in the YAML file, or using language API.
+
+  Example: ``-p environment=staging``
+
+:command:`-P, --params-file PATH`
+  Read parameters from a YAML file. Nested parameter (like {mysql: {user: me}}) are accessible using "." syntax (like \${mysql.user}).
+
+  Example: ``-P params.dig``
```
komamitsu: This file isn't a workflow definition file, right? So `params.yml` (or `.yaml`) in the example seems better to me.

szyn: Thank you for your review! You are right, so I'll correct it. 

But, other parts of this documents also using `params.dig`, right? 
In order to fix those, I'll create another PR. Is this all right?

komamitsu: Good catch. I think `params.dig` in docs should've been fixed in https://github.com/treasure-data/digdag/commit/b8dc67a9d5525125484cdc406e0a7efec47f1ff6.

Let's fix other `params.dig` that aren't related to this change in another PR :)

szyn: Oh, I agree with you.  

Noted with thanks. After correcting this I'll also fix other `params.dig` in another PR ☺️

komamitsu:
LGTM. Thanks!

PR #700, opened 2017-12-08
add log-server.local.compress property

alu:
I added log-server.local.compress that for disable/enable to log compress (by gzip).
Set to the option true (default), task logs will compress as usual.
Set to the option false, task logs will keep plain text format. And then, we can read logs task in progress in realtime.

I think read logs as realtime is important some casesm such as so long time tasks.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
@@ -220,12 +220,12 @@ public class LocalFileLogServerFactory
 
                     if (realtimeFollow) {
                         // Delaying for deterrence duplicate output in digdag log --follow
-                        try {
-                            Thread.sleep(1000);
-                        }
-                        catch (InterruptedException ex) {
-                            Thread.currentThread().interrupt();
-                        }
+                        // try {
+                        //     Thread.sleep(1000);
+                        // }
+                        // catch (InterruptedException ex) {
+                        //     Thread.currentThread().interrupt();
+                        // }
 
                         try (InputStream in = Files.newInputStream(temporaryPath);
                              OutputStream out = new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024)) {
--- a/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
@@ -219,14 +219,6 @@ public class LocalFileLogServerFactory
                     output.close();
 
                     if (realtimeFollow) {
-                        // Delaying for deterrence duplicate output in digdag log --follow
-                        // try {
-                        //     Thread.sleep(1000);
-                        // }
-                        // catch (InterruptedException ex) {
-                        //     Thread.currentThread().interrupt();
-                        // }
-
                         try (InputStream in = Files.newInputStream(temporaryPath);
                              OutputStream out = new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024)) {
                             ByteStreams.copy(in, out);
--- a/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
@@ -137,7 +137,8 @@ public class LocalFileLogServerFactory
 
                         return byteOutput.toByteArray();
                     }
-                } else {
+                }
+                else {
                     return ByteStreams.toByteArray(in);
                 }
             }
@@ -187,7 +188,8 @@ public class LocalFileLogServerFactory
 
                 if (realtimeFollow) {
                     this.output = Files.newOutputStream(temporaryPath, CREATE, APPEND);
-                } else {
+                }
+                else {
                     this.output = new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024);
                 }
             }
@@ -218,13 +220,18 @@ public class LocalFileLogServerFactory
 
                     if (realtimeFollow) {
                         // Delaying for deterrence duplicate output in digdag log --follow
-                        try { Thread.sleep(1000); }
-                        catch (InterruptedException ex) {}
+                        try {
+                            Thread.sleep(1000);
+                        }
+                        catch (InterruptedException ex) {
+                            Thread.currentThread().interrupt();
+                        }
 
                         try (InputStream in = Files.newInputStream(temporaryPath);
                              OutputStream out = new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024)) {
                             ByteStreams.copy(in, out);
-                        } finally {
+                        }
+                        finally {
                             temporaryPath.toFile().delete();
                         }
                     }
--- a/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
@@ -133,7 +133,7 @@ public class LocalFileLogServerFactory
                     try (ByteArrayOutputStream byteOutput = new ByteArrayOutputStream();
                          GZIPOutputStream gzipOutput = new GZIPOutputStream(byteOutput)) {
                         ByteStreams.copy(in, gzipOutput);
-                        gzipOutput.close();
+                        gzipOutput.finish();
 
                         return byteOutput.toByteArray();
                     }
--- a/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
@@ -7,6 +7,7 @@ import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.IOException;
 import java.io.FileNotFoundException;
+import java.io.ByteArrayOutputStream;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.nio.file.FileSystems;
@@ -39,6 +40,7 @@ public class LocalFileLogServerFactory
 
     private final Path logPath;
     private final AgentId agentId;
+    private final boolean enableCompress;
 
     @Inject
     public LocalFileLogServerFactory(Config systemConfig, AgentId agentId)
@@ -47,6 +49,7 @@ public class LocalFileLogServerFactory
             .toAbsolutePath()
             .normalize();
         this.agentId = agentId;
+        this.enableCompress = systemConfig.get("log-server.local.compress", boolean.class, true);
     }
 
     @Override
@@ -126,7 +129,24 @@ public class LocalFileLogServerFactory
         {
             Path path = getPrefixDir(dateDir, attemptDir).resolve(fileName);
             try (InputStream in = Files.newInputStream(path)) {
-                return ByteStreams.toByteArray(in);
+                if (path.toString().endsWith(LogFiles.LOG_PLAIN_TEXT_FILE_SUFFIX)) {
+                    ByteArrayOutputStream byteOutput = new ByteArrayOutputStream();
+                    GZIPOutputStream gzipOutput = new GZIPOutputStream(byteOutput);
+                    byte[] buffer = new byte[1024];
+                    while (true) {
+                        int len = in.read(buffer);
+                        if (len < 0) {
+                            break;
+                        }
+                        gzipOutput.write(buffer, 0, len);
+                    }
+                    byteOutput.close();
+                    gzipOutput.close();
+
+                    return byteOutput.toByteArray();
+                } else {
+                    return ByteStreams.toByteArray(in);
+                }
             }
             catch (FileNotFoundException ex) {
                 throw new StorageFileNotFoundException(ex);
@@ -161,13 +181,17 @@ public class LocalFileLogServerFactory
             {
                 String dateDir = LogFiles.formatDataDir(prefix);
                 String attemptDir = LogFiles.formatSessionAttemptDir(prefix);
-                String fileName = LogFiles.formatFileName(taskName, Instant.now(), agentId.toString());
 
+                String fileName = enableCompress ?
+                    LogFiles.formatFileName(taskName, Instant.now(), agentId.toString()) :
+                    LogFiles.formatPlainTextFileName(taskName, Instant.now(), agentId.toString());
                 Path dir = getPrefixDir(dateDir, attemptDir);
                 Files.createDirectories(dir);
                 Path path = dir.resolve(fileName);
 
-                this.output = new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024);
+                this.output = enableCompress ?
+                    new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024) :
+                    Files.newOutputStream(path, CREATE, APPEND);
             }
 
             @Override
--- a/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
@@ -130,20 +130,13 @@ public class LocalFileLogServerFactory
             Path path = getPrefixDir(dateDir, attemptDir).resolve(fileName);
             try (InputStream in = Files.newInputStream(path)) {
                 if (path.toString().endsWith(LogFiles.LOG_PLAIN_TEXT_FILE_SUFFIX)) {
-                    ByteArrayOutputStream byteOutput = new ByteArrayOutputStream();
-                    GZIPOutputStream gzipOutput = new GZIPOutputStream(byteOutput);
-                    byte[] buffer = new byte[1024];
-                    while (true) {
-                        int len = in.read(buffer);
-                        if (len < 0) {
-                            break;
-                        }
-                        gzipOutput.write(buffer, 0, len);
-                    }
-                    byteOutput.close();
-                    gzipOutput.close();
+                    try (ByteArrayOutputStream byteOutput = new ByteArrayOutputStream();
+                         GZIPOutputStream gzipOutput = new GZIPOutputStream(byteOutput)) {
+                        ByteStreams.copy(in, gzipOutput);
+                        gzipOutput.close();
 
-                    return byteOutput.toByteArray();
+                        return byteOutput.toByteArray();
+                    }
                 } else {
                     return ByteStreams.toByteArray(in);
                 }
--- a/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
@@ -231,13 +231,15 @@ public class LocalFileLogServerFactory
                              OutputStream out = new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024)) {
                             ByteStreams.copy(in, out);
                         }
-                        finally {
-                            temporaryPath.toFile().delete();
-                        }
                     }
                 }
                 catch (IOException ex) {
                     throw Throwables.propagate(ex);
+                }
+                finally {
+                    if (realtimeFollow) {
+                        temporaryPath.toFile().delete();
+                    }
                 }
             }
         }
--- a/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
@@ -39,6 +39,7 @@ public class LocalFileLogServerFactory
     private static final String LOG_GZ_FILE_SUFFIX = ".log.gz";
 
     private final Path logPath;
+    private final boolean realtimeFollow;
     private final AgentId agentId;
 
     @Inject
@@ -47,6 +48,7 @@ public class LocalFileLogServerFactory
         this.logPath = FileSystems.getDefault().getPath(systemConfig.get("log-server.local.path", String.class, "digdag.log"))
             .toAbsolutePath()
             .normalize();
+        this.realtimeFollow = systemConfig.get("log-server.realtime_follow", boolean.class, false);
         this.agentId = agentId;
     }
 
@@ -126,6 +128,7 @@ public class LocalFileLogServerFactory
             throws StorageFileNotFoundException
         {
             Path path = getPrefixDir(dateDir, attemptDir).resolve(fileName);
+            System.out.println("fllowing => " + path);
             try (InputStream in = Files.newInputStream(path)) {
                 if (path.toString().endsWith(LogFiles.LOG_PLAIN_TEXT_FILE_SUFFIX)) {
                     try (ByteArrayOutputStream byteOutput = new ByteArrayOutputStream();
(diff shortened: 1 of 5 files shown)


digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
```diff
@@ -126,7 +129,24 @@ protected void listFiles(String dateDir, String attemptDir, FileMetadataConsumer
         {
             Path path = getPrefixDir(dateDir, attemptDir).resolve(fileName);
             try (InputStream in = Files.newInputStream(path)) {
-                return ByteStreams.toByteArray(in);
+                if (path.toString().endsWith(LogFiles.LOG_PLAIN_TEXT_FILE_SUFFIX)) {
+                    ByteArrayOutputStream byteOutput = new ByteArrayOutputStream();
```
komamitsu: These codes can be written simpler like this.
```
    try (ByteArrayOutputStream byteOutput = new ByteArrayOutputStream();
            GZIPOutputStream gzipOutput = new GZIPOutputStream(byteOutput)) {
        ByteStreams.copy(in, gzipOutput);
        return byteOutput.toByteArray();
    }
```

digdag-core/src/main/java/io/digdag/core/log/LogFiles.java
```diff
@@ -62,6 +63,16 @@ public static String formatFileName(String taskName, Instant firstLogTime, Strin
                 agentId) + LOG_GZ_FILE_SUFFIX;
     }
 
+    public static String formatPlainTextFileName(String taskName, Instant firstLogTime, String agentId)
+    {
+        return String.format(ENGLISH,
```
komamitsu: This code is duplicated with `formatFileName()`. How about extracting this `String.format()` to a private method and calling it from `formatFileName()` and `formatPlainTextFileName()`?

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
alu: I tried your code but always returned empty response from API server so i added ` gzipOutput.close()` before return byte array.

It's OK?

digdag-core/src/main/java/io/digdag/core/log/LogFiles.java
alu: I added generic private function that is `formatFileNameWithSuffix` and then both two methods to be used it.
I made it to be used

alu:
Thank you for review.

> You probably want to watch log messages with tail -f or something, right?

Your right.
We want to watch status or progress workflows from logs.

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
komamitsu: Oops. 👍 

Maybe `gzipOutput.finish()` might be slightly better in terms of understanding the intension from the code.

alu: Thanks, changed it.

digdag-core/src/main/java/io/digdag/core/log/LogFiles.java
```diff
@@ -27,6 +28,15 @@ private LogFiles()
     private static DateTimeFormatter SESSION_TIME_FORMATTER =
         DateTimeFormatter.ofPattern("yyyyMMdd'T'HHmmssxx", ENGLISH);
 
+    private static String formatFileNameWithSuffix(String taskName, Instant firstLogTime, String agentId, String suffix) {
```
szyn: The codes in this project follow [airlift code style](https://github.com/airlift/codestyle). 
Therefore, this `function` need to add a new line.

``` java
    private static String formatFileNameWithSuffix(String taskName, Instant firstLogTime, String agentId, String suffix)
    {
    …
    }
```

komamitsu: Good catch!

@alu Could you address it?

alu: Thanks, i did apply the style!

frsyuki:
Why don't you use REST API to read task logs? `digdag log -f <attempt id>` command lets you to follow task logs using the RES API.
This change itself breaks `digdag log` command because it assumes that the files are compressed at [TaskLogWatcher](https://github.com/alu/digdag/blob/master/digdag-cli/src/main/java/io/digdag/cli/client/TaskLogWatcher.java#L120), and server also assumes content-type of the log files is `application/gzip` at [LogResource](https://github.com/alu/digdag/blob/5cf357243c45b8f40bb1a34bf8873b0f818bbf8f/digdag-server/src/main/java/io/digdag/server/rs/LogResource.java#L120).

frsyuki:
digdag-ui also decompress log files without checking content-type [here](https://github.com/treasure-data/digdag/blob/8a6b676a51b0fd246ce153509f70d4b4e32c0653/digdag-ui/console.jsx#L1767). Actually it can't get content-type from digdag-server because it also uses pre-signed URL which allows digdag-ui fetch log files from S3 directly.

alu:
@frsyuki 
I tried `digdag log -f` with process below.

```sh
# Start digdag server at local
$ digdag server --task-log /tmp/digdag/log --memory
```

```sh
# Push test case
$ digdag push --endpoint 127.0.0.1:65432 --project test test
```

```sh
# Start test case
$ digdag start --endpoint 127.0.0.1:65432 test test --session "$(date '+%Y-%m-%d %H:%M:%S')"
2017-12-21 08:38:39 +0900: Digdag v0.9.20
Started a session attempt:
  session id: 2
  attempt id: 2
  uuid: e85c316d-bcc1-44c0-84c3-8c897de399a3
  project: test
  workflow: test
  session time: 2017-12-21 08:38:39 +0900
  retry attempt name: 
  params: {}
  created at: 2017-12-21 08:38:40 +0900

* Use `digdag session 2` to show session status.
* Use `digdag task 2` and `digdag log 2` to show task status and logs.
```

```sh
# Try following the log
$ digdag logs --endpoint 127.0.0.1:65432 2 -f
```

But digdag server does not respond anything until the task finished.

Test case is below

```yml
# test.dig
+step1:
  sh>: ./test.sh
```

```sh
#!/usr/bin/env bash

# test.sh
for i in {1..600}; do
  echo $i

  sleep 1s
done
```

My process is wrong? Or oversight something?

> digdag log command because it assumes that the files are compressed

I changed logic to generate gzip stream if a log file is plaintext at [here](https://github.com/treasure-data/digdag/pull/700/files#diff-a6b7fa13be3d8e4331fd68ae0a25597fR132).
So cli and ui are can fetch as gzip always.
Is not enough in some cases?

Thanks.

frsyuki:
you're using a scripting operator which is not polling-based...those operators won't upload task logs until until 16MB threshold at [BufferedRemoteTaskLogger](https://github.com/treasure-data/digdag/blob/8a6b676a51b0fd246ce153509f70d4b4e32c0653/digdag-core/src/main/java/io/digdag/core/log/BufferedRemoteTaskLogger.java#L65).

frsyuki:
> I changed logic to generate gzip stream if a log file is plaintext at here.
> So cli and ui are can fetch as gzip always.

In our use case, UI downloads data directly from S3 by using [this DirectDownloadHandle](https://github.com/treasure-data/digdag/pull/700/files#diff-9b620161896955ad0ced22346c19ca08L95) generated at [S3Storage](https://github.com/treasure-data/digdag/blob/8a6b676a51b0fd246ce153509f70d4b4e32c0653/digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3Storage.java#L197-L206). UI doesn't call AbstractFileLogServer#getFile method to fetch contents. Instead, it fetches contents from S3 directly. So UI needs to have a logic to tell that whether the file is gzipped or not.

frsyuki:
you assume that users have ssh access to the digdag server and can read files using a tail command. Is that true for long term? If not, why don't you make UPLOAD_THRESHOLD configurable and make it smaller?

alu:
> In our use case, UI downloads data directly from S3 by using this DirectDownloadHandle generated at S3Storage.

I see, it certainly is.

> you assume that users have ssh access to the digdag server and can read files using a tail command. Is that true for long term?

It's not an assumed way.
My goal is we become following logs the workflow in progress by `digdag log -f` command or web UI.

I tested change `UPLOAD_THRESHOLD` to 1 but result same as [prev test](https://github.com/treasure-data/digdag/pull/700#issuecomment-353217730).
The parameter will affect just use S3 for log storage?

frsyuki:
How about adding an option to LocalFileLogServerFactory.LocalFileDirectTaskLogger so that it uses uncompressed output stream and compresses the file when it closes the file?
You will see uncompressed growing files when a task is running, and then compressed files when it completes.

alu:
I Change behavior of  log file creating
    
1. Create plain text log file until the workflow finishied
2. When the workflow finished then convert to gzip log file from plain text
3. Delete plain text log file

How about it?

Thanks.

alu:
@frsyuki Hi, happy new year.
Can i get any news for this PR?

alu:
@frsyuki My children are waiting for this function, tell me the status please 🙇

komamitsu:
Sorry for our late reply...

This change modifies the current default behaviour of logging. Current logging outputs as a Gzip stream directly. But in this change, it outputs as a plain stream and then compresses it.

I have a concern that this change can degrade default logging performance.... This feature should be optional, I think.

alu:
@komamitsu 
Thanks for reply!

> I have a concern that this change can degrade default logging performance
I agree.
Do you have any ideas that is appropriate name of the option?

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
```diff
@@ -192,7 +207,14 @@ public void log(byte[] data, int off, int len)
...
             {
                 try {
-                    output.close();
+                    temporaryOutput.close();
+
+                    try (InputStream in = Files.newInputStream(temporaryPath);
+                         OutputStream conclusiveOutput = new GZIPOutputStream(Files.newOutputStream(conclusivePath, CREATE, APPEND), 16*1024)) {
+                        ByteStreams.copy(in, conclusiveOutput);
+                        conclusiveOutput.close();
+                        temporaryPath.toFile().delete();
```
komamitsu: Delete the temporary file in `finally` block.

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
```diff
@@ -192,7 +207,14 @@ public void log(byte[] data, int off, int len)
             public void close()
             {
                 try {
-                    output.close();
+                    temporaryOutput.close();
+
+                    try (InputStream in = Files.newInputStream(temporaryPath);
+                         OutputStream conclusiveOutput = new GZIPOutputStream(Files.newOutputStream(conclusivePath, CREATE, APPEND), 16*1024)) {
+                        ByteStreams.copy(in, conclusiveOutput);
+                        conclusiveOutput.close();
```
komamitsu: This stream should be closed by try-with-resources, so this line can be removed?

komamitsu:
How about `log-server.realtime_tail` or something?

BTW, I left some comments. Could you check them when you get a chance?

komamitsu:
> How about log-server.realtime_tail or something?

Ah, `log-server.realtime_follow` is better in terms of consistency with `digdag log -f` option

alu:
@komamitsu 
Thanks for review, and i applied those!

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
```diff
@@ -177,12 +180,17 @@ public LocalFileDirectTaskLogger(LogFilePrefix prefix, String taskName)
...
+
                 String temporaryFileName = LogFiles.formatPlainTextFileName(taskName, Instant.now(), agentId.toString());
                 this.temporaryPath = dir.resolve(temporaryFileName);
-                this.temporaryOutput = Files.newOutputStream(temporaryPath, CREATE, APPEND);
 
-                String conclusiveFileName = LogFiles.formatFileName(taskName, Instant.now(), agentId.toString());
-                this.conclusivePath = dir.resolve(conclusiveFileName);
+                if (realtimeFollow) {
+                    this.output = Files.newOutputStream(temporaryPath, CREATE, APPEND);
+                } else {
```
komamitsu: Could you follow the code format which is applied to Digdag?

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
```diff
@@ -207,13 +215,19 @@ public void log(byte[] data, int off, int len)
...
-                    try (InputStream in = Files.newInputStream(temporaryPath);
-                         OutputStream conclusiveOutput = new GZIPOutputStream(Files.newOutputStream(conclusivePath, CREATE, APPEND), 16*1024)) {
-                        ByteStreams.copy(in, conclusiveOutput);
-                        conclusiveOutput.close();
-                        temporaryPath.toFile().delete();
+                    output.close();
+
+                    if (realtimeFollow) {
+                        // Delaying for deterrence duplicate output in digdag log --follow
+                        try { Thread.sleep(1000); }
```
komamitsu: Same as above < code format

Where did this 1 sec come from?
Also, this can affect the performance of enqueuing tasks of io.digdag.core.agent.MultiThreadAgent#run...

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
```diff
@@ -207,13 +215,19 @@ public void log(byte[] data, int off, int len)
...
             {
                 try {
-                    temporaryOutput.close();
-
-                    try (InputStream in = Files.newInputStream(temporaryPath);
-                         OutputStream conclusiveOutput = new GZIPOutputStream(Files.newOutputStream(conclusivePath, CREATE, APPEND), 16*1024)) {
-                        ByteStreams.copy(in, conclusiveOutput);
-                        conclusiveOutput.close();
-                        temporaryPath.toFile().delete();
+                    output.close();
```
komamitsu: What if this call throws an exception? The temporary file will be remained forever?

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
```diff
@@ -207,13 +215,19 @@ public void log(byte[] data, int off, int len)
...
+
+                    if (realtimeFollow) {
+                        // Delaying for deterrence duplicate output in digdag log --follow
+                        try { Thread.sleep(1000); }
+                        catch (InterruptedException ex) {}
```
komamitsu: Please don't consume an interruption... Call `Thread.currentThread().interrupt()`

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
alu: This sleep for duplicate output in `digdag log --follow`.
If do not this, output will duplicate such as

```
$ digdag log 1 --follow
1 (from plain text)
2 (from plain text)
3 (from plain text)
4 (from plain text)
5 (from plain text)
(finish the workflow, plan text file will be removed, and gzip file will be created)
1 (from gzip all of below)
2
3
4
5
```

I think this way is not well, can we solve the issue another way?

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
alu: OK, thanks.

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
alu: Yes remained, should we retrying deletion?

alu:
@komamitsu Could you review a new commit?

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
```diff
@@ -193,6 +217,24 @@ public void close()
...
+                        catch (InterruptedException ex) {
+                            Thread.currentThread().interrupt();
+                        }
+
+                        try (InputStream in = Files.newInputStream(temporaryPath);
+                             OutputStream out = new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024)) {
+                            ByteStreams.copy(in, out);
+                        }
+                        finally {
+                            temporaryPath.toFile().delete();
```
komamitsu: https://github.com/treasure-data/digdag/pull/700#discussion_r176675750

> Yes remained, should we retrying deletion?

No. How about just deleting it in `finally` block like this?

```java
     if (realtimeFollow) {
         // Delaying for deterrence duplicate output in digdag log --follow
         try {
             Thread.sleep(1000);
         }
         catch (InterruptedException ex) {
             Thread.currentThread().interrupt();
         }

         try {
             output.close();

             try (InputStream in = Files.newInputStream(temporaryPath);
                 OutputStream out = new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024)) {
                 ByteStreams.copy(in, out);
             }
         }
         finally {
             temporaryPath.toFile().delete();
         }
    }
    else {
         output.close();
    }
```

alu: @komamitsu How catch IOException from `output.close();`?

alu: How is this?

```java
try {
    output.close();

    if (realtimeFollow) {
        // Delaying for deterrence duplicate output in digdag log --follow
        try {
            Thread.sleep(1000);
        }
        catch (InterruptedException ex) {
            Thread.currentThread().interrupt();
        }

        try (InputStream in = Files.newInputStream(temporaryPath);
             OutputStream out = new GZIPOutputStream(Files.newOutputStream(path, CREATE, APPEND), 16*1024)) {
            ByteStreams.copy(in, out);
        }
    }
}
catch (IOException ex) {
    throw Throwables.propagate(ex);
}
finally {
    if (realtimeFollow) {
        temporaryPath.toFile().delete();
    }
}
```

alu: @komamitsu Fixed it :D

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
```diff
@@ -193,10 +217,30 @@ public void close()
...
                     output.close();
+
+                    if (realtimeFollow) {
+                        // Delaying for deterrence duplicate output in digdag log --follow
+                        try {
+                            Thread.sleep(1000);
+                        }
+                        catch (InterruptedException ex) {
+                            Thread.currentThread().interrupt();
+                        }
```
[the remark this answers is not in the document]
alu: Thank you for comment!

The reasone is here.
https://github.com/treasure-data/digdag/pull/700#discussion_r176960045

(It may be confusing...)

alu:
@komamitsu @muga Hi! Can I get any news for this PR?

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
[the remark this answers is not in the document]
muga: Got it. Thank you for pointing it. Do you already know why log messages are duplicated? If you don't know, we'd better to take a look the behavior before we start improving it.

muga:
@alu Sorry, what do you think that my previous question doesn't make sense or not? > can you split and divide more smaller tasks? (Probably we need to understand your usecase..)

alu:
@muga 
To be honest, I do not understand your real intention.

digdag-core/src/main/java/io/digdag/core/log/LocalFileLogServerFactory.java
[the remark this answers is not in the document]
muga: Thanks! Let me think about it..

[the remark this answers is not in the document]
muga: @alu can you share small workflow example to reproduce duplicated logs?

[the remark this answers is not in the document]
alu: @muga 

My example is [here](https://github.com/treasure-data/digdag/pull/700#issuecomment-353217730).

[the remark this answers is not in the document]
muga: @alu Thank you. Got it and understand the behavior. Can you tell me why you need to print logs from gzip file instead of plain text in your duplicate scenario?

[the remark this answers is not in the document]
alu: @muga Because `/api/logs/{id}/files` returns gzip handles as final, so it is will be printed.

1. `example.log` will be created after the workflow started
2. `example.log` contents will be printed on console as real time (the api returns `example.log` handle)
3. The workflow completed
4. `example.log` will be compressed into gzip as `example.gzip`
5. `example.log` will be removed
6. The api returns `example.gzip` handle
7. Digdag client thinks `example.gzip` handle is new log file, so print it (duplicated!)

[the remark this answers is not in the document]
muga: Thanks @alu. Understand that it's not trivial for `LocalFileLogServerFactory` to recognize that files are already read or not. I think that `TaskLogWatcher` or `TaskLogState` can check that. What do you think that the client filters and skips `.gzip` files to show logs if the client uses `follow` option?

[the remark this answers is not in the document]
alu: @muga Thanks!

Sounds good to
>  .gzip files to show logs if the client uses follow option?

And implemented it in [this commmit](https://github.com/alu/digdag/commit/683f70d9e773f5d8dad4abaf30088eafa79793a3).

Bu we got new issues...

- `log-server.realtime_follow` is optional, so the new logick should judgment it was enabled (how is do it?)
- If an user exec `logs -f` when after finished a task, the user can't show finished task logs. For example
   1. task1 finished
   2. task1 logs compressed
   2. task2 started
   3. an user exec `logs -f` -> can't show task1 logs because the logs already compressed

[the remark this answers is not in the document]
alu: @muga Can i get some advices for my issues?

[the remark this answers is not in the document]
muga: > log-server.realtime_follow is optional, so the new logick should judgment it was enabled (how is do it?)

Not 100% sure but, my idea was that a watcher can filter and treat both of .log.gz and .log files as same file in `TaskLogState` to avoid duplication by new .log.gz file. If this idea would work for you, your second issue could be solved I think. What do you think?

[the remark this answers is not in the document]
alu: @muga 

I improve my code solved my 2 issues!
Skip show gzip log when a task log already shown as txt format!

How about it!?

Thanks!

alu:
@muga 
Hi!

How do you think my latest code?
I want news :)

alu:
@frsyuki @komamitsu @muga Hi!

Are there issues in latest version?
If so, could you tell me those?

Thanks :D

juncaofish:
Any updates for this pr? Need it..

juncaofish:
> Any updates for this pr? Need it..

?? Any updates for this feature

yoyama:
#1209 substituted this PR. So close.

PR #703, opened 2017-12-14
support virtualenv by directing python executable for python operaotor 

jo8937:
this code can python operator use virtualenv. with directing python executable path.
use like this.

py> Test.test
_pythonexec: /venv/Scripts/python

I know this approach is not recommanded way. 
But maybe useful for python user.


muga:
@jo8937 Sorry I could not follow up this PR but, it will be merged into master on https://github.com/treasure-data/digdag/pull/890.

muga:
This was merged into master as https://github.com/treasure-data/digdag/pull/890

PR #713, opened 2018-01-10
TransactionManager.autoCommit for submodules that may or may not nested

frsyuki:
Following-up of #635.

One thing we found is that transaction doesn't match well with
object-oriented programming because it always causes layer violation.

Idea here is making it possible to ignore about transactions when
transaction is unnecessary. In fact, most of cases don't need
transactions.

The new autoCommit method is allowed in nested way. All of following
code is valid:

```
autoCommit(() -> autoCommit(() -> ...))  // keeps auto commit mode
begin(() -> autoCommit(() -> ...))  // outer starts a transaction, inner shares it
autoCommit(() -> begin(() -> ...))  // outer runs as auto commit, inner starts a transaction
```


komamitsu:
LGTM! 👍

PR #716, opened 2018-01-17
[WIP] Prefer use of auto commit if transaction is unnecessary

frsyuki:
Use of transaction increases potential risk of nested transaction or
not-in-transaction issues. autoCommit can avoid it.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
@@ -81,8 +81,8 @@ public class InProcessTaskCallbackApi
         String taskName = request.getTaskName();
         LogFilePrefix prefix;
         try {
-            StoredSessionAttemptWithSession attempt =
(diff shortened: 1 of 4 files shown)


komamitsu:
Current `LocalThreadTransactionManager#autoCommit` has a potential problem that internal call of `BasicDatabaseStoreManager#transaction` can use an auto-commit mode transaction created by outer `autoCommit` and finish the transaction unexpectedly like:

1. `LocalThreadTransactionManager#autoCommit` is called and an auto commit mode transaction is stored in a thread local storage.
2. In the `autoCommit` method, `BasicDatabaseStoreManager#transaction` is called and it gets the auto commit mode transaction.
3. In the `transaction` method, it tries some operation expecting it's in a transaction.
4. But it's in an auto commit transaction and can commit at unexpected timing.

Even if the changes of this pull request are safe for now, it can cause unexpected behaviour with future changes that modify codes in `transaction`. What do you think?

frsyuki:
@komamitsu you're right...hmm...

yoyama:
I will close this PR because no progress for long time. Feel free to reopen, if needed.

PR #718, opened 2018-01-19
Run Notification outside of a transaction

frsyuki:
Notification module may assume that that is not in a transaction.


komamitsu:
This is very similar to what I would try 👍 

How about adding a comment like this?

```
diff --git a/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java b/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
index c8bfacf63..1f7e63dbc 100644
--- a/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
+++ b/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
@@ -234,6 +234,7 @@ public class WorkflowExecutionTimeoutEnforcer
                 .build();

         try {
+            // Assuming this method creates a new database transaction if needed
             notifier.sendNotification(notification);
         }
         catch (NotificationException e) {
```

PR #720, opened 2018-01-22
Cache opened Storage to reuse S3 threads

frsyuki:
OperatorManager calls ArchiveManager.openArchive through TaskCallbackApi
every time when a task starts. But because ArchiveManager.openArchive
doesn't cache opened Storage, it recreates Storage instance every time.
It ends up instantiating S3Storage every time. This overhead is big
because S3Storage creates ExecutorService for TransferManager.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/storage/ArchiveManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/storage/ArchiveManager.java
@@ -86,9 +86,9 @@ public class ArchiveManager
             .build(
                     new CacheLoader<ArchiveType, Storage>()
                     {
-                        public Storage load(ArchiveType key)
+                        public Storage load(ArchiveType type)
                         {
-                            return openStorage(key);
+                            return openStorage(type);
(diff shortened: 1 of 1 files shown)


digdag-core/src/main/java/io/digdag/core/storage/ArchiveManager.java
```diff
@@ -65,21 +70,38 @@ public String getPath()
...
     @Inject
     public ArchiveManager(StorageManager storageManager, Config systemConfig)
     {
         this.storageManager = storageManager;
+        this.storageCache = CacheBuilder.newBuilder()
+            .maximumSize(2)
+            .build(
+                    new CacheLoader<ArchiveType, Storage>()
+                    {
+                        public Storage load(ArchiveType key)
```
komamitsu: `type` would be more proper than `key`

frsyuki: 👍

PR #722, opened 2018-01-22
Add support for pool_name to td job operators

frsyuki:
Treasure Data jobs support pool_name parameter to run queries in an
isolated resource pool.

### DIFF
--- a/digdag-docs/src/operators/td.md
+++ b/digdag-docs/src/operators/td.md
@@ -183,6 +183,16 @@
     path: /endpoint
   ```
 
+* **pool_name**: NAME
+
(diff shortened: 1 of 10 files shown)


komamitsu:
Added some comments on the documents.

BTW, I'm wondering the new attribute can be sent to Presto. Or, we'd better add a validation for the combination of `engine` and `pool_name`?

frsyuki:
oh I totally put wrong document. Fixed!

frsyuki:
@komamitsu Presto also supports resource pools. It's TD API's bug if it's not available...

PR #723, opened 2018-01-23
Retry if ProjectArchives.extractTarArchive fails in ExtractArchiveWorkspaceManager

komamitsu:
We've observed a few cases that ProjectArchives.extractTarArchive failed in ExtractArchiveWorkspaceManager#withExtractedArchive due to unexpected EOF of gzipped archive. This workaround would mitigate that.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
@@ -68,7 +68,7 @@ public class ExtractArchiveWorkspaceManager
                         });
             }
             catch (RetryExecutor.RetryGiveupException e) {
-                Throwables.propagate(e);
+                throw Throwables.propagate(e.getCause());
             }
             return func.run(workspacePath.get());
         }
--- a/digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
@@ -31,7 +31,6 @@ public class ExtractArchiveWorkspaceManager
     public ExtractArchiveWorkspaceManager(TempFileManager tempFiles)
     {
         this.tempFiles = tempFiles;
-
     }
 
     @Override
--- a/digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
@@ -40,8 +40,12 @@ public class ExtractArchiveWorkspaceManager
         TempDir workspacePath = null;
         try {
             try {
-                // A temporary workspace should be created for each try
-                // And always close the temporary workspace
+                // Here has retrying because there were some observations that downloading of S3 caused too early EOF.
+                // S3Storage uses ResumableInputStream but it can't rescue from too early end of streams.
+                // Because ProjectResource.putProject validates the file, getting IOException here is always unexpected.
+                //
+                // A temporary workspace should be created for each try.
+                // And the created temporary workspace should be closed finally.
                 workspacePath = RetryExecutor.retryExecutor()
                         .retryIf(exception -> true)
                         .withInitialRetryWait(EXTRACT_MIN_RETRY_WAIT_MS)
--- a/digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
@@ -1,6 +1,7 @@
 package io.digdag.core.agent;
 
 import com.google.common.base.Optional;
+import com.google.common.base.Throwables;
 import com.google.inject.Inject;
 import io.digdag.core.TempFileManager.TempDir;
 import io.digdag.core.TempFileManager;
@@ -8,12 +9,20 @@ import io.digdag.core.archive.ProjectArchives;
 import io.digdag.spi.StorageObject;
 import io.digdag.spi.TaskRequest;
 import java.io.IOException;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicReference;
+
+import io.digdag.util.RetryExecutor;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class ExtractArchiveWorkspaceManager
     implements WorkspaceManager
 {
+    private static final int EXTRACT_RETRIES = 10;
+    private static final int EXTRACT_MIN_RETRY_WAIT_MS = 1000;
+    private static final int EXTRACT_MAX_RETRY_WAIT_MS = 30000;
+
     private final Logger logger = LoggerFactory.getLogger(getClass());
 
     private final TempFileManager tempFiles;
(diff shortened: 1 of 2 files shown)


komamitsu:
@frsyuki Can you take a look at this change?

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
```diff
@@ -22,19 +31,53 @@
...
+                            catch (Throwable e) {
+                                if (newWorkSpacePath != null) {
+                                    newWorkSpacePath.close();
+                                }
+                                throw e;
+                            }
+                        });
+            }
+            catch (RetryExecutor.RetryGiveupException e) {
+                Throwables.propagate(e);
```
frsyuki: I think `Throwables.propagate(e.getCause())` is better just to cleanup stacktrace in log messages.

And `Throwables.propagate` is marked as deprecated. I guess that it's because we often miss `throw` as you also did...Throwables.propagate returns a Throwable and doesn't throw.

So we should use `Throwables.throwIfUnchecked` like this instead:

```java
Throwables.throwIfUnchecked(e.getCause());
throw new RuntimeException(e.getCause());
```

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
```diff
@@ -22,19 +31,53 @@
     public ExtractArchiveWorkspaceManager(TempFileManager tempFiles)
     {
         this.tempFiles = tempFiles;
+
```
frsyuki: blank line typo?

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
```diff
@@ -22,19 +31,53 @@
...
+                            TempDir newWorkSpacePath = null;
+                            try {
+                                newWorkSpacePath = createNewWorkspace(request);
+                                Optional<StorageObject> in = archiveProvider.open();
+                                if (in.isPresent()) {
+                                    ProjectArchives.extractTarArchive(newWorkSpacePath.get(), in.get().getContentInputStream());
+                                }
+                                return newWorkSpacePath;
+                            }
+                            catch (Throwable e) {
```
frsyuki: you can use `finally` here and remove `throw e` at the end.

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
```diff
@@ -22,19 +31,53 @@
...
+                            }
+                        });
+            }
+            catch (RetryExecutor.RetryGiveupException e) {
+                Throwables.propagate(e);
             }
             return func.run(workspacePath.get());
         }
+        finally {
+            if (workspacePath != null) {
```
frsyuki: I think that this null check is unnecessary if you don't assign null when you declare `workspacePath`.
Or  how about using try-with-resource here like this?:

```java
try (TempDir  workspacePath = RetryExecutor.retryExecutor()...) {
    ...
}
catch (RetryExecutor.RetryGiveupException e) {
    ...
}
```

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
```diff
@@ -22,19 +31,53 @@
...
-        try (TempDir workspacePath = createNewWorkspace(request)) {
-            Optional<StorageObject> in = archiveProvider.open();
-            if (in.isPresent()) {
-                ProjectArchives.extractTarArchive(workspacePath.get(), in.get().getContentInputStream());
+        TempDir workspacePath = null;
+        try {
+            try {
+                // A temporary workspace should be created for each try
+                // And always close the temporary workspace
+                workspacePath = RetryExecutor.retryExecutor()
```
frsyuki: Let's add a comment here like....

> Here has retrying because there were some observations that downloading of S3 caused too early EOF. S3Storage uses ResumableInputStream but it can't rescue from too early end of streams. Because ProjectResource.putProject validates the file, getting IOException here is always unexpected.

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
komamitsu: Oops

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
komamitsu: With `finally` here, returned `TempDir` from here to outside RetryExecutor#run will be always closed and below `func.run` can't use, I think.

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
komamitsu: `Throwables` has these implementations as of that throw an exception not just returning an exception object and I can't find `@Deprecated` annotation on `propagate`. Also, I can't find `Throwables.throwIfUnchecked`

```
  public static RuntimeException propagate(Throwable throwable) {
    propagateIfPossible(checkNotNull(throwable));
    throw new RuntimeException(throwable);
  }

  public static void propagateIfPossible(@Nullable Throwable throwable) {
    propagateIfInstanceOf(throwable, Error.class);
    propagateIfInstanceOf(throwable, RuntimeException.class);
  }

  public static <X extends Throwable> void propagateIfInstanceOf(
      @Nullable Throwable throwable, Class<X> declaredType) throws X {
    // Check for null is needed to avoid frequent JNI calls to isInstance().
    if (throwable != null && declaredType.isInstance(throwable)) {
      throw declaredType.cast(throwable);
    }
  }
```

So, I think just `Throwables.propagate(e.getCause())` is enough.

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
komamitsu: In the other words, this `close` is only for when `ProjectArchives.extractTarArchive` fails

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
komamitsu: One thing I wanted to do in this change is to create a new workspace for each try. So workspaces need to be created in `RetryExecutor.run()`.

frsyuki: oh right. I was missing about it

digdag-core/src/main/java/io/digdag/core/agent/ExtractArchiveWorkspaceManager.java
frsyuki: doesn't this work?:
```
}
finally {
    if (newWorkSpacePath != null) {
        newWorkSpacePath.close();
    }
}
```

frsyuki: oh no. it doesn't work. now I got it....!

komamitsu:
Thanks!

PR #725, opened 2018-01-24
Add an instruction to run `git pull origin --tags` before release op

komamitsu:
@frsyuki PTAL.

Without this instruction, generated `digdag-docs/src/releases/release-x.x.x.rst` could include incorrect `Contributors`.


frsyuki:
Good point. LGTM 👍

PR #726, opened 2018-02-02
Add License to footer

toru-takahashi:
I heard there is a company whose lawyer rejected to use the Digdag logo because http://docs.digdag.io/logo.html just says “Feel free to use these logos” and does not specify the exact license to use them (like CC0, or some explicit description that makes sure it does require any fee to use it).


muga:
I fixed the message a bit. That's same as the footer in https://www.digdag.io/ What do you think?

muga:
Will merge it after CI completed.

Issue #736, opened 2018-03-15
Sites which run many tasks can affect the processing of other sites' tasks in MultiThreadAgent#run


komamitsu:
I took a look at `MultiThreadAgent#run` and noticed it calls `DatabaseTaskQueueServer#lockSharedAgentTasks` and `tryLockSharedAgentTasks(siteId)`.

If some sites run a lot of tasks concurrently up to 10 (by default), `tryLockSharedAgentTasks()` seems to wait for each site which is running many tasks probably holding locks on `queued_task_locks`. As a result, unrelated sites could be affected by those sites running many tasks, if I understand it correctly.

So, we'd better make those methods not blocked for task-consuming sites to guarantee to process queued tasks without delay. What do you think? CMIIW.

frsyuki:
I'm unclear whether you're mentioning about lock wait on the database or task execution wait.

If it's about lock wait on database, there is small overhead but effect is very small. If you use PostgreSQL, `tryLockSharedAgentTasks` uses `lock_shared_tasks` function. It uses `pg_try_advisory_xact_lock` to lock a siteId first. This returns immediately if the siteId is already locked. If you use H2 database, it's guaranteed that only one Java process accesses the database. Threads of the process uses `localLockMap` first to lock a siteId. If the siteId is already locked by another thread, it returns immediately. So it's effect is still very small.

If it's about task execution wait, it assumes that you have enough number of threads to execute tasks in fair manner. First, `lockSharedAgentTasks` iterates sites from smaller siteId to bigger (because `getActiveSiteIdList` behaves so). If smaller siteIds have many tasks and fill up all available threads of MultiThreadAgent, tasks of bigger siteIds won't start. But notice that number of threads of MultiThreadAgent is unlimited by default. It starts as many threads as needed. So advanced users who set `agent.max-task-threads` should monitor the queue, and make sure that total number of tasks in the queue is smaller than total number of threads of MultiThreadAgent. 

It's anyways good idea to randomize order of siteIds to iterate.

komamitsu:
Thanks for the explanation. Actually, I had a bit wrong idea on `MultiThreadAgent#run` and my concern above didn't make sense.

Anyway, the pull request looks good.

PR #737, opened 2018-03-15
Randomize iteration across multiple sites over the shared queue

frsyuki:
lockSharedAgentTasks was locking tasks of smaller siteIds first.
Randomizing siteIds makes it more fair so that bigger siteIds can get
threads even when smaller siteIds filled up limited number of threads
(default is unlimited though).

This also slighly improves overhead when smaller siteIds tend to have
more tasks than siteMaxConcurrency (10) because attempts to lock the
site using pg_try_advisory_xact_lock or LocalLockMap is not zero
overhead.

Related to #736


komamitsu:
LGTM! 👍

PR #708, opened 2018-01-04
Add dependencies for copied tasks for grouping task retry

serihiro:
# Abstract
- This PR fixes #701 (which I reported).
- Modify [DatabaseSessionStoreManager#copyInitialTasksForRetry](https://github.com/serihiro/digdag/blob/726aa07cac62baab5f6b2ee7fea6393b2cd664ae/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java#L848) to add dependencies for copied tasks, for grouping task retry.

# Detail
- In 0.9.21, when digdag executes the following workflow, `+task3` is retried despite that the upstream task(`+task2`) was failed.

```yaml
+first:
  echo>: ""
  append_file: out
+doit:
  _retry: 3
  +task1:
    echo>: "try"
    append_file: out
  +task2:
    fail>: task failed expectedly
  +task3:
    echo>: "skipped"
    append_file: out
```
- I think this is not expected behavior, and @komamitsu gave me a comment that this looks like a bug [here(Qiita comment)](https://qiita.com/seri_k/items/e97ad0957376641eedc0#comment-5ae9f19d9682c8f6aa42).
- Then, I found that copied tasks, for retrying, do not have any upstreams when they are generated in  `DatabaseSessionStoreManager#copyInitialTasksForRetry`.
- I thought this is the cause of this phenomenon.
- So, I modified to set the before tasks of each copied tasks as upstream tasks.

# Note
- This is my first PR in this product, so I wonder whether this is the best approach.
- For example, I used `DatabaseTaskControlStore#addDependencies` to add dependencies, but I'm not sure that using this method is the suitable in this case...
- So please point out if I have modified with wrong way. I'll fix it.
- This is my [research note](https://gist.github.com/serihiro/0bb10d2a9ddcbfde17dbd09bdd4cb2f3#%E8%AA%BF%E6%9F%BB%E4%B8%AD%E5%BE%85%E3%81%A6%E3%82%88retry%E3%81%AF%E3%81%A9%E3%81%93%E3%81%A7%E3%82%84%E3%81%A3%E3%81%A6%E3%82%93%E3%81%A0) (Sorry, this is written in Japanese)

PR #738, opened 2018-03-19
Add dependencies for copied tasks for grouping task retry including nested tasks

frsyuki:
Following-up of #708.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -882,7 +882,7 @@ public class DatabaseSessionStoreManager
             }
 
             DatabaseTaskControlStore store = new DatabaseTaskControlStore(handle);
-            boolean firstTask = true;
+            Map<Long, Long> oldIdToNewId = new HashMap<>();
             for (StoredTask task : tasks) {
                 Task newTask = Task.taskBuilder()
                     .from(task)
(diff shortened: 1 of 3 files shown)


digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -871,20 +872,31 @@ public boolean copyInitialTasksForRetry(List<Long> recursiveChildrenIdList)
...
                     .state(TaskStateCode.BLOCKED)
                     .stateFlags(TaskStateFlags.empty())
                     .build();
-                addSubtask(tasks.get(0).getAttemptId(), newTask);
+                long newTaskId = addSubtask(tasks.get(0).getAttemptId(), newTask);
+
+                // addSubtask doesn't copy fields of StoredTask because its second argument is Task.
+                // Copy StoredTask::getUpstreams here
+                oldIdToNewId.put(task.getId(), newTaskId);
+                for (long oldUpstreamId : task.getUpstreams()) {
```
komamitsu: 1. It seems the task ID mapping table `oldIdToNewId` isn't completely built up here. So new dependencies can use old task ID, I guess. Don't we need to build `oldIdToNewId` first and then add dependencies?

2. `DatabaseTaskControlStore#addDependencies` is a more abstract API, so we'd better use it?

IIUC, those codes can be like this.
```java
            Map<Long, Long> oldIdToNewId = new HashMap<>();
            for (StoredTask task : tasks) {
                Task newTask = Task.taskBuilder()
                        .from(task)
                        .state(TaskStateCode.BLOCKED)
                        .stateFlags(TaskStateFlags.empty())
                        .build();
                long newTaskId = addSubtask(tasks.get(0).getAttemptId(), newTask);

                // Build task id mapping table
                oldIdToNewId.put(task.getId(), newTaskId);
            }

            for (StoredTask task : tasks) {
                long newTaskId = oldIdToNewId.get(task.getId());

                // addSubtask doesn't copy fields of StoredTask because its second argument is Task.
                // Copy StoredTask::getUpstreams here
                store.addDependencies(newTaskId,
                        task.getUpstreams().stream()
                        .map(
                                upstreamId -> {
                                    Long newUpstreamId = oldIdToNewId.get(upstreamId);
                                    if (newUpstreamId == null) {
                                        return upstreamId;
                                    }
                                    else {
                                        return newUpstreamId;
                                    }
                                })
                        .collect(Collectors.toList()));
            }
```

frsyuki: 1.

Dependency is possible only from a later task to a former task. Dependent task is is always smaller. So we don’t have to prebuild the mapping.

2.

I also thought it first but I thought that dao.insertTaskDependency is already well abstracted

frsyuki:
Good point. I add one

PR #740, opened 2018-03-23
[Docs] Fix wrong arguments

szyn:
I updated docs about command reference because some arguments do not work correctly.
I think these docs should have fixed in https://github.com/treasure-data/digdag/commit/d8785a11ba7750ef0388c24d808b0cb2447bcdd0.

### DIFF
--- a/digdag-docs/src/command_reference.rst
+++ b/digdag-docs/src/command_reference.rst
@@ -405,16 +405,16 @@ start
 
 .. code-block:: console
 
-    $ digdag start <project-name> <+name> --session <hourly | daily | now | yyyy-MM-dd | "yyyy-MM-dd HH:mm:ss">
+    $ digdag start <project-name> <workflow-name> --session <hourly | daily | now | yyyy-MM-dd | "yyyy-MM-dd HH:mm:ss">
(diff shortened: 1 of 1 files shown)


muga:
Good catch. Thank you! We don't need "+" sign in workflow name any more.

I have one concern about the use of `workflow-name`. According to the commit d8785a11ba7750ef0388c24d808b0cb2447bcdd0 that you pointed, we'd better to rename `+name` argument with `name` instead of `workflow-name` to keep consistency of argument names in examples. I don't say that `workflow-name` is not preferable here and just wanted to care of such consistency. What do you think?

Or you also may want to change all `name` arguments in the source code as well? For example, around `cli.Main` https://github.com/treasure-data/digdag/blob/master/digdag-cli/src/main/java/io/digdag/cli/Main.java#L292-L343

szyn:
Thank you for pointing that out.

I agree with you. In this case, should have to keep consistency of argument names.
Considering this fact, I'll correct that from `workflow-name` to `name`.

muga:
Cool. Thank you for fixing. I merge it.

PR #748, opened 2018-04-03
Fix typo. Duration needs a unit.

### LINKED ISSUE #747: Can't set socket_timeout option of redshfit operator
in digdag (documents) [https://docs.digdag.io/operators/redshift.html],

>The timeout value used for socket read operations. If reading from the server takes longer than this value, the connection is closed. Default: 1800.

but, following configuration does not work.

```
redshift:
  socket_timeout: 1800
```

Error:

```
2018-04-03 19:20:13 +0900 [ERROR] (0019@[0:default]+task): Configuration error at task +task: Expected class io.digdag.util.DurationParam for key 'socket_timeout' but got 1800 (number) (config)
> Can not instantiate value of type [simple type, class io.digdag.util.DurationParam] from Integral number (1800); no single-int-arg constructor/factory method
 at [Source: N/A; line: -1, column: -1] (json mapping)
```

hiroyuki-sato:
Fix #747

### DIFF
--- a/digdag-docs/src/operators/redshift.md
+++ b/digdag-docs/src/operators/redshift.md
@@ -202,20 +202,20 @@ _export:
 
 * **connect_timeout**: NAME
 
-  The timeout value used for socket connect operations. If connecting to the server takes longer than this value, the connection is broken. *Default*: `30`.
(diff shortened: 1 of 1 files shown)


hiroyuki-sato:
Thank you!

Issue #536, opened 2017-04-05
Skip backflow


cosmok:
Hi,

When the digdag server restarts it tries to backfill/catchup all the schedules for which digdag was not running.

This is fine most of time, but, with heavy operations, backfill can slow the server down considerably.

I suggest that a new option be added (say disable_backfill) to schedule config that will disable backfill.

e.g,
```
schedule:
  minutes_interval>: 1
  disable_backfill: true

```

frsyuki:
Good point.

hiroyuki-sato:
Hello, @cosmok 

The `skip_delayed_by` introduced in Digdag 0.9.25.
Please try it!

muga:
@cosmok I will close this ticket after https://github.com/treasure-data/digdag/issues/782 complete. Please let us know if you have any questions.

hiroyuki-sato:
Hello, @cosmok 

Could you check this test?
https://gist.github.com/hiroyuki-sato/4e9356c3492b2e6df699b2338524f89f#with-skip_delayed_by

```yaml
timezone: Asia/Tokyo

schedule:
  minutes_interval>: 1
  skip_delayed_by: 3m

+setup:
  sh>: echo start ${session_time}
```

* last_session_time: 2018-05-12:00
* Stopped time is 2018-05-28 15:12:30
* Restart time is 2018-05-28 15:19:00

It backfill as follows.
* Without skip_delayed_by: backfill 15:13:00 ~ 15:19:00
* With skip_delayed_by (3m): backfill 15:16:00(current_time - 3m) ~ 15:19:00

hiroyuki-sato:
@cosmok 
Your configuration is incorrect.
`skip_delayed_by:`  is correct, not `skip_delayed_by>:`

nyerramaneni:
Hi , in my case i want to skip all the sessions between pause and resume . How can I skip all. Is there any other option other than skip_delayed_by .

hiroyuki-sato:
Hello, @nyerramaneni 

Have you ever tried `skip_delayed_by: 1s`?
Or #946 and #995 may help.

If your use case doesn't make the same it, could you tell us more detail about it?

nyerramaneni:
Hi , @hiroyuki-sato 
yes tried with that it works for me . Other than this is there any posibility to stop paused sessions.

[...]

PR #752, opened 2018-04-07
implement disable_backfill.(issue #536)

osiire:
Hi
because I need, I have just implemented **disable_backfill**  option as issue #536 mentioned.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
+++ b/digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
@@ -36,6 +36,7 @@ import io.digdag.spi.ScheduleTime;
 import io.digdag.spi.Scheduler;
 import io.digdag.core.session.ImmutableStoredSessionAttempt;
 import io.digdag.client.config.ConfigFactory;
+import io.digdag.util.DurationParam;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -199,10 +200,10 @@ public class ScheduleExecutor
 
             Config scheduleConfig = SchedulerManager.getScheduleConfig(def);
             boolean skipOnOvertime = scheduleConfig.get("skip_on_overtime", boolean.class, false);
-            Optional<Integer> skipDelay = scheduleConfig.getOptional("skip_delayed_by", Integer.class);
+            Optional<DurationParam> skipDelay = scheduleConfig.getOptional("skip_delayed_by", DurationParam.class);
 
             // task should run at scheduled time within skipDelay.
-            if (skipDelay.isPresent() && now.isAfter(sched.getNextRunTime().plusSeconds(skipDelay.get()))) {
+            if (skipDelay.isPresent() && now.isAfter(sched.getNextRunTime().plusSeconds(skipDelay.get().getDuration().getSeconds()))) {
                 logger.info("Now={} is too late from scheduled time={}. It's over skip_delayed_by={}. Skipping this schedule: {}", now, sched.getNextScheduleTime(), skipDelay.get(), sched);
                 nextSchedule = sr.nextScheduleTime(sched.getNextScheduleTime());
             }
--- a/digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
+++ b/digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
@@ -150,7 +150,7 @@ public class ScheduleExecutor
             // here uses limit=1 because selecting multiple rows with FOR UPDATE
             // has risk of too often deadlock.
             return sm.lockReadySchedules(now, 1, (store, storedSchedule) -> {
-                runSchedule(new ScheduleControl(store, storedSchedule));
+                runSchedule(new ScheduleControl(store, storedSchedule), now);
             });
         });
         return count > 0;
@@ -178,7 +178,7 @@ public class ScheduleExecutor
         }
     }
 
-    private void runSchedule(ScheduleControl lockedSched)
+    private void runSchedule(ScheduleControl lockedSched, Instant now)
     {
         StoredSchedule sched = lockedSched.get();
 
@@ -199,8 +199,14 @@ public class ScheduleExecutor
 
             Config scheduleConfig = SchedulerManager.getScheduleConfig(def);
             boolean skipOnOvertime = scheduleConfig.get("skip_on_overtime", boolean.class, false);
+            boolean disableBackfill = scheduleConfig.get("disable_backfill", boolean.class, false);
 
-            if (!activeAttempts.isEmpty() && skipOnOvertime) {
+            // task should run at scheduled time within one second delay if disable_backfill is true.
+            if (disableBackfill && now.isAfter(sched.getNextRunTime().plusSeconds(1))) {
+                logger.info("Now={} is later than scheduled time={} and disable_backfill is true. Skipping this schedule: {}", now, sched.getNextScheduleTime(), sched);
+                nextSchedule = sr.nextScheduleTime(sched.getNextScheduleTime());
+            }
+            else if (!activeAttempts.isEmpty() && skipOnOvertime) {
                 logger.info("An attempt of the scheduled workflow is still running and skip_on_overtime = true. Skipping this schedule: {}", sched);
                 nextSchedule = sr.nextScheduleTime(sched.getNextScheduleTime());
             }
--- a/digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
+++ b/digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
@@ -199,11 +199,11 @@ public class ScheduleExecutor
 
             Config scheduleConfig = SchedulerManager.getScheduleConfig(def);
             boolean skipOnOvertime = scheduleConfig.get("skip_on_overtime", boolean.class, false);
-            int backfillLimit = scheduleConfig.get("backfill_limit", int.class, -1);
+            Optional<Integer> skipDelay = scheduleConfig.getOptional("skip_delayed_by", Integer.class);
 
-            // task should run at scheduled time within backfill_limit delay.
-            if (backfillLimit >= 0 && now.isAfter(sched.getNextRunTime().plusSeconds(backfillLimit))) {
-                logger.info("Now={} is too late from scheduled time={}. It's over backfill_limit={}. Skipping this schedule: {}", now, sched.getNextScheduleTime(), backfillLimit, sched);
+            // task should run at scheduled time within skipDelay.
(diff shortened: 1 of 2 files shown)


digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
```diff
@@ -199,8 +199,13 @@ private void runSchedule(ScheduleControl lockedSched)
 
             Config scheduleConfig = SchedulerManager.getScheduleConfig(def);
             boolean skipOnOvertime = scheduleConfig.get("skip_on_overtime", boolean.class, false);
+            boolean disableBackfill = scheduleConfig.get("disable_backfill", boolean.class, false);
 
-            if (!activeAttempts.isEmpty() && skipOnOvertime) {
+            if (disableBackfill && now.isAfter(sched.getNextRunTime())) {
```
komamitsu: This `sched` comes from `io.digdag.core.database.DatabaseScheduleStoreManager#lockReadySchedules` and the method selects schedules which have older next run time than current time https://github.com/treasure-data/digdag/blob/8cf4af37be92602595bb6b10a41ef1e5961f3fb2/digdag-core/src/main/java/io/digdag/core/database/DatabaseScheduleStoreManager.java#L253

So it seems you don't need to check scheduler's next run time here.

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
```diff
@@ -199,8 +199,13 @@ private void runSchedule(ScheduleControl lockedSched)
...
+                logger.info("Now={} is later than scheduled time={}. Skipping this schedule: {}", now, sched.getNextScheduleTime(), sched);
```
komamitsu: Could you modify this message to say that the schedule will be skipped since backfill is disable?

digdag-core/src/test/java/io/digdag/core/schedule/ScheduleExecutorTest.java
```diff
@@ -163,4 +163,50 @@ public void testDefaultConcurrentOvertimeExecution()
         // Verify that the schedule progressed to the next time
         verify(scs).updateNextScheduleTimeAndLastSessionTime(SCHEDULE_ID, nextScheduleTime, now);
     }
+
+    @Test
+    public void testDisableBackfill()
+            throws Exception
+    {
+        // set backfill_limit to 30 seconds
```
komamitsu: What's `backfill_limit` ?

digdag-core/src/test/java/io/digdag/core/schedule/ScheduleExecutorTest.java
```diff
@@ -163,4 +163,50 @@ public void testDefaultConcurrentOvertimeExecution()
...
+                .set("daily>", "12:00:00");
+
+        // Indicate that there is no active attempt for this workflow
+        when(sessionStore.getActiveAttemptsOfWorkflow(eq(PROJECT_ID), eq(WORKFLOW_NAME), anyInt(), any(Optional.class)))
+                .thenReturn(ImmutableList.of());
+
+        // Run the schedule executor...
+        scheduleExecutor.runScheduleOnce(now.plusSeconds(1));
+
+        // Verify that another attempt was not started
```
komamitsu: Is this comment correct?

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
osiire: No. This condition is important part.  where clause you mentioned is "after equal". the code is "after". it's means the only when just "next_run_time" is allowed to run.

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
osiire: done.

digdag-core/src/test/java/io/digdag/core/schedule/ScheduleExecutorTest.java
osiire: my mistake. corrected

digdag-core/src/test/java/io/digdag/core/schedule/ScheduleExecutorTest.java
osiire: corrected.

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
osiire: I found the condition is too strict.
new code allows one second delay to run.

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
```diff
@@ -199,8 +199,14 @@ private void runSchedule(ScheduleControl lockedSched)
 
             Config scheduleConfig = SchedulerManager.getScheduleConfig(def);
             boolean skipOnOvertime = scheduleConfig.get("skip_on_overtime", boolean.class, false);
+            boolean disableBackfill = scheduleConfig.get("disable_backfill", boolean.class, false);
 
-            if (!activeAttempts.isEmpty() && skipOnOvertime) {
+            // task should run at scheduled time within one second delay if disable_backfill is true.
+            if (disableBackfill && now.isAfter(sched.getNextRunTime().plusSeconds(1))) {
```
komamitsu: I think the best and safest delay depends on use cases and situations. For instance, if Digdag is handling a lot of tasks, it's likely Digdag tries to start handling a task whose next run time is already a few second ago even in normal situation.

How about making the delay configurable not boolean?

komamitsu: Skipping tasks that have recent `StoredSchedule#getCreatedAt` might also work.

osiire: OK, I'll change the keyword from `disable_backfill` to `backfill_limit`.
e.g. forbid backfill execution if it's over 30 seconds from scheduled time.
backfill_limit: 30

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
```diff
@@ -199,8 +199,14 @@ private void runSchedule(ScheduleControl lockedSched)
 
             Config scheduleConfig = SchedulerManager.getScheduleConfig(def);
             boolean skipOnOvertime = scheduleConfig.get("skip_on_overtime", boolean.class, false);
+            int backfillLimit = scheduleConfig.get("backfill_limit", int.class, -1);
```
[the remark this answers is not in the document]
komamitsu: `backfill_limit` sounds a bit too abstract to me since this option doesn't affect explicit `backfill` command and it can affect even normal tasks that are just delayed to be executed.

How about `skip_delayed_tasks: 30s` or something?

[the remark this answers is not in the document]
osiire: `tasks` sounds redundant for me.
`skip_if_late: 30`
`delay_limit: 30`
`forbid_late_run: 30`
which do you like?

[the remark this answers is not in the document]
komamitsu: IMO, a redundant name is better than an ambiguous name, so I'm not sure names in your comment are better than `skip_delayed_tasks`.

@muga What do you think?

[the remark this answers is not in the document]
osiire: I agree with you. > "we should come up with a good name for those who don't have deep context on this ticket."

This new config option is for `schedule:` section, which already has `skip_on_overtime` option.
`skip` prefix is good idea hence both has same effects.
Besides, new option corresponds to int value that means how many seconds the schedule allow to late for. 
Therefore, `skip_if_late_for` is one of the good name I think.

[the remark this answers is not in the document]
muga: `skip` is fine in this context and "late for ..." is not used for time. Moreover the skipped target is different between `skip_on_overtime` and this option so, I think that it's better to include "tasks" work in the option name explicitly to avoid any confusing. It enables skipping **tasks** if they were delayed by time specified by users. How about `skip_tasks_delayed_by`??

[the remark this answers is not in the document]
osiire: I'm not sure what *tasks* means, but I agree with `skip_tasks_delayed_by`.  this name is clear for me to specify delay time argument.
If @komamitsu agree with, I'll push a commit for this change.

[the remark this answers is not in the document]
komamitsu: Ah, strictly speaking, `schedule` sounds better than `task` in this case because this option skips all tasks associated with delayed schedules. 

Also, this option works under `schedule:` option. So it might be clear w/o `schedule`. How about `skip_delayed_by`?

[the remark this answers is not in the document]
muga: Got it. and sorry to your confusing a bit.. `skip_deployed_by` sounds good to me.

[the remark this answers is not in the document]
osiire: I totally agree.

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleExecutor.java
```diff
@@ -199,8 +199,14 @@ private void runSchedule(ScheduleControl lockedSched)
 
             Config scheduleConfig = SchedulerManager.getScheduleConfig(def);
             boolean skipOnOvertime = scheduleConfig.get("skip_on_overtime", boolean.class, false);
+            Optional<Integer> skipDelay = scheduleConfig.getOptional("skip_delayed_by", Integer.class);
```
komamitsu: https://github.com/treasure-data/digdag/pull/752#discussion_r181556017

https://github.com/treasure-data/digdag/blob/1ed050a5a3c0708d85cde7c37dfef06e2094f4e7/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java#L79

With DurationParam.class not Integer.class, user can use a time unit like `30s` / `1m` as they want. If a specified figure is handled as only seconds, the option name needs to have `second` or something.

Could you try to use DurationParam.class?

osiire: Thank you for your advice.

komamitsu:
@osiire Thanks!

PR #751, opened 2018-04-07
Change workspace prefix not to use task name to avoid failing workspace creation

muga:
This PR changes workspace prefix. The name includes task name. If the task name is long, workspace name is also long. Digdag (0.9.24) server cannot create workspace by "File name too long". This error happens when I tried to reproduce https://github.com/treasure-data/digdag/issues/729.

In this PR, workspace prefix uses ~~workflow name~~ combination with projectId, workflowName, sessionId and attemptId instead of task name. Since the workspace name is internal design so, users might not affect this change directly.

```
2018-04-06 22:24:23.075 +0000 [ERROR] (0142@[1:muga_digdag_issues_729_ng]+muga_digdag_issues_729_ng+repeat^sub+loop-0+aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_aaa) io.digdag.core.agent.OperatorManager: Task failed with unexpected error: java.nio.file.FileSystemException: /tmp/digdag-tempdir6619218628117090050/workspace/+muga_digdag_issues_729_ng+repeat^sub+loop-0+aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_aaa_7247227962002122889: File name too long
[...]

Issue #729, opened 2018-02-05
A task is never finished when "File name too long" by log file occurs


toru-takahashi:
## Environment
digdag 0.9.24 (Server mode)

## Problem
When a digfile is pushed, the following validation works well.

```
Failed to store: {"message":"Validating project failed\nworkflow /tmp/digdag-tempdir5894056088905123786/push/new-project_1377544233838658341/new.dig Validating workflow failed\ntask name must not be longer than 255 characters: \"
```

On the other hands, the following dig file doesn't be validated. When this dig file run, the following error occurs and the task keeps to show "running" status.

```
+repeat:
  loop>: 10
  _do:
    +no_problem:
      sh>: echo "hello world"
    +aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa_aaa:
      if>: ${i != 0}
      _do:
        sh>: echo "${i}"
```

```
[root@ip-172-31-46-53 logs]# tail server.log
Caused by: java.nio.file.FileSystemException: /opt/digdag/logs/tasklogs/2018-02-05/0.1new@20180205T072606+0000/+new+repeat^sub+loop-0+aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa@5a78070f37baf7c0.3079@ip-172-31-46-53.ec2.internal.log.gz: File name too long
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:91)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
[...]

PR #757, opened 2018-04-10
Change temp file prefix used by BufferedRemoteTaskLogger

muga:
This PR fixes a bug that `BufferedRemoteTaskLogger` cannot upload log files on S3 correctly. The `BufferedRemoteTaskLogger` writes logs tmp files first before uploading them on S3. The tmp file prefix uses task name, if task name is too long, it cannot create tmp files. 

In this PR, the prefix is changed with combination with projectId, workflowName and sessionTime. sessionTime is formatted by ISO-8601 and used as string.

Reported on https://github.com/treasure-data/digdag/issues/729 and assumes #751.


muga:
Merged.

PR #760, opened 2018-04-11
fix(digdag-ui): fixed resolve version

Jerry-Hong:
https://github.com/shakacode/bootstrap-loader/issues/341


szyn:
I think that it might be a good idea to upgrade bootstrap-loader than adding a new dependency. 
This problem will be fixed in this PR.
https://github.com/shakacode/bootstrap-loader/pull/343

hiroyuki-sato:
Hello, @Jerry-Hong and @szyn 

Resolve 1.7.1 has been released. this version fixed resolve 1.7.0 issue. 
So we don't need change package anymore.

PR #765, opened 2018-04-20
Remove setPosixFilePermissions method call for Windows

muga:
This PR removes `Files.setPosixFilePermissions` method call for Windows. "digdag download" command doesn't work on Windows because Windows doesn't support POSIX pretty much. It fails when it calls `Files.setPosixFilePermissions` method.


frsyuki:
LGTM 👍

muga:
sounds good. thank you!

PR #770, opened 2018-04-25
[Fix] Group Task Retry including nested tasks does not update parent id of copied tasks.

nak-tera:
https://github.com/treasure-data/digdag/pull/738 added dependencies for copied tasks but didn’t do parent id.
So we add the implements that update parent id of copied tasks.
 
Without updating parent id, Group Retry unexpectedly works as follow.
Thanks.

# issue
Workflow Definition
group_retry.dig
```
+wf:
  _retry: 3

  +t1:
    echo>: foo

  +t2:
    +t2_nested:
      echo>: bar

  +t3:
    fail>: failure
```
Result

We expect task 384 has task 383 as a parent, but actually, the parent id is 379.
So task 384 belongs to task group that has task 379 as a parent.
As a result, task 384 and task 382 start processing parallelly.

# After Fixed

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -886,6 +886,7 @@ public class DatabaseSessionStoreManager
             for (StoredTask task : tasks) {
                 Task newTask = Task.taskBuilder()
                     .from(task)
+                    .parentId(task.getParentId().transform(it -> oldIdToNewId.containsKey(it) ? oldIdToNewId.get(it) : it))
                     .state(TaskStateCode.BLOCKED)
                     .stateFlags(TaskStateFlags.empty())
                     .build();
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -886,7 +886,7 @@ public class DatabaseSessionStoreManager
             for (StoredTask task : tasks) {
                 Task newTask = Task.taskBuilder()
                     .from(task)
-                    .parentId(task.getParentId().transform(it -> oldIdToNewId.containsKey(it) ? oldIdToNewId.get(it) : it))
+                    .parentId(task.getParentId().transform(it -> oldIdToNewId.getOrDefault(it, it)))
                     .state(TaskStateCode.BLOCKED)
                     .stateFlags(TaskStateFlags.empty())
                     .build();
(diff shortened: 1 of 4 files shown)


digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -886,6 +886,7 @@ public boolean copyInitialTasksForRetry(List<Long> recursiveChildrenIdList)
             for (StoredTask task : tasks) {
                 Task newTask = Task.taskBuilder()
                     .from(task)
+                    .parentId(task.getParentId().transform(it -> oldIdToNewId.containsKey(it) ? oldIdToNewId.get(it) : it))
```
komamitsu: How about simply `it -> oldIdToNewId.getOrDefault(it, it)`?

komamitsu:
Left a minor comment. Overall LGTM!

muga:
@nak-tera Thank you. LGTM too. Can you help to add tests to confirm if this correctly works or not?

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
nak-tera: It's better! I'll change it.

nak-tera:
@muga OK, I'll do it this week.

nak-tera:
@muga I added the tests. Can you check them?

digdag-core/src/test/java/io/digdag/core/workflow/TaskDependenciesTest.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    @After
+    public void destroy()
+            throws Exception {
+        embed.close();
+    }
+
+    @Test
+    public void TaskDependenciesTest()
```
muga: Do you have any reasons why first letter of the method name is not lowercase? How about `testTaskDependencies`?

digdag-core/src/test/java/io/digdag/core/workflow/TaskDependenciesTest.java
```diff
@@ -0,0 +1,208 @@
...
+            assertThat(tasks.get(2).getUpstreams(), is(Collections.singletonList(tasks.get(1).getId())));
+            assertThat(tasks.get(3).getUpstreams(), is(Collections.emptyList()));
+            assertThat(tasks.get(4).getUpstreams(), is(Collections.emptyList()));
+
+            return null;
+        });
+    }
+
+    @Test
+    public void GroupRetryTaskDependenciesTest()
```
muga: ditto

digdag-core/src/test/java/io/digdag/core/workflow/TaskDependenciesTest.java
nak-tera: Oh, I made a mistake...
I wil fix it.

nak-tera:
@muga Sorry for the late reply.
Thank you for your comments. I fixed it.

muga:
@nak-tera thank you! merging.. @komamitsu

PR #771, opened 2018-04-25
[doc] Fix store_last_results description in redshift

hiroyuki-sato:
`store_last_results` in redshift isn't BOOLEAN. 
The valid value is `first`. `all` and `false`.
I copied this fix from pg operator.
Should I add `false`?


hiroyuki-sato:
@muga Thank you for the review. Please take a look when you have time.

muga:
Thank you! Merged.

PR #774, opened 2018-05-03
Fix syntax error

yuokada:


muga:
@yuokada thank you! merging.

PR #775, opened 2018-05-08
[edit documentation] make wording clearer

chengvt:
Make wording clearer and match the argument name in the command reference documentation with the one in help file shown in the command line (shown below).
```
Usage: digdag backfill <project-name> <workflow-name>
```

PR #777, opened 2018-05-08
[doc] Use `name` instead of `workflow-name` in cli.

hiroyuki-sato:
Follow-up: #775

Use `name` instead of `workflow-name` in `digdag backfill` command line usage.
This change keeps consistency document and commands line usage.
#740 change it in the document. 
https://github.com/treasure-data/digdag/pull/740/files#diff-6da101123e994bcefae83a68f19bda25R558

I also found `workflow-name` in `digdag schedules` command usage.


muga:
Thank you! merging.

Issue #783, opened 2018-05-28
[feature request] support --task-log in local-mode.


yoyama:
It looks like that there is no way to save task log in local mode.
It is important to check result of tasks by check logs.
So I hope  get --task-log option in local mode same as server mode.

hiroyuki-sato:
@yoyama v0.9.32 introduced this feature. you can close this case.

PR #784, opened 2018-05-28
Add --task-log option in local mode

yoyama:
This patch supports --task-log option in local mode. (Issue #783)
I added LogModule to DigdagEmbed in class Run.

### DIFF
--- a/digdag-docs/src/command_reference.rst
+++ b/digdag-docs/src/command_reference.rst
@@ -106,6 +106,11 @@ Options:
 
   Example: ``--max-task-threads 5``
 
+:command:`-O, --task-log DIR`
+  Store task logs to this directory.
+
+  Example: ``--task-log digdag.log``
+
 :command:`-p, --param KEY=VALUE`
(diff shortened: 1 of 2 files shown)


digdag-docs/src/command_reference.rst
```diff
@@ -106,6 +106,11 @@ Options:
 
   Example: ``--max-task-threads 5``
 
+:command:`-O, --task-log DIR`
+  Store task logs to this directory.
+
+  Example: ``--task-log digdag.log``
```
muga: `digdag.log` looks like file name. we'd better to change it with directory-like name. for example, `log/tasks` or something. what do you think?

yoyama: @muga Thank you for your pointing out.
I agree with you and fix it.
If there are something other issues, please let me know.

Best regards,
  You Yamagata

yoyama:
Sorry for my late response.
I have already fixed the document.
Could you confirm it?

Best regards,
  You Yamagata

muga:
@yoyama Thank you for updating! Merging..

PR #789, opened 2018-06-07
Upgrade Gradle, SpotBugs, and JaCoCo to be compatible with JDK 10

frsyuki:
* Upgrade Gradle to 4.8
* FindBugs doesn't work with Java 10. Replace it to SpotBugs.
* Upgrade JaCoCo to 0.8.1.


wreulicke:
need to rewrite here?

https://github.com/treasure-data/digdag/blob/1d3c0aa09e0129ac7c416d554572d4e17f8b08e7/ci/run_findbugs.sh#L7

muga:
I merged https://github.com/treasure-data/digdag/pull/793 to upgrade Gradle 4.8 first so, will modify this change without Gradle 4.8 upgrading.

PR #790, opened 2018-06-07
Update JmxAgent to be compatible with JDK 11

frsyuki:
Since access to sun.management.Agent is prohibited since JDK 9, JmxAgent
class needs another compatible code. As a limitation with JDK 9, users
must set `jdk.attach.allowAttachSelf=true` system property to enable
JMX.

Behavior doesn't change with JDK 8.


muga:
We'd better to upgrade JDK11 for Digdag.

yoyama:
LGTM. But there are some conflicts. So I create another PR to follow up.

yoyama:
Follow-up PR #1342 include this PR has been merged. So close. Thanks!

PR #793, opened 2018-06-10
Update gradle wrapper

wreulicke:
This PR fixes followings.

* Update gradle wrapper
* Overwriting Gradle's built-in tasks is deprecated.
  here: https://docs.gradle.org/4.8/release-notes.html#overwriting-gradle's-built-in-tasks
* fix below warning for publishing.

```
As part of making the publishing plugins stable, the 'deferred configurable' behavior of the 'publishing {}' block is now deprecated. 
Please add 'enableFeaturePreview('STABLE_PUBLISHING')' to your settings file and do a test run by publishing to a local repository.
If all artifacts are published as expected, there is nothing else to do. 
If the published artifacts change unexpectedly, please see the migration guide for more details: https://docs.gradle.org/4.8/userguide/publishing_maven.html#publishing_maven:deferred_configuration. 
In Gradle 5.0 the flag will be removed and the new behavior will become the default.
```

* fix below warning for buildUi

```
Using TaskInputs.file() with something that doesn't resolve to a File object has been deprecated 
and is scheduled to be removed in Gradle 5.0.
 Use TaskInputs.files() instead.
```

## Note

* This PR has a deprecation yet. so, I will try to fix deprecations in another PR.

### need upgrade shadow plugin

```
A constructor for `org.gradle.api.internal.java.JavaLibrary` is used by Shadow plugin v1.2.x, and has been preserved for compatibility.
This has been deprecated and is scheduled to be removed in Gradle 5.0. 
If you're using the Shadow plugin, try upgrading to v2.x.
```


wreulicke:
Sorry, I missed following PR.
https://github.com/treasure-data/digdag/pull/789

PR #794, opened 2018-06-10
Update shadow plugin

wreulicke:
Update shadow plugin.

I found deprecations when I try to upgrade gradle.
PR: https://github.com/treasure-data/digdag/pull/793


muga:
@wreulicke thank you! merging. will check if it works on master.

PR #796, opened 2018-06-14
build.gradle: Clean digdag-ui build in 'clean' task

muga:
This PR makes minor change to remove digdag-ui build when 'clean' task is executed.


muga:
merging @frsyuki

PR #797, opened 2018-06-14
Update Node.js 8.x for digdag-ui and introduce NPM package locks

muga:
This PR tries to upgrade Node.js 8.11.3 for digdag-ui. We uses Node.js 7.x but, the version doesn't appear in release schedule list. 8.x is more active now. better to use 8.x. It updates dependencies as well.
ref: https://github.com/nodejs/Release#release-schedule

It also tries to introduce NPM package locks to mitigate any confusions related to digdag-ui build. 

assumes https://github.com/treasure-data/digdag/pull/796


muga:
Looks like digdag-ui works fine by Node.js 8.x. Merging

PR #816, opened 2018-06-28
[wip] Not restart DigdagServer for each test method

muga:
To make CI tests faster, this PR makes change not to restart DigdagServer for each test method.


yoyama:
@muga I will close this PR because it is too old and CI failed. If needed, please re-open it.

PR #818, opened 2018-06-29
Copyright year in a document use current year automatically.

hiroyuki-sato:
What do you think to update copyright year like the below?
(Format style may need to change.)
It uses current year automatically, so footer doesn't need to change every year.

digdag-footer


muga:
Thank you. looks nice to me :-) will merge it.

hiroyuki-sato:
Thank you!

PR #826, opened 2018-07-02
Modify sample of bq_load> operator

kamikaseda:
`fields:` is required in `schema:` parameter.
Reference: https://github.com/treasure-data/digdag/blob/d6d29a7161f69169055ea3b2e83f1d113292fcf7/digdag-tests/src/test/resources/acceptance/bigquery/load.dig


muga:
@kamikaseda Thank you. I will merge after CI jobs pass

PR #828, opened 2018-07-03
[Doc] Fix the following warnings.

hiroyuki-sato:
This PR fix the following warnings.

```
WARNING: Pygments lexer name 'yml' is not known
WARNING: Could not lex literal_block as "json". Highlighting skipped.
```

### DIFF
--- a/digdag-docs/src/operators/bq_load.md
+++ b/digdag-docs/src/operators/bq_load.md
@@ -232,20 +232,20 @@
 
   You can write schema within .dag file directly.
 
-  ```yml
+  ```yaml
   +step:
     bq_load>: gs://<bucket>/path/to_file
     ...
     schema:
       fields:
-        - name: "name",
+        - name: "name"
           type: "string"
         ...
   ```
 
   Or you can write it as external file.
 
-  ```json
+  ```java
   {
     "fields": [
       {"name": "name", "type": "STRING"},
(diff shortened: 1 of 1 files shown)


digdag-docs/src/operators/bq_load.md
```diff
@@ -232,37 +232,37 @@
...
-        - name: "name",
+        - name: "name"
           type: "string"
         ...
   ```
 
   Or you can write it as external file.
 
-  ```json
+  ```java
```
muga: As you might know, it seems like `Could not lex literal_block as "json"` is kind of syntax error because the content cannot be parsed by JSON parser. Included `,` or `...` like that.
ref: https://github.com/sphinx-doc/sphinx/issues/2413

We could keep `json` type if we would fix the JSON content. But, the current content also looks good to me. So, you think that we'd better to change type with `java` to make highlighting?

muga: Got it. I did compare highlights between `java` and `json` for this JSON content. I think that we'd better to use java highlighting and `...`. I will put LGTM on this PR. 


^ java highlighting


^ json highlighting

muga: > I'm not sure why but current digdag.io documents doesn't highlight codes. (It's configuration problem?)

Could you show me more concretely? I still getting your above question.

hiroyuki-sato: @muga Thank you for the comment. I misunderstood.

I thought digdag document exist code identifier like the below. And It's not highlighted.

    ```json

But those are no code identifier except this PR.

    ```

muga: Got it. thank you!

muga:
Thank you! merging

hiroyuki-sato:
Thank you!

PR #832, opened 2018-07-18
[wip] Change CommandExecutor SPI

muga:
This PR changes CommandExecutor SPI that are used by command based operators like `py`, `rb` and `sh`. The goal of this PR is that those operators could be executed on different instances from those tasks are executed by OperatorManager. The current SPI was designed supposing that a command task and the operator are executed on same instance. But, basically we don't need to suppose that.


muga:
I closed this PR to make cleanup. Will open another for this.

PR #850, opened 2018-08-27
[Experimental] Show logs by task

serihiro:
# What is this PR ?
- This PR is POC implementation of digdag-ui
- This PR modifies the format of task log in views so that logs are splitted for each tasks.

# Background of this PR
- Currently all task logs are unified, so it is difficult to distinguish that what task each line is  belongs to.

# e.g.

screen shot 2018-08-27 at 12 21 05

### DIFF
--- a/digdag-ui/console.jsx
+++ b/digdag-ui/console.jsx
@@ -1778,7 +1778,7 @@ class LogFileView extends React.Component {
       return (
         <div>
           <h3 id={'logs-' + this.props.file.taskName + this.props.order.toString()}
-              style={{display: 'block', paddingTop: '50px', marginTop: '-50px'}}>{this.props.file.taskName}</h3>
+              className='log-view'>{this.props.file.taskName}</h3>
           <pre>{pako.inflate(this.state.data, {to: 'string'})}</pre>
         </div>
       )
--- a/digdag-ui/console.jsx
+++ b/digdag-ui/console.jsx
@@ -1352,8 +1352,16 @@ function sortTasksForTreeView (tasks: Array<Task>): Array<Task> {
   // First, divide tasks into rootTasks and taskGroups.
   const rootTasks: Array<Task> = []
   const taskGroups: Map<string, Array<Task>> = new Map()  // {parentId => Array<Task>}
+  const taskOrders: Map<string, Number> = new Map()
   tasks.forEach(t => {
     const parentId: ?string = t.parentId
+    if (taskOrders.get(t.fullName) == null) {
+      taskOrders.set(t.fullName, 1)
+    } else {
+      taskOrders.set(t.fullName, taskOrders.get(t.fullName) + 1)
+    }
+    t.order = taskOrders.get(t.fullName)
+
     if (parentId != null) {
       const group: ?Array<Task> = taskGroups.get(parentId)
       if (group != null) {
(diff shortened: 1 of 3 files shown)


muga:
@serihiro I like this idea that you're suggesting. Because I actually faced similar problem that it was not easy to find a specific task log messages in the unified log messages. it would be helpful for my case.

This change looks good to me basically. This is not trivial but, if we could introduce the link to each task log in the task in "Tasks" section, it would be great. Users first check task failures in "Tasks" section and may want to see those task logs on "Logs" section. We'd better to introduce association or something between "Tasks" and "Logs" sections I think. Please take a look if you get a chance.

serihiro:
@muga 
I added hash link like this. ( c9e0a3f2de029b55e809666bf2b121ab255686e8 )
When user clicks a taskId then jump to the LogFileView of the clicked taskId. How about this?

digdag-ui-

hiroyuki-sato:
Hello, @serihiro 

This is a comment from a user. 

This change seems totally good. 
My favorite style is a collapse and expands view.
If a user clicks taskId link, It expands ( or collapse) the log view.

serihiro:
Hello @hiroyuki-sato . 

Thank you for your comment !

Your suggestion sounds good. But, we think that there is few differences between current implementation and your idea, in a point of view achieving the purpose(being easy to read each task log).

So, experimentally we plan to introduce the current implementation. Of course, good opinions are always welcome, and we can keep improving digdag-ui. If you have more good idea, let's discuss with issue or PR.

muga:
@hiroyuki-sato If you could make sharing the comment from the user who we are missing, it would be great to improve UI. 

@serihiro LGTM. Please let me know if you still have any change and improvement that you're expecting before merging this PR into master?

serihiro:
@muga 
Thank you for your review! I do not have any remaining tasks about this PR.

muga:
@dzannotti It would be great if you can take a look this PR.

digdag-ui/console.jsx
```diff
@@ -1763,9 +1774,17 @@ class LogFileView extends React.Component {
...
 
   render () {
-    return this.state.data
-      ? <span>{pako.inflate(this.state.data, {to: 'string'})}</span>
-      : null
+    if(this.state.data){
+      return (
+        <div>
+          <h3 id={'logs-' + this.props.file.taskName + this.props.order.toString()}
+              style={{display: 'block', paddingTop: '50px', marginTop: '-50px'}}>{this.props.file.taskName}</h3>
```
dzannotti: this shouldn't be an inline `style` should be part of style.less

digdag-ui/console.jsx
```diff
@@ -1763,9 +1774,17 @@ class LogFileView extends React.Component {
...
+          <pre>{pako.inflate(this.state.data, {to: 'string'})}</pre>
+        </div>
+      )
+    }else{
```
dzannotti: nit: this doesn't pass linting rules

digdag-ui/console.jsx
```diff
@@ -1796,6 +1815,23 @@ class AttemptLogsView extends React.Component {
     this.fetch()
   }
 
+  logFiles () {
+    if (!this.state.files.length) {
```
dzannotti: I think this would read better if destructured `const { files } = this.state`

muga:
@dzannotti thank you for reviewing! let you know again once fixing them with @serihiro

serihiro:
@dzannotti 
Thank you for your reviewing and comments!
I fixed by your comments. Could you check again ?

dzannotti:
👍 LGTM

serihiro:
@dzannotti 
Thank you for your reviewing !

@muga 
I fixed all issues. Could you take care this PR ?

muga:
@serihiro @dzannotti Thank you! I will merge once CI will pass.

Issue #709, opened 2018-01-04
[Q] How can I set retry wait interval of _retry parameter?


jo8937:
I want to retry after 1 hours.
How can I set this?
If this feature is not included, would it be possible to add in next release?

hiroyuki-sato:
Hello, @jo8937 

Do you want to specify retry interval in Operator or TaskGroup?, I think It's impossible yet. 
The current implementation, a task will retry immediately until `_retry` count reach. 

Please write your use case or create PR. 

* [Twitter](https://twitter.com/seri_k/status/949252318268030976)  Japanese text.

jo8937:
That's what I want!
but I see, it is not impossible yet. I'll try to see source code for PR.

The twitter thread very useful to know internal structure of digdag retry. Thank you!

hiroyuki-sato:
Hello, @jo8937  `exponential` backoff introduced in v0.9.30
Could you try it?  If it works well, please close this case.
http://docs.digdag.io/workflow_definition.html#retrying-failed-tasks-automatically

jo8937:
@hiroyuki-sato LGTM! Thank you. It is better than what I want. I've tried it and it works fine. I'll close this issue.

PR #852, opened 2018-08-29
Retry interval #709

yoyama:
I would like to implement retry interval which was discussed in issue #709.
To implement,  I changed _retry parameter as follows.

``_retry: num interval constant|exponential``

interval is seconds to wait for. (default 0)
constant or exponential set interval increments.(default constant)
if constant is set,  interval is fixed.
if exponential is set, interval increments with interval * 2^ retry_count.

Examples

1.  ``_retry: 3``

Same as original behavior. Retry 3 times without interval.

2. ``_retry: 3 10``  or
``_retry: 3 10 constant``

Wait for 10 seconds for each retry

3. ``_retry: 3 10 exponential``

The retry intervals are  10, 20, 40 for each retry.


muga:
Sorry to be late. This is good feature. I will take a look.

muga:
The basic logic looks good. What do you think that retry params could be represented as Json object like following? Because the string representation is confused by me. 
```
_retry:
  limit: 3
  interval: 10
  interval_type: exponential
```
If the value is integer instead of Json object, the value will be treated as retry_limit as usual. 
```
_retry: 3
```
What do you think?

yoyama:
@muga Thank you for your comment.

At first I also tried to implement with JSON. 
But I noticed that I needed to register the new key names to WorkflowCompiler.GROUPING_TASK_CONFIG_KEYS.
(Sorry it may be my misunderstanding because it was so long ago)

Is it ok to increase keys for validation? 
If so, I would like to change the implementation to JSON.

Best regards,
  You Yamagata

muga:
@yoyama got your point and agree with you. I also don't want to register new key name. What do you think about the following idea prototyping roughly, not care of interval things there? to keep backward compatibility of the behavior of _retry, we'd better to keep evaluating _retry value as integer if the value is integer. on the other hand, we can check if the value is json object type or not. 

```java
diff --git a/digdag-plugin-utils/src/main/java/io/digdag/util/RetryControl.java b/digdag-plugin-utils/src/main/java/io/digdag/util/RetryControl.java
index cc777a91c2..d0cccde4ca 100644
--- a/digdag-plugin-utils/src/main/java/io/digdag/util/RetryControl.java
+++ b/digdag-plugin-utils/src/main/java/io/digdag/util/RetryControl.java
@@ -1,5 +1,6 @@
 package io.digdag.util;

+import com.fasterxml.jackson.databind.JsonNode;
 //import io.digdag.core.workflow.TaskConfig;

@@ -17,7 +18,20 @@ public class RetryControl
     private RetryControl(Config config, Config stateParams, boolean enableByDefault)
     {
         this.stateParams = stateParams;
-        this.retryLimit = config.get("_retry", int.class, enableByDefault ? 3 : 0);
+        final JsonNode retry = config.getInternalObjectNode().get("_retry");
+        if (retry != null && (retry.isNumber() || retry.isObject())) {
+            if (retry.isNumber()) {
+                this.retryLimit = retry.intValue();
+            }
+            else { // object
+                this.retryLimit = retry.has("limit") ?
+                        retry.get("limit").intValue() :
+                        (enableByDefault ? 3 : 0);
+            }
+        }
+        else {
+            this.retryLimit = enableByDefault ? 3 : 0;
+        }
         this.retryCount = stateParams.get("retry_count", int.class, 0);
     }
```

yoyama:
@muga Thank you for your comment.
I agree with your idea. It is very excellent!
I will change the implementation with your idea.

Best regards,
  You Yamagata

muga:
@yoyama thank you. got it. please let me know if ready to take a look.

yoyama:
@muga Thank you for your comment.  I changed the code according to our discussion.
But CI failed.
I don't know the cause of issue. Please wait a moment until fixing it.

Best regards,
  You Yamagata

digdag-plugin-utils/src/main/java/io/digdag/util/RetryControl.java
```diff
@@ -10,21 +12,80 @@ public static RetryControl prepare(Config config, Config stateParams, boolean en
...
+        private RetryIntervalType(final String type) {
+            this.type = type;
+        }
+        public static RetryIntervalType find(String value) {
+            for (RetryIntervalType r : RetryIntervalType.values()) {
+                if (r.type.equalsIgnoreCase(value)) {
+                    return r;
+                }
+            }
+            throw new ConfigException("Invalid retry_interval_type");
```
muga: I think that we'd better to show actual `value` value as the message like this. what do you think?
```
"Invalid retry_interval_type "  + value
```

digdag-plugin-utils/src/main/java/io/digdag/util/RetryControl.java
```diff
@@ -10,21 +12,80 @@ public static RetryControl prepare(Config config, Config stateParams, boolean en
...
+                    this.retryInterval = 0;
+                }
+                if(retry.has("interval_type")){
+                    this.retryIntervalType = RetryIntervalType.find(retry.get("interval_type").textValue());
+                } else {
+                    this.retryIntervalType = RetryIntervalType.CONSTATNT;
+                }
+            } else { // Unknown format
+                throw new ConfigException("Invalid _retry format");
+            }
```
muga: Please fix several coding style according to ours. 
```
  }
  else if (... ...) {
    ... ...
  }
  else {
    ... ...
  }
```

digdag-plugin-utils/src/main/java/io/digdag/util/RetryControl.java
yoyama: @muga Thank you for your pointing out. I fixed it.

yoyama:
@muga Thank you so much for your reviewing in detail. 
I fixed both coding style issue and adding information to error message.
If there are something other issues, please let me know.

Best regards,
  You Yamagata

muga:
@yoyama looks good! thank you for improving. merging.

PR #853, opened 2018-08-30
[POC] Add fetch job information operator

serihiro:
# What is this PR ?
- This pr add `td_job_show>` operator (the name is a temporary name)
- `td_job_show>` operator gets `jobId` as an argument and fetch the job information of the jobId, and stores parameter `${td.stored_jobs[jobId]}` (this is a temporary specification)

# Example

```yaml
_export:
  td:
    database: serizawa_test

+job_show1:
  td_job_show>: 8911629

+echo_fetched_job_num_records:
  echo>: "${td.stored_jobs['8911629'].num_records}"
```

# TODO
- [x] poc implementation
- [ ] fix specification
- [ ] add test
- [ ] add document

### DIFF
--- a/digdag-standards/build.gradle
+++ b/digdag-standards/build.gradle
@@ -15,7 +15,7 @@ dependencies {
     compile 'com.sun.mail:javax.mail:1.5.6'   // 'com.sun.mail:smtp' doesn't work because enabling mail.debug property throws java.lang.NoClassDefFoundError: com/sun/mail/util/MailLogger
 
     // td
-    compile 'com.treasuredata.client:td-client:0.8.5'
+    compile 'com.treasuredata.client:td-client:0.8.6'
     compile 'org.msgpack:msgpack-core:0.8.11'
     compile 'org.yaml:snakeyaml:1.17'
 
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/OperatorModule.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/OperatorModule.java
@@ -14,6 +14,7 @@ import io.digdag.standards.operator.redshift.RedshiftUnloadOperatorFactory;
 import io.digdag.standards.operator.td.TDResultExportOperatorFactory;
 import io.digdag.standards.operator.td.TdDdlOperatorFactory;
 import io.digdag.standards.operator.td.TdForEachOperatorFactory;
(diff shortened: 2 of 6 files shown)


muga:
@serihiro `${td.stored_jobs[jobId]}` is interesting. I'm curious about the latency of task pulling when stored_jobs object will be large. So, I want to ask you about the following:

- How many jobs will be stored by this operator?
- Then, how size of the stored_jobs object you expect?

`query` and `result_schema` fields could be large. for example, `query` is sometimes built and generated by tools like 5MB or 10MB.

serihiro:
Or, adding `max_available_store_jobs_size` settings to digdag and adopt LIFO to the store 🤔

serihiro:
@muga 
I did an experiment to confirm how much memory is consumed by `stored_jobs`.
I measured `Config` object size in `TdJobShowOperatorFactory` before execution of storing job into `StoreParams` by [JOL](http://openjdk.java.net/projects/code-tools/jol/)

```diff
@@ -59,6 +60,7 @@ public class TdJobShowOperatorFactory
             this.systemDefaultConfig = TDOperator.systemDefaultConfig(systemConfig);
             this.jobId = params.get("_command", String.class);
             this.storeKey = params.getOptional("store_key", String.class);
+            System.out.println("config size:" + GraphLayout.parseInstance(request.getConfig()).totalSize());
         }
```

I tried to run a workflow which contains 5 `td_job_show>:` operators.
And I found that Config object size is increasing by 7KB-10KB for each execution. So this increation is a size of a job stored by `td_job_show`. So when 100 jobs are stored this operator consumes up to about 1MB memory. I think it does not cause a serious problem.

But as you said, `query` and `result_schema` could be large size, so it may affect a latency to serialize `Config` object. 

So I think `query` and `result_schema` should be excluded. On the other hand, the size of other items(`num_records` , `types` , `status` etc) is almost static. So the total size just scales by linearly.

|config size|diff with before execution|
|--|--|
|9163264 | -|
|9531472 | 368208|
|9538648 | 7176|
|9549328 | 10680|
|9558640 | 9312|

muga:
@serihiro Thank you for taking the PoC. It might not easy for us to make decision about how we handle `query` and `result_schema` parameters now. About the rest of parameters, we could fix `td` operator and handle them like https://github.com/treasure-data/digdag/pull/860

I close this ticket. will reopen if need.

PR #860, opened 2018-09-07
Add new td attribute

serihiro:
# What is this PR?
- This PR add new attributes `num_records` on StoreParams as `${td.last_job.num_records}` 

# Background
- Currently `${td.last_job_id}` is set to StoreParams at `BaseTdJobOperator.runTask` 

https://github.com/treasure-data/digdag/blob/c67fcb263094fec45f12b054434e08cce985c97a/digdag-standards/src/main/java/io/digdag/standards/operator/td/BaseTdJobOperator.java#L61-L64

- In some use-cases, it is helpful if more job attributes is set to StoreParams in the same timing
- But `TDJob` object, which has the job information, may become very large size in total because of `query` or `result_schema`, and consume much memory.
- So we decided to add only `num_records` which is thought to be the most useful for now

### DIFF
--- a/digdag-standards/build.gradle
+++ b/digdag-standards/build.gradle
@@ -15,7 +15,7 @@ dependencies {
     compile 'com.sun.mail:javax.mail:1.5.6'   // 'com.sun.mail:smtp' doesn't work because enabling mail.debug property throws java.lang.NoClassDefFoundError: com/sun/mail/util/MailLogger
 
     // td
-    compile 'com.treasuredata.client:td-client:0.8.5'
+    compile 'com.treasuredata.client:td-client:0.8.6'
     compile 'org.msgpack:msgpack-core:0.8.11'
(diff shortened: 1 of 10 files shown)


digdag-standards/build.gradle
```diff
@@ -15,7 +15,7 @@ dependencies {
     compile 'com.sun.mail:javax.mail:1.5.6'   // 'com.sun.mail:smtp' doesn't work because enabling mail.debug property throws java.lang.NoClassDefFoundError: com/sun/mail/util/MailLogger
 
     // td
-    compile 'com.treasuredata.client:td-client:0.8.5'
+    compile 'com.treasuredata.client:td-client:0.8.6'
```
serihiro: This update is needed to access the attribute `TDJob.numRecords`

muga: 👍

PR #843, opened 2018-08-17
[WIP] Implement td result export operator

serihiro:
# About this PR
- This PR implements `td_result_export` operator

# TODO
- [x] Bump up [td-client-java](https://github.com/treasure-data/td-client-java/pull/107)
- [x] Solve independency conflicts
- [x] Implement `TDResultExportOperatorFactory`
- [ ] Add Integration Test for `td_result_export`

PR #865, opened 2018-09-18
Downgrade okhttp 3.9.0 to 3.4.1

muga:
CI build have been failing after okhttp was upgraded. This PR downgrades the version to 3.4.1. We also will investigate why for future okhttp upgrade.


muga:
The cause of CI failing may be that td-client was upgraded from 0.7 to 0.8 on https://github.com/treasure-data/digdag/pull/843. As we will investigate why td-client causes the failing, I will close this PR.

Issue #857, opened 2018-09-05
digdag sessions/attempts over 100 entries.


hiroyuki-sato:
Related #830

A user wants to see a status (`digdag attempts` and `digdag sessions`) over 100 entries.
But the limitation is hardcoded.
https://github.com/treasure-data/digdag/blob/master/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java#L415

Ref (Japanese text)
* [Twitter1](https://twitter.com/tkmoteki/status/1036991082662252544)
* [Twitter2](https://twitter.com/tkmoteki/status/1036990610744401925)

muga:
#880 was merged and will be released as v0.9.31. https://github.com/treasure-data/digdag/milestone/10

hiroyuki-sato:
Thank you!

PR #880, opened 2018-09-25
Make page_size configurable

serihiro:
# What is this PR ?
- Make `GET /api/attempts` and `GET /api/sessions` enable to take `page_size` parameter and change the number of records in response
- But this PR does not change the current behavior(if `page_size` is not specified, use `100` as a default value)
- Set internal MAX_PAGE_SIZE(up to 1000)
- This feature is the implementation of #857 

# TODO
- [x] Introduce Validator
- [x] Support `page_size` parameter at `GET /api/attempts`
- [x] Support `page_size` parameter at `GET /api/sessions`
- [x] Support `page_size` option at `digdag` cli

### DIFF
--- a/digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
+++ b/digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
@@ -85,10 +85,10 @@ public class AttemptResource
             @QueryParam("workflow") String wfName,
             @QueryParam("include_retried") boolean includeRetried,
             @QueryParam("last_id") Long lastId,
-            @QueryParam("page_size") int pageSize)
+            @QueryParam("page_size") Integer pageSize)
             throws ResourceNotFoundException
     {
-        int validPageSize = QueryParamValidator.validatePageSize(pageSize, MAX_ATTEMPTS_PAGE_SIZE, DEFAULT_ATTEMPTS_PAGE_SIZE);
+        int validPageSize = QueryParamValidator.validatePageSize(Optional.fromNullable(pageSize), MAX_ATTEMPTS_PAGE_SIZE, DEFAULT_ATTEMPTS_PAGE_SIZE);
 
         return tm.begin(() -> {
             List<StoredSessionAttemptWithSession> attempts;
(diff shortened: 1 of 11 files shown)


digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
```diff
@@ -157,6 +157,8 @@
     private static final Logger logger = LoggerFactory.getLogger(ProjectResource.class);
     private static final int ARCHIVE_TOTAL_SIZE_LIMIT = 2 * 1024 * 1024;
     private static final int ARCHIVE_FILE_SIZE_LIMIT = ARCHIVE_TOTAL_SIZE_LIMIT;
+    private static final int MAX_SESSIONS_PAGE_SIZE = 1000;
```
muga: ditto

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
```diff
@@ -53,6 +53,8 @@
     private final AttemptBuilder attemptBuilder;
     private final WorkflowExecutor executor;
     private final ConfigFactory cf;
+    private static final int MAX_ATTEMPTS_PAGE_SIZE = 1000;
```
muga: What do you think that we'd better to extract the max page size as system config? The constant value may fit with someone. may not fit with others. For example, in my application, I want to keep using `100` as page size and max page size. But I believe that `1,000` looks reasonable to other users'.

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
serihiro: Thank you for your reviewing!
Yes, I think it makes sense. I will try modifying to make `max page size` configurable by SystemConfig.

muga: Thank you!

serihiro:
@muga 
I added 3 commits.
I created new keys `api.max_attempts_page_size` and `api.max_sessions_page_size` in order to make configurable max page size. If you have a better idea of naming, please let me know it.

Could you review this ? 🙇 

- [x] Refactor QueryParamValidator  defc46e
- [x] Make MAX_PAGE_SIZE configureable 2862d51
- [x] Update documents 1f9de8a

PR #914, opened 2018-10-31
Added the  gradle plugin to make IntelliJ able to process the annotations

pdelagrave:
To get IntelliJ parsing the annotations to generate the necessary classes and make the code readable/browsable.
https://plugins.gradle.org/plugin/net.ltgt.apt-idea

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -3,6 +3,7 @@ plugins {
     id 'com.jfrog.bintray' version '1.8.1'
     id 'com.github.kt3k.coveralls' version '2.6.3'
     id 'com.github.spotbugs' version '1.6.2'
+    id 'net.ltgt.apt-idea' version '0.19'
 }
 
 apply plugin: 'com.github.johnrengelman.shadow'
@@ -30,6 +31,7 @@ allprojects {
     apply plugin: 'jacoco'
     apply plugin: 'com.jfrog.bintray'
     apply plugin: 'maven-publish'
+    apply plugin: 'net.ltgt.apt-idea'
 
     repositories {
         mavenCentral()


komamitsu:
@pdelagrave Thanks for your contribution!

BTW, you can build all the Digdag projects including `digdag-client` on IntelliJ with this change, right?

pdelagrave:
@komamitsu yes, you can build the whole project on IntelliJ/gradle with or without this PR.

Digdag uses https://immutables.github.io to generate immutable versions of classes. The immutable classes are generated automatically at compile time when the immutable library jar is in the classpath.

This is using java's [APT](https://docs.oracle.com/javase/7/docs/technotes/guides/apt/GettingStarted.html#what-is-apt)  feature/tool. Because it's a compile-time only thing, IntelliJ isn't seeing the immutable classes and this causes lot's of "cannot resolve symbol" errors when browsing the code.

The PR adds a gradle plugin for IntelliJ so it can scan the dependencies using APT before compile time, and add potentially generated classes to its index.

**side note**: it seems like the `continuous-integration/appveyor/pr` failure isn't related to the PR, is this a known flake?

hiroyuki-sato:
Hello, @pdelagrave 
Thank you for your contribution!.

I can't build Digdag with your PR in my IntelliJ environment.
Could you tell me what is wrong?
Did you install any plugin for this PR?

The attached image says "cannot resolve symbol ImmutableRestProject"



* clone https://github.com/pdelagrave/digdag
* checkout `apt-idea-plugin` Branch
* Open project in IntelliJ
* Import Gradle project
* `Build` -> `Build Project`

Environment 

* macOS: 10.13.6
* Java: 1.8.0_192-b12
* IntellJ: 2018.2.5

Best regards.


```
% git clone https://github.com/pdelagrave/digdag
Cloning into 'digdag'...
remote: Enumerating objects: 17, done.
remote: Counting objects: 100% (17/17), done.
remote: Compressing objects: 100% (16/16), done.
remote: Total 263230 (delta 1), reused 7 (delta 0), pack-reused 263213
Receiving objects: 100% (263230/263230), 77.28 MiB | 470.00 KiB/s, done.
Resolving deltas: 100% (224330/224330), done.
```

```
% cd digdag
% git checkout -b apt-idea-plugin origin/apt-idea-plugin
Branch 'apt-idea-plugin' set up to track remote branch 'apt-idea-plugin' from 'origin'.
Switched to a new branch 'apt-idea-plugin'
```

pdelagrave:
Hi @hiroyuki-sato 

There's 2 different independent issues at play here:
1. Building the project from inside IntelliJ
2. Having IntelliJ to find generated classes

For the first issue, I finally figured out that there's an important configuration option to be enabled to fully have IntelliJ be fully integrated with an existing gradle build configuration: `Delegate IDE build/run actions to gradle` needs to be enabled. This isn't presented to the user when first importing a gradle project to IntelliJ.
image
Enabling this will make it possible for the user to use the Build/Build project menu in IntelliJ and have gradle actually do the build, which will succeed in the same way as if done in a separate terminal using `./gradlew cli` for example.

---

For the second issue, it doesn't matter if the project was built by IntelliJ calling gradle or by the user calling gradle in a terminal.
The only way I could find to have IntelliJ resolve the generated Immutable classes was to use the `apt-idea` gradle plugin.

Once `apt-idea` added in the gradle config and IntelliJ having reloaded this gradle config, it's a matter of building the project at least once so IntelliJ can pick up and resolve the generated classes. It doesn't matter if the build was done from inside IntelliJ or from the terminal, as long as the build was done by gradle (or any tools that'd put the generated classes in a directory where the `apt-idea` plugin expect them to be)

I've tried many things with the IntelliJ built-in annotation processor but to no avail.
I haven't tried the suggested solution that was recently appended at the end of the Digdag's readme last week, as I found it easier not to involve maven and use a plugin dedicated to solve this very problem.

Let me know if you can get it working with this update.

Thanks

Pierre

hiroyuki-sato:
Hello, @pdelagrave
It seems to work well. Thank you!

PR #927, opened 2018-12-04
Add sessions status filter

tosametal:
Add `StatusFilter` component to filter sessions by status.

The below is a sample screen.

### DIFF
--- a/digdag-ui/console.jsx
+++ b/digdag-ui/console.jsx
@@ -643,21 +643,34 @@ class ProjectsView extends React.Component {
 }
 
 class StatusFilter extends React.Component {
+  static Status = class Status {
+    static ALL = "All";
+    static SUCCESS = "Success";
+    static FAILURE = "Failure";
+    static PENDING = "Pending";
+    static CANCELED = "Canceled";
+    static CANCELING = "Canceling";
+
+    static allStatus () {
+      return [Status.ALL, Status.SUCCESS, Status.FAILURE, Status.PENDING, Status.CANCELED, Status.CANCELING]
+    }
+  };
+
   state = {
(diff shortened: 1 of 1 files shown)


muga:
@tosametal Thanks! That's great. I will take a look.

muga:
@tosametal the function and the impl logic looks good to me. one thing is that we may need to explain what is the drop-down filter. Can we put `Status: ` message on the left of the drop-down button? What do you think? 

@dzannotti if you get a chance, can I consult you about the look and feel of the change? 

I also think that we'd better to support pagination buttons or automatic scrolling for page loading but, that would be future improvement.

tosametal:
@muga 
Thank you for reviewing.
I agree with your suggestion. I will add `Status: ` message.
Also, since I noticed that status name `Failed` should be `Failure`, I will fix it.

> I also think that we'd better to support pagination buttons or automatic scrolling for page loading but, that would be future improvement.

If I have a chance, I will try to send PR about this improvement!

szyn:
@dzannotti PTAL? 🙇

digdag-ui/console.jsx
```diff
@@ -642,6 +642,48 @@ class ProjectsView extends React.Component {
...
 }
 
+class StatusFilter extends React.Component {
+  state = {
+    selectedStatus: 'All'
+  };
+
+  filterSessionsByStatus(sessions, selectedStatus) {
+    switch (selectedStatus) {
+      case 'Success':
```
dzannotti: this cases should use a common constant with the rendering function, so strings can't go out of sync (also we should add a type so flow can ensure consitency)

PR #936, opened 2019-01-09
Introduce AccessController

muga:
This PR introduces access control manager to check permissions that users can access to resources. The controller is injected in all endpoints provided by digdag-server. If the controller recognizes that users don't have enough permissions to access, it returns 403 Forbidden.


frsyuki:
@muga Design looks good. Let's add tests and merge

muga:
Thanks. I will test it.

doctor:
Is this feature now available for us to use? - to disable Edit option in the UI

PR #954, opened 2019-01-28
Change workflowName in ifOperatorDelayedEvalElseDo

yoyama:
Sometimes ifOperatorDelayedEvalDo failed with the message "Session already exists".
The reason is ifOperatorDelayedEvalElseDo and ifOperatorDelayedEvalDo use same workflowName.
To avoid it workflowName used in ifOperatorDelayedEvalElseDo changed.


muga:
Looks good. Let's merge after CI passed.

PR #956, opened 2019-01-28
[doc]. Change IntelliJ build way

hiroyuki-sato:
I copied https://github.com/treasure-data/digdag/pull/914#issuecomment-437123569

I think, this step better than `Maven` way.


muga:
Thanks. It looks good. Will merge it.

hiroyuki-sato:
Thank you!

PR #960, opened 2019-01-30
[Doc] Use inline literal

hiroyuki-sato:
`--task-log`, `--enable-swagger` and `attempt_id` use inline literal.


yoyama:
Thank you for your fix.

PR #962, opened 2019-02-01
Update documentation link from HTTP to HTTPS

szyn:


szyn:
@hiroyuki-sato Hi, thank you for your pointed out. 
Nice catch 😊 I overlooked that... I updated footer link as well 7bd2f9d

PR #964, opened 2019-02-01
Clarification in the usage text

ehaupt:
The -P run parameter also accepts parameters from a JSON file. This patch clarifies the usage text of the run sub-command.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/Run.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/Run.java
@@ -196,7 +196,7 @@ public class Run
         err.println("    -o, --save DIR                   uses this directory to read and write status files (default: .digdag/status)");
         err.println("        --no-save                    doesn't save status files at .digdag/status");
         err.println("    -p, --param KEY=VALUE            overwrites a parameter (use multiple times to set many parameters)");
-        err.println("    -P, --params-file < PATH.yml | PATH.json >  reads parameters from a YAML/JSON file");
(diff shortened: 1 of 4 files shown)


muga:
Hi @ehaupt You're right. As you know, since yaml is basically superset of JSON so, yaml parser that we've used in Digdag can read JSON files. What do you think that we have special reasons why we'd better explicitly to describe about JSON files as well?

ehaupt:
Hi @muga Thank you for getting back so quickly. I was recently working on a workflow that went into production. In the documentation I specified a sample showing a YAML data structure to be passed to the workflow. The integration team contacted me and asked if it wouldn't be possible to supply JSON structures as it would be more convenient for them. I realized that this would be no problem with the internal parser digdag uses but conceded that this might be not obvious for all the users. The additional mention of JSON might be indeed superfluous but might help some users. I'm happy with or without the addition.

muga:
@ehaupt Thank you for commenting about the behind of the PR! Got it and it makes sense I think.

May I ask you to move your change from source code for commands to http://docs.digdag.io/command_reference.html? It'd better place where we mention about it instead of each source code. The HTML is generated with `digdag-docs/src/command_reference.rst`. What do you think?

ehaupt:
@muga Thanks again for your feedback. Great idea! I've reverted the commits and added a simple mention in the rst file.

muga:
@ehaupt Thank you for fixing! We'd better to fix other parts in the doc but the fix itself is good and will fix others later.

PR #967, opened 2019-02-05
fix can't refer to directories outside the project

hisahisa:
Hello
I am using digdag in own project.
I use the docker function, it can't see the folder of hostOS from the docker container, so I fixed it. I added volume_host and volume_guest tags to the same layer of the image tag. 

usage
```
+ step_create_nantoka:
  _export:
    docker:
      image: $ {docker_python_image_name}
      volume_host: /Users/hisashi/IdeaProjects/sftp_data
      volume_guest: /home/alpine/x_folder/
  py>: tasks.test_yogo1.xyz
```

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/DockerCommandExecutor.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/DockerCommandExecutor.java
@@ -87,6 +87,13 @@ public class DockerCommandExecutor
             command.add("-v").add(String.format(ENGLISH,
                         "%s:%s:rw", projectPath, projectPath));  // use projectPath to keep pb.directory() valid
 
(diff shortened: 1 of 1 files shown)


muga:
Hi @hisahisa, thank you for suggesting! I'm considering about the behavior of server mode and it looks like it may not be easy to setup or use the options that you're proposing on digdag-server cluster. because the options need to assume same setting on multiple digdag-server instances. I'm now guessing that your Digdag environment is local mode or single digdag-server instance by server mode. 

Our Docker command that is built by DockerCommandExecutor already has volume mount option for project path between host and guest so, your Python script can see data under project path. What do you think? Is it not enough for your use case?

muga:
Thank you for letting me know. I close this ticket but please feel free to reopen it if need.

Issue #944, opened 2019-01-22
[issue] python digdag.env.export seems not working (2)


sonots-zozo:
Reopen the issue https://github.com/treasure-data/digdag/issues/170

The example tells how to use `digdag.env.export`, but it seems not working because the `+dependent1` task is not a child of `+set_export` task.
https://github.com/treasure-data/digdag/blob/19dfc0c00e980a6d2c7dfb0f65cc66d62ac63268/examples/params.dig#L47-L56

The result was as:

```
2019-01-22 16:15:07 +0900 [INFO] (0017@[0:default]+params+export_params_2+set_export): py>: examples.params.set_export
2019-01-22 16:15:07 +0900 [INFO] (0017@[0:default]+params+export_params_2+dependent1): py>: examples.params.show_export
show_export: table = None
```

although it aims to set `table = carried information`.

How can we use `digdag.env.export`, then? In my understanding, following .dig is also prohibited.

```
timezone: UTC

+export_params:
  py>: params.set_export

  +dependent1:
    py>: params.show_export
```

yoyama:
How about using **digdag.env.store** instead of **digdag.env.export** ?

**params.py**
```python
from __future__ import print_function
import digdag

def set_export():
    digdag.env.export({"table": "carried information"})

def set_store():
    digdag.env.store({"table": "carried information"})

def show_export(table=None):
    print("show_export: table = " + str(table))

```
**show1.sh**
```
#!/bin/bash                                                                                                                  

echo "table = $table"
```

**test1.dig**
```
timezone: UTC

+export_params:
  py>: params.set_store

+task1:
  py>: params.show_export

+task2:
  sh>: ./show1.sh

```

Results:
```
2019-01-22 17:02:28 +0900 [INFO] (0017@[0:default]+test1+export_params): py>: params.set_store
2019-01-22 17:02:29 +0900 [INFO] (0017@[0:default]+test1+task1): py>: params.show_export
show_export: table = carried information
2019-01-22 17:02:29 +0900 [INFO] (0017@[0:default]+test1+task2): sh>: ./show1.sh
table = carried information
Success.
```

sonots:
I am asking how to use `digdag.env.export` here. I mean that I am doubting its functionality is not working.

(Also, digdag.env.store shows the parameters set at Digdag Web UI. So, we can not use it to map digdag secrets to environment variables).

yoyama:
As you mentioned, I also think the example does not work.
[...]

PR #969, opened 2019-02-05
Fix documentation of digdag.env.export

sonots:
Fix https://github.com/treasure-data/digdag/issues/944

* digdag.env.export does not affect following tasks, but affect only child tasks.
* child tasks can be created only via digdag.env.add_subtask currently.


yoyama:
LGTM
Very thanks!

PR #976, opened 2019-02-08
Fix td_load> parameter issue.

yoyama:
If td_load> operator with no parameter, it throws AssertError which not inherit RuntimeException.
As result exception is through OperatorManager and caught by MultiThreadAgent and never finish.

This patch change AssertError to ConfigException and add test for config parameter.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/td/TdLoadOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/td/TdLoadOperatorFactory.java
@@ -85,7 +85,7 @@ public class TdLoadOperatorFactory
             if ( numExistParams > 1) {
                 throw new ConfigException("Only the command or one of the config and name params may be set");
             }
-            else if (numExistParams <= 0) {
+            else if (numExistParams == 0) {
                 throw new ConfigException("No parameter is set");
             }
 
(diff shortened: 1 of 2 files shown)


muga:
good catch 👍

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdLoadOperatorFactory.java
```diff
@@ -104,7 +108,7 @@ else if (config.isPresent()) {
                 this.sessionName = Optional.absent();
             }
             else {
-                throw new AssertionError();
+                throw new ConfigException("Unexpected config error");
```
muga: We already have `numExistParams <= 0` that you added in this PR so, I think that we don't need to the `else` statement any more. or, 2nd `else if` can be changed with `else`? What do you think?

```
else { // config.isPresent()
    this.embulkConfig = Optional.of(config.get().getInternalObjectNode());
    this.sessionName = Optional.absent();
}
```

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdLoadOperatorFactory.java
```diff
@@ -81,9 +81,13 @@ protected TdLoadOperator(OperatorContext context)
                 throw new ConfigException("The parameters config and name cannot both be set");
             }
 
-            if (Stream.of(command, name, config).filter(Optional::isPresent).count() > 1) {
+            long numExistParams = Stream.of(command, name, config).filter(Optional::isPresent).count();
+            if ( numExistParams > 1) {
                 throw new ConfigException("Only the command or one of the config and name params may be set");
             }
+            else if (numExistParams <= 0) {
```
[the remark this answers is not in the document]
yoyama: Thanks for your pointing out. I agree.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdLoadOperatorFactory.java
yoyama: Yes, I also think this code may never reached. I will fix it with your proposal. Thanks!

PR #978, opened 2019-02-08
Update jacoco

wreulicke:
Update jacoco.

jacoco officially supports Java11 from 0.8.3

You can see changelogs below.
* https://github.com/jacoco/jacoco/releases/tag/v0.8.3
* https://github.com/jacoco/jacoco/releases/tag/v0.8.2

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,7 +45,7 @@ allprojects {
     }
 
     jacoco {
-        toolVersion = '0.8.1'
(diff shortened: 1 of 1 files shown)


muga:
Hi @wreulicke, thank you for suggesting. Digdag now using Java8 but, let us try to upgrade. If it doesn't work, it will be reverted.

wreulicke:
Thanks for your review.

PR #981, opened 2019-02-08
Update gradle and spotbugs

wreulicke:
* Update gradle wrapper
* Update spotbugs and gradle plugin
* Fix deprecations about gradle
  * enableFeaturePreview('STABLE_PUBLISHING') is not required since gradle 5.
  * Migrate new API from deprecated API of jacoco plugin

Deprecated list since Gradle 5.2 is here.
https://docs.gradle.org/current/javadoc/deprecated-list.html


gradlew
```diff
@@ -28,7 +28,7 @@ APP_NAME="Gradle"
 APP_BASE_NAME=`basename "$0"`
 
 # Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-DEFAULT_JVM_OPTS=""
+DEFAULT_JVM_OPTS='"-Xmx64m"'
```
muga: May I ask you about the reason why we'd better to put `Xmx64m` here?

gradlew.bat
```diff
@@ -14,7 +14,7 @@ set APP_BASE_NAME=%~n0
 set APP_HOME=%DIRNAME%
 
 @rem Add default JVM options here. You can also use JAVA_OPTS and GRADLE_OPTS to pass JVM options to this script.
-set DEFAULT_JVM_OPTS=
+set DEFAULT_JVM_OPTS="-Xmx64m"
```
muga: ditto

gradlew
wreulicke: This change is since gradle 5 and Gradle generated this.

This change is described here.
https://docs.gradle.org/current/userguide/upgrading_version_4.html#rel5.0:default_memory_settings

muga: @wreulicke Got it, thanks. I just curious about it.

muga:
@wreulicke It seems like the upgrading started breaking AppVeyor CI jobs. Since AppVeyor Github integration also broken at that time, we didn't notice those failure.. I will revert it first for safe releasing new version of Digdag.

wreulicke:
oh, sorry.

PR #989, opened 2019-02-12
Update status-filter

szyn:
I've updated the status filter as the [inline forms](https://getbootstrap.com/docs/3.4/css/#forms-inline).

screen shot 2019-02-13 at 0 36 19


yoyama:
LGTM 
Thanks!

PR #1004, opened 2019-02-21
[doc] Add session command to the document

kamikaseda:


yoyama:
Thanks for your many contributions.
I checked the command, and I noticed **sessions** and **session** is same command.

In the code, "sessions" is set as alias of "session" as follows.

https://github.com/treasure-data/digdag/blob/master/digdag-cli/src/main/java/io/digdag/cli/Main.java#L139

I also tested both commands and it looks like same behavior.
Could you confirm it ?

I think sessions command help also mislead.

```
2019-02-21 15:47:48 +0900: Digdag v0.9.33
Usage: digdag sessions                         show sessions for all workflows                   
       digdag sessions <project-name>          show sessions for all workflows in a project      
       digdag sessions <project-name> <name>   show sessions for a workflow                      
       digdag session  <session-id>            show a single session
```

kamikaseda:
@yoyama Oh, I see.

In my opinion, we have 2 patterns of modification.

*Pattern 1*
```
Usage: digdag session(s)                         show sessions for all workflows                   
       digdag session(s) <project-name>          show sessions for all workflows in a project      
       digdag session(s) <project-name> <name>   show sessions for a workflow                      
       digdag session(s)  <session-id>            show a single session
```
And add sample command to doc.

```
$ digdag sessions
$ digdag sessions myproj
$ digdag sessions myproj main
$ digdag session <session-id>
```

*Pattern 2*
Apply my current PR.
Even `sessions` and `session` is same behavior, current help is understandable and no problem because `sessions` returns session list and `session` returns single session.


How do you think?

yoyama:
@muga Which is better?

muga:
@yoyama it's ok for me to take original commit by @kamikaseda.

PR #1014, opened 2019-02-22
[doc] Modify from Job Name to Unique ID

kamikaseda:
Document says `td_load>` can use "Job name" instead of yml. However, "Job name" is not understandable and make a confuse. It would be said "Unique ID".
Refer: https://support.treasuredata.com/hc/en-us/articles/360001474328-Reference-an-Input-Data-Transfer#Configuring%20your%20Unique%20ID%20Incremental%20Data%20Transfer

And remove duplicated "td.last_job_id" item.

### DIFF
--- a/digdag-docs/src/operators/td.md
+++ b/digdag-docs/src/operators/td.md
@@ -208,8 +208,7 @@
 
 ## Output parameters
 
-* **td.last_job_id**
-* **td.last_job.id**
+* **td.last_job_id** or **td.last_job.id**
 
   The job id this task executed.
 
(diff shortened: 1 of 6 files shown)


kamikaseda:
I mis-read `td.last_job_id` and `td.last_job.id`.... It would be better to re-write.

muga:
@kamikaseda thanks for fixing. As I took a look the impl. of td_load operator. At least the command name doesn't represent TD job name. You're correct. But, I think that "Unique ID" also will make users confused. According to the impl., that's the name of data transfer, former bulk_load_session. What do you think that we'd better to change "Unique ID" with the name of data transfer in both of Digdag doc and support doc?

kamikaseda:
@muga Naming rules is wholly Treasure Data internal issue. I think we should unify a name based on Treasure Data's official document for now.

muga:
@kamikaseda got it, thanks. I think that `uniq id` still make confising but, better than `job name`. let's merge it.

Issue #922, opened 2018-11-26
Invalid cron doesn't return right error message


darkchilli:
Digdag v0.9.31
Issue: When pushing to digdag server from digdag-client, not returning helpful error message.

--- Dig file ---

schedule:
  cron>: 1 4,8,12,16,20
...
...

--- digdag push ---
digdag push corporate -e xx.xx.xx.xx:65432
2018-11-26 10:53:46 +1100: Digdag v0.9.31
Creating .digdag/tmp/archive-8215978397101272324.tar.gz...
  Archiving xxxxxxx.dig
Workflows:
  xxxxxxx.dig
error: HTTP 500 Internal Server Error (internal server error)

--- Expect ---
Invalid cron error should be returned to client, at the moment it seems to be only available on digdag server side.
Digdag server log:
 Caused by: it.sauronsoftware.cron4j.InvalidPatternException: invalid pattern: "1 4,8,12,16,20"

hiroyuki-sato:
Hello, @darkchilli 

Thank you for reporting the issue.
I think It is better to report the right message when invalid format pushed.

Could you try  `cron>: 1 4,8,12,16,20 * * *`?

yoyama:
I will close it, because #1026 may solve this issue. If you still have the issue feel free to reopen.

Issue #859, opened 2018-09-06
pushing dig project to server should verify digs


yanivyalda:
I pushed a couple of dig files to the server today without checking locally. The incremented version was displayed but the old code still ran. It would be great to verify the uploaded dig's 

Have I missed something?

hiroyuki-sato:
Hello, @yanivyalda 

Digdag doesn't stop running workflow when you push new revision.
Could you tell us what is "should verify" mean?

For example,
Does this mean Digdag should reject push command when an old revision is running?

yanivyalda:
Hi @hiroyuki-sato 

With Verify. I mean it should check the syntax of the dig before pushing. What happens I think is the version number updates but the code still stays the same. I'm going to run a few more tests today to get the correct steps to replicate.

Rejecting push command when old revision is running is a good feature but I have not encountered this.

yanivyalda:
I have tested the following today and here are the steps

**Syntax checking works**
+test1:
sh>: echo 'I should fail'

digdag push track --project -r 0.0.0.1 -endpoint XXXX
_2018-09-07 09:24:05 +1000: Digdag v0.9.27
Creating .digdag/tmp/archive-3086874888794404513.tar.gz...
error: Validating project failed_

**Syntax checking fails**
+test2:
&nbsp;&nbsp;&nbsp;&nbsp;inner_test:
&nbsp;&nbsp;&nbsp;&nbsp;sh>: echo 'I should fail'

digdag push track --project -r 0.0.0.1 -endpoint XXXX
2018-09-07 09:24:21 +1000: Digdag v0.9.27
Creating .digdag/tmp/archive-7202906967196748735.tar.gz...
.
.
.
Use `digdag workflows` to show all workflows.

hiroyuki-sato:
Hello, @yanivyalda 

Thank you for investigating.

The `inner_test:` is valid syntax in Digdag.
`sh>` operator just ignore `inner_test` value.

Digdag uses those values as parameter.
For example, `mail>:` and `sh>:`

```yaml
+step1:
  mail>: body.txt
  subject: workflow started
  to: [me@example.com]
```

```yaml
+sample:
  sh>: set
  shell: [/bin/zsh]
```

yanivyalda:
hi @hiroyuki-sato that was a markdown typo
it should have been
+inner_test:

I also noticed this, the console updated the revision
  revision: 0.0.35
but the ui still kept the old version.

hiroyuki-sato:
[...]

PR #1026, opened 2019-02-26
Digdag push validation on the server side.

hiroyuki-sato:
Hello, @muga , @yoyama 

Could you advise me how to implement a validation?

`digdag push` requires validation on the server side.
It requires two things at least. And this issue will Fix two issues.
* task validation(#859).
* schedule validation(#922).

I implement the first one. Second schedule validation seems to require `SchedulerManager` class.
https://github.com/hiroyuki-sato/digdag/commit/5004fb635e4ec598255f66593c45d7ab7f4ce5e6#diff-84a9dc43263ac1f78fa3b2aca8234f6fR576

I'm not familiar with Google Guice. I'm not sure how to instantiate it. Do you have any idea for that?

This code almost copies of `digdag check` code.
https://github.com/treasure-data/digdag/blob/master/digdag-cli/src/main/java/io/digdag/cli/Check.java

Best regards.

### DIFF
--- a/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
+++ b/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
@@ -556,23 +556,8 @@ public class ProjectResource
                     if (md5Count.getCount() != contentLength) {
                         throw new IllegalArgumentException("Content-Length header doesn't match with uploaded data size");
                     }
-                    WorkflowDefinitionList defs = meta.getWorkflowList();
-                    for (WorkflowDefinition def : defs.get()) {
-                        Workflow wf = compiler.compile(def.getName(), def.getConfig());
 
-                        // validate workflow and schedule
-                        for (WorkflowTask task : wf.getTasks()) {
-                            // raise an exception if task doesn't valid.
-                            task.getConfig();
-                        }
-                        Revision rev = Revision.builderFromArchive("check", meta, getUserInfo())
-                                .archiveType(ArchiveType.NONE)
-                                .build();
-                        // raise an exception if "schedule:" is invalid.
-                        srm.tryGetScheduler(rev, def);
-
-                    }
-
+                    validateWorkflowAndSchedule(meta);
                 }
 
                 ArchiveManager.Location location =
@@ -702,6 +687,26 @@ public class ProjectResource
             throw new IllegalArgumentException(String.format(ENGLISH,
                         "Size of a file in the archive exceeds limit (%d > %d bytes): %s",
                         entry.getSize(), MAX_ARCHIVE_FILE_SIZE_LIMIT, entry.getName()));
+        }
+    }
+
+    private void validateWorkflowAndSchedule(ArchiveMetadata meta)
+    {
+        WorkflowDefinitionList defs = meta.getWorkflowList();
+        for (WorkflowDefinition def : defs.get()) {
+            Workflow wf = compiler.compile(def.getName(), def.getConfig());
+
+            // validate workflow and schedule
+            for (WorkflowTask task : wf.getTasks()) {
+                // raise an exception if task doesn't valid.
+                task.getConfig();
+            }
+            Revision rev = Revision.builderFromArchive("check", meta, getUserInfo())
+                    .archiveType(ArchiveType.NONE)
+                    .build();
+            // raise an exception if "schedule:" is invalid.
+            srm.tryGetScheduler(rev, def);
+
         }
     }
 
--- a/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
+++ b/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
@@ -1,6 +1,8 @@
 package io.digdag.server.rs;
 
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.net.URI;
 import java.time.Instant;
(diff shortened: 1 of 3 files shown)


muga:
Hi @hiroyuki-sato, thank you for taking this improvement.

> I'm not familiar with Google Guice. I'm not sure how to instantiate it. Do you have any idea for that?

`ScheduleManager` is already injected in digdag-server `ServerModule`. Please add the `ScheduleManager` parameter to the constructor in `ProjectResource` like `WorkflowCompiler`. Then, we can use the object in the class without the object creation explicitly. Your code change basically sounds good to me.

hiroyuki-sato:
@muga Thank you for the advice.

I implement it. I'll add integration tests later.


### validate tasks.

```
digdag push hoge
2019-02-26 16:39:41 +0900: Digdag v0.9.33
Creating .digdag/tmp/archive-5515688681553749096.tar.gz...
  Archiving fuga.dig
  Archiving scripts/myscript.sh
Workflows:
  fuga.dig
error: Status code 400: {"message":"io.digdag.core.repository.ModelValidationException: Validating workflow failed\n+fuga+setup contains invalid keys: 'sh': \"{\"sh\":\"echo start ${session_time}\"}\"","status":400}
```

### validate schedule.

```
2019-02-26 16:42:19 +0900: Digdag v0.9.33
Creating .digdag/tmp/archive-6443414997521903051.tar.gz...
  Archiving fuga.dig
  Archiving scripts/myscript.sh
Workflows:
  fuga.dig
error: Status code 400: {"message":"hourly>: scheduler requires mm:ss format: test","status":400}
```

muga:
@hiroyuki-sato it's cool. thanks!

hiroyuki-sato:
@muga. Please take a look when you have time.

I created `ValidateProjectIT.java` for testing this PR.
Is it better to place those tests in another class?

Those comment part doesn't need to this class. I will just remove it.
What do you think?. I'm not sure it possible to validate more.

```java
                        // validate workflow and schedule
//                        Set<String> required = new HashSet<>();
                        for (WorkflowTask task : wf.getTasks()) {
                            // raise an exception if task doesn't valid.
                            task.getConfig();
//                            String require = config.getOptional("require>", String.class).orNull();
//                            if (require != null && required.add(require)) {
//                                f.ln("  -> %s", require);
//                            }
```

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
```diff
@@ -549,6 +556,28 @@ public RestProject putProject(@QueryParam("project") String name, @QueryParam("r
...
+//                            }
+                        }
+                        Revision rev = Revision.builderFromArchive("check", meta, getUserInfo())
+                                .archiveType(ArchiveType.NONE)
+                                .build();
+                        // raise an exception if "schedule:" is invalid.
+                        srm.tryGetScheduler(rev, def);
+
+                    }
+
```
muga: To keep the logic of putProject method simple, can you separate your validation logic from this method and copy the logic to another private method that you will create?

hiroyuki-sato:
@muga Thank you for the comment. I removed unnecessary comments and moved validation into a separate method. Please take a look when you have time.

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
```diff
@@ -1,6 +1,8 @@
 package io.digdag.server.rs;
 
+import java.util.HashSet;
```
yoyama: Unused import?

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
```diff
@@ -1,6 +1,8 @@
...
 import java.util.List;
+import java.util.Set;
```
yoyama: Unused import?

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
```diff
@@ -82,9 +86,12 @@
 import io.digdag.core.session.SessionStoreManager;
 import io.digdag.core.session.StoredSessionWithLastAttempt;
 import io.digdag.core.storage.ArchiveManager;
+import io.digdag.core.workflow.Workflow;
 import io.digdag.core.workflow.WorkflowCompiler;
+import io.digdag.core.workflow.WorkflowTask;
 import io.digdag.server.GenericJsonExceptionHandler;
 import io.digdag.spi.DirectDownloadHandle;
+import io.digdag.spi.Scheduler;
```
yoyama: Unused import?

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
hiroyuki-sato: Separated validation method.

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
hiroyuki-sato: Removed

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
hiroyuki-sato: Removed

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
hiroyuki-sato: Removed

hiroyuki-sato:
@yoyama Thank you for the comment. There is still a much-unused import statement. That's why I can't use Optimize imports feature in IntelliJ. 😢 
Anyway, I removed those imports which I added.
Please take a look when you have time.

muga:
merged..

hiroyuki-sato:
Thank you!

PR #1044, opened 2019-03-05
Fix database.ThreadLocalTransactionManager to add instanceof for 4th exception handling

muga:
follow-up of https://github.com/treasure-data/digdag/pull/936.

On https://github.com/treasure-data/digdag/pull/936, we made change of begin method in database.ThreadLocalTransactionManager. But we also needed to fix and add 4th exception handling.


muga:
@yoyama thanks! merging.

Issue #946, opened 2019-01-23
[feature request] Add an option to disable scheduler with digdag server


sonots:
Currently, digdag server has following options:

* --disable-executor-loop
* --disable-local-agent

But, `--disable-executor-loop` stops both schedule executor and workflow executor.
I want to disable only the schedule executor.

The aim is for development purpose.

I am mainly using `digdag start` with `digdag server` for development because they work slightly different with `digdag run`.
While I am developing dig files, I do not need the scheduler.

hiroyuki-sato:
Hello, @sonots 

If you don't write `schedule:` part in dig file, It doesn't run scheduled tasks.

What do you think to use it until implement `--disable-scheduler`?

sonots:
I want to write `schedule:` part on dig file because it will be written for deployment. I do not want to even comment it out.

My workaround which I come up with is that `digdag push && digdag disable` immediately (although it has a potential risk that the workflow is executed by the scheduler)

takashima0411:
In my workaround, It separates schedule and dig file.

example.
production.dig
```
schedule:
  daily>: 20:00:00
```

local.dig
```
// no schedule settings here
```

sample.dig
```
!include :  env.dig

+some_scheduled_tasks:...
```

In local mode, copy local.dig to env.dig.
In production mode, copy production.dig to env.dig by CI server.

sonots:
fixed via #1048

PR #1048, opened 2019-03-12
Add digdag server --disable-scheduler option

sonots-zozo:
Implemented the feature request https://github.com/treasure-data/digdag/issues/946.

Recently, I am setting up a staging environment. I want to disable digdag scheduler on staging.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/schedule/ScheduleConfig.java
+++ b/digdag-core/src/main/java/io/digdag/core/schedule/ScheduleConfig.java
@@ -1,0 +1,27 @@
+package io.digdag.core.schedule;
+
+import io.digdag.client.config.Config;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.immutables.value.Value;
+
+@Value.Immutable
+@JsonSerialize(as = ImmutableScheduleConfig.class)
+@JsonDeserialize(as = ImmutableScheduleConfig.class)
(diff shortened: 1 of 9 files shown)


sonots-zozo:
I could not find any tests for digdag server commands. Please let me know how and where we should write it.

sonots-zozo:
I am glad if this is included in 0.9.36.

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleConfig.java
```diff
@@ -0,0 +1,27 @@
+package io.digdag.core.schedule;
```
sonots: I referred AgentConfig.java

digdag-core/src/main/java/io/digdag/core/schedule/ScheduleConfigProvider.java
```diff
@@ -0,0 +1,24 @@
+package io.digdag.core.schedule;
+
+import java.lang.management.ManagementFactory;
```
yoyama: Unused import?

sonots: removed. (also from AgentConfigProvider.java).

yoyama:
LGTM. Thanks!

sonots:
@yoyama @muga I noticed that schedules are shown via `digdag shedules` even if digdag server is working with `--disable-scheduler`. It is mis-leading.

Is it possible to hide or non-register schedules if digdag server starts with `--disable-scheduler`?

PR #1052, opened 2019-03-18
update redshift.md

kajuki1113:
I added the explanation about the timing of status table being dropped.
In my understanding, cleanup method do this and the condition is defined as Duration.ofHours(24).

If this is not correct, please point out.

### DIFF
--- a/digdag-docs/src/operators/redshift.md
+++ b/digdag-docs/src/operators/redshift.md
@@ -171,6 +171,7 @@ _export:
 
   Whether this operator uses a strict transaction to prevent generating unexpected duplicated records just in case. *Default*: `true`.
   This operator creates and uses a status table in the database to make an operation idempotent. But if creating a table isn't allowed, this option should be false.
+  If the query that created the status table completed 24 hours ago, this operator drop the table in the cleanup step.
 
   Examples:
 


yoyama:
I am not familiar with redshift operator, but it may be correct.
In addition this value is default and it looks like be configurable with _status_table_cleanup_ param.

https://github.com/treasure-data/digdag/blob/dbe0284c1300916ce4ce54781979db9ba051083a/digdag-standards/src/main/java/io/digdag/standards/operator/jdbc/AbstractJdbcOperator.java#L69

kajuki1113:
> I am not familiar with redshift operator, but it may be correct.
I am sorry for set you as reviewer...
Thank you for review.


> In addition this value is default and it looks like be configurable with status_table_cleanup param.
Yes, I could find it. However, I cannot find how to set it in the operator.

hiroyuki-sato:
Hello, @kajuki1113 

Have you ever tried the following?
I can't test it but It may work.

```yaml
+task1
  redshift>: xxx
  #status_table: __digdag_status
  #status_table_schema: another_schema
  status_table_cleanup: 48s
```

Or

```yaml
_export
  redshift:
    #status_table: __digdag_status
    #status_table_schema: another_schema
    status_table_cleanup: 48s

+task1:
  redshift> xxx
```

kajuki1113:
@hiroyuki-sato 
Thank you for great advise.
As a result of trying, status_table_cleanup work fine.
I will add the explanation later.

muga:
@kajuki1113 thank you for fixing the document. Since "status_table_cleanup" is already extracted as an option so, we'd better to add a description for this option separately. What do you think?

PR #1054, opened 2019-03-19
Fix ParamServer DEFAULT_TTL value.

### LINKED ISSUE #1053: ParamServer returns empty values
digdag version: 0.9.32

I'm getting inconsistent results using the paramstore. After 1 day I get empty values returned. I'm using a postgres database to store the values. The values look good in the database.

The problem seems to be a wrong TTL setting in file `ParamServerClient.java` (https://github.com/treasure-data/digdag/commit/7b6b91d59ec1f7494d1ef472f6d4e47e86962d32). It should be 90 days but it's actually calculated incorrectly and it ends up being 1.5 days. 

The documentation does not mention any reference to this TTL, which seems very important. This should definitely be added.

Aside from getting the correct TTL calculated, we should be able to configure the TTL outside of the code. For my own purposes, I'd like the param store to serve as a permanent storage for system-wide configurations. Currently there is no way to set system-wide values.

hiroyuki-sato:
This PR fix #1053. 90 days in sec is `60sec * 60min * 24 hours * 90days`

I have some notes.

* I added a simple test for Redis server. ~~This PR doesn't contain Redis test: Original codes doesn't contain expire test in Redis too~~.
* A user wants to change `DEFAULT_TTL_IN_SEC`  (#1053)

I think it is useful that if `DEFAULT_TTL_IN_SEC` is configurable.

But current `ParamServerClient` which set TTL can't get System Configuration parameter.

https://github.com/treasure-data/digdag/blob/master/digdag-standards/src/main/java/io/digdag/standards/operator/param/ParamServerClientFactory.java

It needs to change constructor like the following

```java
ParamServerClient paramServerClient = ParamServerClientFactory.build(connection, objectMapper,systemConfig);
```

@yoyama, @muga What do you think?

* Configurable `DEFAULT_TTL_IN_SEC`  is better?
* How `ParamServerClientFactory.build` method change (Add systemConfiguration paramter? or add TTL param only?)

### DIFF
--- a/digdag-tests/src/test/java/acceptance/ParamSetRedisIT.java
+++ b/digdag-tests/src/test/java/acceptance/ParamSetRedisIT.java
@@ -8,6 +8,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 
 import static java.util.Arrays.asList;
+import static junit.framework.TestCase.assertTrue;
 import static org.hamcrest.core.Is.is;
 import static org.junit.Assert.assertThat;
 import static utils.TestUtils.addWorkflow;
@@ -17,6 +18,9 @@ import static utils.TestUtils.main;
 public class ParamSetRedisIT
         extends BaseParamRedisIT
 {
+    private static long TTL_90_DAYS_PLUS_1SEC = 60 * 60 * 24 * 90 + 1; // 90days + 1sec
+    private static long TTL_89_DAYS = 60 * 60 * 24 * 89; // 89days
+
     @Test
     public void setValueToRedis()
             throws IOException
@@ -36,6 +40,13 @@ public class ParamSetRedisIT
                 projectDir.resolve("set.dig").toString()
         );
         assertCommandStatus(status);
+
+        long key1_ttl = redisClient.ttl("0:key1");
+        long key2_ttl = redisClient.ttl("0:key2");
+
+        assertTrue(key1_ttl < TTL_90_DAYS_PLUS_1SEC && key1_ttl > TTL_89_DAYS);
+        assertTrue(key2_ttl < TTL_90_DAYS_PLUS_1SEC && key2_ttl > TTL_89_DAYS);
+
         assertThat(redisClient.get("0:key1"), is("{\"value_type\":0,\"value\":{\"value\":\"value1\"}}"));
         assertThat(redisClient.get("0:key2"), is("{\"value_type\":0,\"value\":{\"value\":\"value2\"}}"));
     }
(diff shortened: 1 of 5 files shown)


hiroyuki-sato:
This commit is the `param_server.ttl` sample implementation.
https://github.com/hiroyuki-sato/digdag/commit/9bb7e906a2a9fbf31732ce03cd6e1600209c3cbf

digdag-tests/src/test/java/acceptance/ParamSetRedisIT.java
```diff
@@ -36,6 +40,13 @@ public void setValueToRedis()
                 projectDir.resolve("set.dig").toString()
         );
         assertCommandStatus(status);
+
+        long key1_ttl = redisClient.ttl("0:key1");
+        long key2_ttl = redisClient.ttl("0:key2");
+
+        assertTrue(key1_ttl < TTL_90_DAYS_PLUS_1SEC && key1_ttl > TTL_89_DAYS);
+        assertTrue(key2_ttl < TTL_90_DAYS_PLUS_1SEC && key2_ttl > TTL_89_DAYS);
+
```
yoyama: Basic question. Why compare as ttl < TTL_90_DAYS_PLUS_1SEC not ttl <= TTL_90_DAYS?

yoyama:
@hiroyuki-sato Sorry for my late response.
I also think it is useful if TTL is configurable.
But it may be better to create another PR for it.

digdag-tests/src/test/java/acceptance/ParamSetRedisIT.java
hiroyuki-sato: To tell the truth, I have no idea how to test `param_set>` with the Redis server.

`param_set>`  set parameter X with TTL 90days.(60sec * 60min * 24 hours * 90days)
Redis can't set past TTL like PostgreSQL.

If assert check immediately after `param_set`, It will be `key1_ttl == TTL_90_DAYS`.
But if that test delay some reason, The TTL will `TTL_90_DAYS - XXsec`
That's why I checked >= TTL_89_DAYS

If TTL configurable, that test will be easy.

* set TTL 1sec;
* `param_set>` xxx
* wait 1sec
* Check the value already expire.

yoyama: Thanks!

muga:
good catch!! @hiroyuki-sato 👍

hiroyuki-sato:
Thank you!

PR #1059, opened 2019-03-24
[wip] Add secretKey field to ac.ProjectTarget to use the field value as target of access control

muga:
This PR adds secretKey field to ac.ProjectTarget to use the field value as target of access control, which was merged from #936.


yoyama:
I will close this PR. It is old and very large number of file will be modified.

PR #1065, opened 2019-03-27
Add RestTask#getError method to pass error config to clients from AttemptResource#getTasks

muga:
Though it's worth for users to check the reason why tasks failed by both of user errors and system errors, AttemptResource#getTasks endpoint doesn't return error configs to users. Users don't have change to see and handle the error except error task.

This PR fixes RestTask that is model class to pass error configs to clients from AttemptResource#getTasks.


muga:
Thanks for checking. merged.

PR #1068, opened 2019-03-28
UI: Show attempt params to attempt/session views

RyochanUedasan:
I've made a following change:
 - Show attempt params to attempt/session views.

A lot of sessions are invoked with params (`digdag start ... -p xxx=yyy ...`), I have trouble with figuring out what I want to see, so make it easy to find which param is used.

I think `.params-view`'s width is not needed, so deleted it.


sgswtky:
👍

yoyama:
@szyn Could you review if possible ?

yoyama:
@RyochanUedasan Thanks for your contribution!
@szyn Thanks for your review!
I will merge it later.

PR #1070, opened 2019-03-29
Add client options in usage

hiroyuki-sato:
Hello, @yoyama, @muga 

What do you think to add this message?
https://github.com/treasure-data/digdag/blob/master/digdag-cli/src/main/java/io/digdag/cli/client/ClientCommand.java#L46-L57

```
  Client options:
    -e, --endpoint URL               Server endpoint (default: http://127.0.0.1:65432)
    -H, --header  KEY=VALUE          Additional headers
    --disable-version-check          Disable server version check
    --disable-cert-validation        Disable certificate verification
```

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/Main.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/Main.java
@@ -336,7 +336,7 @@ public class Main
         err.println("  Options:");
         showCommonOptions(env, err);
         err.println("  Client options:");
-        showClientCommonOptions(env, err);
+        showClientCommonOptions(err);
         if (error == null) {
             err.println("Use `<command> --help` to see detailed usage of a command.");
             return systemExit(null);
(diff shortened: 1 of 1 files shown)


digdag-cli/src/main/java/io/digdag/cli/Main.java
```diff
@@ -353,4 +355,14 @@ public static void showCommonOptions(Map<String, String> env, PrintStream err)
         err.println("    --version                        show client version");
         err.println("");
     }
+
+    public static void showClientCommonOptions(Map<String, String> env, PrintStream err)
```
yoyama: It looks like the arg. env is not used.

yoyama:
@hiroyuki-sato  Thanks for your proposal.
I think it is basically good.
But I added a comment. Please check it.

digdag-cli/src/main/java/io/digdag/cli/Main.java
hiroyuki-sato: Removed `env` argument

hiroyuki-sato:
@yoyama Thank you for the comment. Fixed your comment. Please take a look when you have time.

PR #1072, opened 2019-04-01
HikariCP JMX metrics

yoyama:
I would like to introduce following to improve DB related monitoring.

1. Enable HikariCP JMX metrics when JMX is enabled
2. Introduce database.leakDetectionThreshold.

### DIFF
--- a/digdag-docs/src/command_reference.rst
+++ b/digdag-docs/src/command_reference.rst
@@ -356,7 +356,7 @@ In the config file, following parameters are available
 * database.idleTimeout (seconds in integer, default: 600)
 * database.validationTimeout (seconds in integer, default: 5)
 * database.maximumPoolSize (integer, default: available CPU cores * 32)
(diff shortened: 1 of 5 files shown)


digdag-docs/src/command_reference.rst
```diff
@@ -356,6 +356,7 @@ In the config file, following parameters are available
 * database.idleTimeout (seconds in integer, default: 600)
 * database.validationTimeout (seconds in integer, default: 5)
 * database.maximumPoolSize (integer, default: available CPU cores * 32)
+* database.leakDetectionThreshold (HikariCP leakDetectionThreshold in integer. default: 0. To enable, set to >= 2000.)
```
hiroyuki-sato: Could you add unit description(`milliseconds`)?

yoyama: @hiroyuki-sato Thanks, I will fix it later.

PR #1073, opened 2019-04-03
Fix the issue which causes infinite task retry.

yoyama:
If workflow has generated subtasks with reference to undefined variable, nashorn throw AssertionError.
But digdag cannot catch it correctly and retry infinitely.
In addtion if some Operators also throw AssertionError, same issue will happen.
This PR handles these cases.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -158,7 +158,7 @@ public class OperatorManager
                                 ex.getError(cf).get());  // TODO is error set?
                     }
                 }
-                catch (RuntimeException | AssertionError ex) {
+                catch (RuntimeException | AssertionError ex) { // Avoid infinite task retry cause of AssertionError by Operators
                     if (ex instanceof ConfigException) {
                         logger.error("Configuration error at task {}: {}", request.getTaskName(), formatExceptionMessage(ex));
                     }
@@ -203,7 +203,7 @@ public class OperatorManager
         catch (RuntimeException ex) {
             throw new RuntimeException("Failed to process variables", ex);
         }
-        catch (AssertionError ex) {
+        catch (AssertionError ex) { // Avoid infinite task retry cause of AssertionError by ConfigEvalEngine(nashorn)
             throw new RuntimeException("Unexpected error happened in ConfigEvalEngine: " + ex.getMessage(), ex);
         }
         logger.debug("evaluated config: {}", config);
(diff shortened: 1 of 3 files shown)


digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -158,7 +158,7 @@ private void runWithHeartbeat(TaskRequest request)
                                 ex.getError(cf).get());  // TODO is error set?
                     }
                 }
-                catch (RuntimeException ex) {
+                catch (RuntimeException | AssertionError ex) {
```
muga: minor: May I ask you to comment about the reason why we need to handle "AssertionError" here?

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -203,6 +203,9 @@ private void runWithWorkspace(Path projectPath, TaskRequest request)
         catch (RuntimeException ex) {
             throw new RuntimeException("Failed to process variables", ex);
         }
+        catch (AssertionError ex) {
```
muga: ditto

yoyama: In this case we can catch the AssertionError because of ConfigEvalEngine(it calls nashorn).

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
yoyama: In this case we can catch AssertionError which is thrown by operators.run() and around processing in runWithWorkspace(). For example EMR operator may throw AssertionError.

yoyama: If we don't catch AssertionError here, it will be caught by above modification, but I guess by catching ConfigEvalEngine related AssertionError in this place may make clear the cause. So I also catch it.

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
yoyama: If we don't catch AssertionError here, it will be caught by above modification. So this fix is not required. But I guess by catching ConfigEvalEngine related AssertionError in this place may make clear the cause. So I also catch it.

yoyama:
@muga Thanks for your pointing out. I replied them. If you have any additional comments, please let me know.

muga:
@yoyama Ah sorry! my last comment did make your confusing. My point was that I wanted to ask you to make comments in the source code. Because, in general, i don't think that handling assertion error is good approach to solve something as you know. But we need to do so, we'd better to explain the reason. Your change itself make sense 👍

yoyama:
@muga Sorry for my misunderstanding, I added comments to code. Thanks for your pointing out.
@hiroyuki-sato also thanks for pointing out my misunderstanding

Issue #1055, opened 2019-03-19
[question] how to start a new session attempt of a workflow over the REST API


ehaupt:
How do you start a new session attempt of a workflow over the REST API?

I am looking for the equivalent of:

`
$ digdag start myproj mywf --session now -P ./my_input.yml
`

I've studied the available REST API calls that are exposed by swagger but couldn't find the corresponding option.

hiroyuki-sato:
Hello, @ehaupt 

I've never tried it. but It may `PUT /api/atttempts`.

https://github.com/treasure-data/digdag/blob/master/digdag-cli/src/main/java/io/digdag/cli/client/Start.java#L175

https://github.com/treasure-data/digdag/blob/master/digdag-client/src/main/java/io/digdag/client/DigdagClient.java#L718-L724

![api_attempts](https://user-images.githubusercontent.com/767650/54608493-2a7d9800-4a94-11e9-919b-e364fde4b532.png)

ehaupt:
Thank you @hiroyuki-sato. I'll test it. I'm currently trying to integrate a workflow in a self-care portal.

hiroyuki-sato:
Hello, @ehaupt 

JFYI, Here is 3rd party digdag clients.
* [mog](http://iroilog.net/2017/06/release-mog/) Digdag Client (golang)
* [digdag_client](https://github.com/saicologic/digdag_client) Ruby

ehaupt:
Hello @hiroyuki-sato, thank you again. This is very helpful.

ehaupt:
@hiroyuki-sato: I can confirm that `PUT /api/atttempts` works for my purpose.

It wouldn't accept `"sessionTime": "now"` but using ISO 8601 date format worked like a charm.

hiroyuki-sato:
Hello, @ehaupt 

Congratulation! 
JFYI `now` create current time with ISO8601 format. So maybe you made the same implementation.

https://github.com/treasure-data/digdag/blob/master/digdag-cli/src/main/java/io/digdag/cli/client/Start.java#L125

ehaupt:
Hello, @hiroyuki-sato

I've seen that someone else had a similar question and I was wondering if I should submit a quick [example](https://github.com/treasure-data/digdag/tree/master/examples) workflow with instructions on how to launch it over the REST API via curl.

PR #577, opened 2017-06-07
Proposal: optional Swagger API

akirakoyasu:
Add CLI option `--enable-swagger` for REST API document

## Test
Console A: Build digdag & run with `--enable-swagger`
```
./gradlew cli
./pkg/digdag-0.9.13-SNAPSHOT.jar server --memory --enable-swagger
```
Console B: Run Swagger-UI
```
docker run -dp 8080:8080 swaggerapi/swagger-ui
```
Browser: Access Swagger-UI
```
http://localhost:8080/?url=http://localhost:65432/api/swagger.json
```

PR #1075, opened 2019-04-05
Add an example workflow with instructions on how to call via REST API.

ehaupt:
In response to #1055 provide an example workflow with instructions on how to call via REST API.


hiroyuki-sato:
Hello, @ehaupt 
Thank you for creating the REST API example. :+1: 
@yoyama, @muga What do you think?

muga:
Hi @ehaupt @hiroyuki-sato, thank you for suggesting! REST API example is good idea 👍 .dig file is probably not good place I think.

Can we generate and show curl command example on our swagger API doc? We introduced enable-swagger option https://github.com/treasure-data/digdag/pull/577. I think that swagger doc is good place for show such example. Rather swagger-ui has "try it" button for each endpoint. The curl command generated by the button will be shown. Does it work for you maybe? 

@yoyama any idea?

muga:
@ehaupt We talked about this in-person, we want to merge the example that you made into master. Could you create new `examples/REST_API.md` file and move your content in `resttest.dig` to the new file? We'd better to collect REST API examples and separate them from .dig examples.

ehaupt:
@muga Good idea, I'll do that. I'll follow up shortly.

muga:
@ehaupt Thank you for fixing! it looks good to me. @yoyama @hiroyuki-sato What do you think?

examples/REST_API.md
```diff
@@ -0,0 +1,26 @@
...
+
+$ digdag push resttest
+```
+3. Call the workflow via REST API with curl to pass a message (workflowId has to be the value from the previous command)
+```
+$ curl -X PUT "http://localhost:65432/api/attempts" \
+       -H  "accept: application/json" \
+       -H  "Content-Type: application/json" \
+       -d "{ \"params\": { \"msg\": \"Hello from REST API.\" }, \"sessionTime\": \"$(date -Iseconds)\", \"workflowId\": \"1\"}"
+```
```
hiroyuki-sato: Does this example work on your platform?
I got the following error. (CentOS 7)

```
curl -X PUT "http://localhost:65432/api/attempts" \
>        -H  "accept: application/json" \
>        -H  "Content-Type: application/json" \
>        -d "{ \"params\": { \"msg\": \"Hello from REST API.\" }, \"sessionTime\": \"$(date -Iseconds)\", \"workflowId\": \"1\"}"
{"message":"Invalid ISO time format: %s2019-05-01T18:24:57+0900 (through reference chain: io.digdag.client.api.ImmutableRestSessionAttemptRequest$Json[\"sessionTime\"])","status":400}
```

Digdag allows iso8601 format like `2019-05-01T13:38:52+09:00` only. It require colon like '+09:00'. However `date -Iseconds` command returns `+0900`(without the colon).

What do you think to use date string like `2019-05-01T13:38:52+09:00` explicitly until fix format issue?

ehaupt: @hiroyuki-sato: this is interesting. There are slightly different outputs on other platforms:

FreeBSD:
```
$ uname -srvm
FreeBSD 12.0-RELEASE-p3 FreeBSD 12.0-RELEASE-p3 GENERIC  amd64
$ date -Iseconds
2019-05-01T21:12:41+02:00
```

Linux (RHEL7):
```
$ uname -srvm
Linux 3.10.0-957.10.1.el7.x86_64 #1 SMP Thu Feb 7 07:12:53 UTC 2019 x86_64
$ date -Iseconds
2019-05-01T21:13:31+0200
```

The following expression would produce the desired output in both cases:
```
$ date -Iseconds | sed -e 's|\([0-9]\{2\}\)\([0-9]\{2\}\)$|\1:\2|'
```

On a side note, I was wondering if it might be desirable to be able to specify a keyword such as `now` which would interpreted internally as the current moment?

Please let me know if the proposed workaround is acceptable. I would then update the pull request.

muga:
@ehaupt @hiroyuki-sato thanks! I was missing it. `date -Iseconds` doesn't work on my mac book unfortunately. Can you consider to replace it with the fix value that @hiroyuki-sato suggested above if you don't have strong reasons about your workaround? :-)

ehaupt:
Hello @muga, thanks again for your feedback. I wasn't aware that MacOS' ```date``` lacks a ```-Iseconds``` option. It's probably best as (@hiroyuki-sato suggested) to hardcode the string. It will probably be self-explanatory.

I've updated the pull request accordingly.

muga:
@ehaupt thank you for considering about it. looks good!

Issue #974, opened 2019-02-08
Export values in py tasks with sub param


jaymed:
I'm trying to call a python task with a parameter that contain a list of dictionary values by using `_export` variables. The variables used to define the values in the list of dictionaries are not being interpolated.

Please consider the following example. In this case I would like `arg2` to be an array of dictionary values. In this case the `arg1` param evaluates correctly but the value of `subarg2` contains the literal definition of `relevant_date` instead of the interpolated value.  

```
timezone: UTC
_export:
  relevant_date: ${moment(session_time).utc().format('YYYYMMDD')}

+setup:
  py>: tasks.MyWorkflow.step1
  arg1: ${relevant_date}
  arg2:
    subarg2: ${relevant_date}
```
  
The output looks like this:
```
20190207
{u'subarg2': u"${moment(session_time).utc().format('YYYYMMDD')}"}
```

The sample python task looks like this:

```
import digdag

class MyWorkflow(object):
    def step1(self, arg1, arg2):
        print (arg1)
        print (arg2)

```
Please help me understand how I can properly interpolate the value of `subarg2`.

hiroyuki-sato:
Hello, @jaymed 

You can't use `${}` syntax in `_export` part in the current implementation.
I'll add more detail later.

chezou:
seems like same issue as https://github.com/treasure-data/digdag/issues/678 ?

jaymed:
> seems like same issue as #678 or #862 ?

Thanks, @chezou. It's exactly like #678. I'll have to come up with a different way to pass nested parameters to my python function.

I'm also finding that secrets can't be evaluated as `py` parameters, even if they aren't nested.

```
+setup:
  py>: tasks.MyWorkflow.step1
  arg1: ${secret:password}
```

Without the ability to pass in dates from `_export` data or secrets to `py` operators, I'm really limited in what I can do with `digdag`.

Issue #678, opened 2017-11-01
A moment.js parameter which defined in export part,  cannot reference properly.


hiroyuki-sato:
* Digdag 0.9.20

The `${nested.param2}` parameter can't extract moment.js result if it reference `_export` value. 

## 1. Test code

```yaml
_export:
  now: ${moment(session_time).utc().format('YYYY-MM-DD HH:mm:ss Z')}

+export_test:
  sh>: env
  param1: ${now}
  nested:
    param2: ${now}
```

### 1.1 Expected behavior

```
param1=2017-10-31 00:00:00 +00:00
nested={"param2":"2017-10-31 00:00:00 +00:00"}
```

### 1.2 Actual behavior

```
param1=2017-10-31 00:00:00 +00:00
nested={"param2":"${moment(session_time).utc().format('YYYY-MM-DD HH:mm:ss Z')}"} # <-- Can't extract
```

## 2. Test code2

It worked as I expected if I write moment.js script directly on nested.params2 .

```yaml
_export:
  now: ${moment(session_time).utc().format('YYYY-MM-DD HH:mm:ss Z')}

+export_test:
  sh>: env
  param1: ${now}
  nested:
    param2: ${now}

+param_test:
  sh>: env
  param1: ${moment(session_time).utc().format('YYYY-MM-DD HH:mm:ss Z')}
  nested:
    param2: ${moment(session_time).utc().format('YYYY-MM-DD HH:mm:ss Z')}
```

export_test result

```
param1=2017-10-31 00:00:00 +00:00
nested={"param2":"${moment(session_time).utc().format('YYYY-MM-DD HH:mm:ss Z')}"}
```

param_test result.

```
param1=2017-10-31 00:00:00 +00:00
nested={"param2":"2017-10-31 00:00:00 +00:00"}
```

## 3. Reference

* [Twitter](https://twitter.com/kazukgw/status/925018998307045378)

PR #1077, opened 2019-04-08
[WIP] Fix ConfigEvalEngine.evalObjectRecursive().

yoyama:
Please NOT merge yet.
I am still investigating. And hope to check by CI.

If nested, no information provided, failed to substitute the variable.
This is related to #974, #678.


muga:
@yoyama  This is good improvement 👍What do you think that we can include this bug fix in v0.9? Do you have any concerns?

yoyama:
@muga Thanks for your pointing out. It is important point. 
The compatibility is broken when the workflows use ${...} in nested data structures.
If users provide such as data as argument, it will not work.
But I don't know how often happen such case.

yoyama:
It is old and I will close it. I may re-open in the future if needed.

PR #1082, opened 2019-04-12
Adding http_call operator

frsyuki:
This new `http_call>` operator is useful when a web application wants to
use digdag as its background execution engine and the web application
wants to manage parameters for the execution.

Here is example usage:

1. Workflow calls web app using http_call operator.
2. Web app generates a workflow with some parameters embedded.
3. http_call operator runs the returned workflow as subtasks.
4. Optionally, the web app appends http operator at the end of the
generated workflow so that web app receives a callback.


muga:
I'll merge this PR after I will update README for the operator. My concern was that the naming would make users confusing. But it would be mitigated with the documentation improvement.

PR #1090, opened 2019-04-23
Fix to support digdag log url-encoded string

### LINKED ISSUE #837: The log is not displayed contented in url encoded character in task name prefix
The log is not displayed contented in url encoded character like `%7B` in task name prefix.

Like this command.
```
% digdag log 1 +hoge+repeat^sub+for-0=order=0=%7B
```

hiroyuki-sat summarized this issue.
https://github.com/hiroyuki-sato/digdag-support/tree/master/twitter_digdag_log

katsuyan:
FIx #837

And I have one question.
Why is task name url-encoded ?

### DIFF
--- a/digdag-client/src/main/java/io/digdag/client/DigdagClient.java
+++ b/digdag-client/src/main/java/io/digdag/client/DigdagClient.java
@@ -66,6 +66,8 @@ import javax.ws.rs.core.UriBuilder;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.UnsupportedEncodingException;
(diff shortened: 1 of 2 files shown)


muga:
@katsuyan Thank you for fixing! Your change looks good. May I ask you to try to make unit tests for this change? `client.DigdagClientTest` would be helped for writing your tests.

ref. digdag-client/src/test/java/io/digdag/client/DigdagClientTest.java

katsuyan:
Added tests.

muga:
Thanks @katsuyan, the change looks very good 👍

PR #1095, opened 2019-05-08
Add job_retry parameter to td query operators

kamikaseda:
`td>`, `td_wait>`, `td_wait_table>` and `td_for_each>` operators support `job_retry` parameter which  is used for count of retry upon failure.
https://github.com/treasure-data/digdag/search?q=job_retry&unscoped_q=job_retry

It's better to be explained in documentation.

### DIFF
--- a/digdag-docs/src/operators/td.md
+++ b/digdag-docs/src/operators/td.md
@@ -154,9 +154,8 @@
 
   Set automatic job retry count (default: 0).
 
-.. note::
+  We recommend that you not set retry count over 10. If the job is not succeessful less than 10 times retry, it needs some fix a cause of failure.
 
-    We recommend that you not set retry count over 10. If the job is not succeessful less than 10 times retry, it needs some fix a cause of failure.
 
 * **result_connection**: NAME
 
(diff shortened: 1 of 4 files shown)


yoyama:
@kamikaseda Thanks for your document improvements. Basically it looks like good.
But I can't confirm maximum retry count is really 10. Please wait for my confirmation.

kamikaseda:
@yoyama Yes, you are correct. There is no limitation. But our TD Console has a list of retry count as bellow. So I though it's better to write 0 to 10 as recommended values. (I think 10 is enough basically....)
https://t.gyazo.com/teams/treasure-data/a27125366981c706eb28acc57a341f02.png

yoyama:
@kamikaseda Oh, I understand. But even though  retry count is not limited actually, this explanation may be confused.
How about remove "From `0` to `10` times ," and add another explanation which we recommend retry number should be less than 10?

kamikaseda:
@yoyama Sounds good to me. I've modified that point.

yoyama:
@kamikaseda Thanks to update. Could you confirm the ".. note::" will be displayed as intended?

kamikaseda:
@yoyama Surely, ".. note::" didn't work in md. I've modified to simple comment.

Issue #1079, opened 2019-04-09
Authenticator as a System Plugin


pdelagrave:
Hi,

We need to integrate Digdag within a system using oAuth2.

Before starting working on a PR we'd like to have feedback on our plans:
1. Making the Authenticator interface a system plugin and move it to the SPIs.
2. Make the JwtAuthenticator the default plugin to keep things as they are.
3. Create an oAuth2 "Authenticator" plugin that could be enabled and configured by properties.
4. Make the WebUI able to do oAuth2 Implicit Grant auth when configured to do so.

1 and 2 would be one PR while 3 and 4 another one.

Let me know of any issue with this plan.

Thanks,

Pierre

TakahikoKawasaki:
Some notes.

"3.1.2. Implicit Grant" in "[OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/draft-ietf-oauth-security-topics/?include_text=1)" says as follows.

> _In order to avoid these issues, clients **SHOULD NOT use the implicit grant** (response type "token") or any other response type issuing access tokens in the authorization response, such as "token id_token" and "code token id_token", unless the issued access tokens are sender-constrained and access token injection in the authorization response is prevented._

OAuth should not be used for authentication. OpenID Connect is a better option for the purpose. See "[The problem with OAuth for Authentication](http://www.thread-safe.com/2012/01/problem-with-oauth-for-authentication.html)".

muga:
Hi @pdelagrave, I'm not 100% sure what you want to design and impl. But at least for your 1st and 2nd items, it's ok for us to make extension point for authenticator as system plugin. And can you take a look @TakahikoKawasaki's comment and any thoughts? 

@TakahikoKawasaki thank you for your comment.

pdelagrave:
@TakahikoKawasaki 
> Some notes.
> 
> "3.1.2. Implicit Grant" in "[OAuth 2.0 Security Best Current Practice](https://datatracker.ietf.org/doc/draft-ietf-oauth-security-topics/?include_text=1)" says as follows.
> 
> > _In order to avoid these issues, clients **SHOULD NOT use the implicit grant** (response type "token") or any other response type issuing access tokens in the authorization response, such as "token id_token" and "code token id_token", unless the issued access tokens are sender-constrained and access token injection in the authorization response is prevented._
> 
[...]

PR #1098, opened 2019-05-08
Issue #1079: Moved the Authenticator interface to the SPI module.

pdelagrave:
Content:
- Moved the Authenticator interface to the SPI module.
- Made the Authenticator implementations to be provided as system plugins.
- Refactored the JwtAuthenticator into a plugin and made it the default one to keep everything backward compatible.

Notes:
- Didn't move `io.digdag.server.AuthenticatorTest` to the SPI module as there's no tests in there. Is it worth creating a test directory in SPI for that test class?

- It's currently impossible to load only a specific implementation of a system plugin if there's more than one implemention of the same interface. To get an instance of a specific implementation, `io.digdag.core.plugin.PluginSet` requires to load them all first. While not a big deal for auth at the moment, that's not ideal. Ideally we'd only load the plugin the user wants based on the configuration.
This would require its own refactor (for example `io.digdag.core.plugin.PluginSet` not to getInstance() on the plugins as the only option).
This has at least 2 drawbacks:
  * Wasting memory.
  * The additional code required by the plugins to expect a missing configuration: a user will configure one plugin and not the other one, but more than one plugin might be in the classpath and will be loaded anyway.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/Server.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/Server.java
@@ -133,7 +133,7 @@ public class Server
         Properties props = buildServerProperties();
         ConfigElement ce = PropertyUtils.toConfigElement(props);
         ServerConfig serverConfig = ServerConfig.convertFrom(ce);
-        ServerBootstrap.start(new ServerBootstrap(version, serverConfig));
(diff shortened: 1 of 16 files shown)


muga:
@pdelagrave thanks! I will take a look

PR #1099, opened 2019-05-09
Basic auth authenticator

pdelagrave:
Added a Basic Auth Authenticator plugin for the server.
Added a --basic-auth CLI client option.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/BasicAuthParameterValidator.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/BasicAuthParameterValidator.java
@@ -1,0 +1,19 @@
+package io.digdag.cli;
+
+import com.beust.jcommander.IParameterValidator;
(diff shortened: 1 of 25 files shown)


muga:
@pdelagrave I will take a look this and https://github.com/treasure-data/digdag/pull/1098

pdelagrave:
Right, that's a basic Basic Auth implemention with only one user supported.
* Meant as a quick way to secure the endpoints rather than a production grade solution
* Improvement upon having only the `JwtAuthenticator` option for the user to choose from
* Serves as a second example for using the new `Authenticator` SPI

yoyama:
@pdelagrave  Thanks for your explanation. No problem.
We think #1098 and #1099 are slightly large modification and separate from other PRs.
So We will merge #1098 and #1099 to master after 0.9.38 will be released in next week.
And will release as 0.9.39.
Sorry for our late progress but please wait for a while.

PR #1101, opened 2019-05-13
[v0.10] Not pass system environment variables to command executor implicitly

muga:
In v0.10, we will introduce new command executor SPI and a few implementations of remote command executors, which execute code by scripting operators on remote environment from digdag-server. system environment variable passed from digdag-server don't make sense for code. We'd better not pass them to command executor implicitly. "_env" still enables passing them to the code.

This change may break the backward compatibility of the behavior of DockerCommandExecutor. Because v0.10's DockerCommandExecutor will not pass system envs to code implicitly. Instead SimpleCommandExecutor automatically and implicitly passes them to code. That same behavior as current.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/PyOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/PyOperatorFactory.java
@@ -197,7 +197,6 @@ public class PyOperatorFactory
             }
 
             final Map<String, String> environments = Maps.newHashMap();
-            environments.putAll(System.getenv());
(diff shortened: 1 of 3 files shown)


muga:
Yes, it's that line. It looks like `ProcessBuilder pb` inherits local environment variables by default. Please check the result of `pb.environment()` method call. `DockerCommandExecutor` also uses `pb.environment()` but `docker` command doesn't inherit them even though the user doesn't specify `-e` options explicitly.

yoyama:
@muga Thanks! No problem.

muga:
Thanks, merging..

PR #1102, opened 2019-05-14
[Doc] Var defined in _export is not overwritable by --param

kamikaseda:
I received inquiries about this usage some times. It's better to be written in doc.


muga:
Thanks, the change itself looks good. Please give me one sec to understand the behavior correctly before merging.

yoyama:
I confirmed it is correct.
```
_export:
  param1: aaaa

+task1:
  echo>: param1 ${param1}
  
+task2:
  echo>: param2 ${param2}
```
If i executed above workflows as follows, param1 is export value (aaaa).

```
digdag start test1 dig1 --session now --param param1=bbbb,param2=cccc
```

muga:
@yoyama thank you for taking!

PR #1105, opened 2019-05-15
[doc] secret key generation example

hiroyuki-sato:
Maybe, many users don't know how to generate a secret key.
This PR adds an example of how to generate a secret key.


hiroyuki-sato:
Thank you!

PR #1107, opened 2019-05-17
AccessController: introduce spi.ac.SecretTarget

muga:
This PR introduces "spi.ac.SecretTarget" in AccessController SPI. We will introduce AccessController and the targets SPI in v0.10. "spi.ac.SecretTarget" will be one of the targets.


muga:
@yoyama please take a look when you get a chance.

PR #1111, opened 2019-05-17
Minor fix to remote comments of log endpoints that was deleted on #1110

muga:
Follow up #1110.


muga:
Thanks. I will merge it after CI is passed.

PR #1112, opened 2019-05-17
Add project_id to spi.ac.ProjectTarget

muga:
This PR adds project_id to spi.ac.ProjectTarget in AccessController SPI introduced by v0.10. The id information would be useful.


muga:
@yoyama thanks. will merge it after CI job is passed.

PR #1113, opened 2019-05-17
Introduce spi.ac.{Attempt,Schedule}Target as AccessController SPI

muga:
This PR introduces "spi.ac.AttemptTarget" and "ScheduleTarget" in AccessController SPI. "WorkflowTarget" is used on the attempt kill endpoint to check the kill permission check. But fine-grain information like session_id and attempt_id would be helpful for the permission check for example.

It assumes https://github.com/treasure-data/digdag/pull/1112.


muga:
@yoyama Thank you for taking a look. will merge after CI is passed.

Issue #901, opened 2018-10-18
[feature-request] Use Custom CommandExecutors


muga:
Some users request the option for using their custom CommandExecutor to execute command operator e.g. `py>`, `rb>`. They want to execute the operators on different place from Digdag where are running. Currently we don't have any extension point to load custom CommandExecutor and specify it for command operators.

civitaspo:
I wrote Scripting Operators without custom CommandExecutors. > https://github.com/civitaspo/digdag-operator-ecs_task#experimental-scripting-operators
This is what I wanted to do. When the CommandExecutor can be implemented, we will reimplement these Scripting Operators.

PR #1127, opened 2019-06-02
Introduce EcsCommandExecutor

muga:
This PR introduces EcsCommandExecutor, which enables submitting scripting operators, e.g. "py>", "rb>" and "sh>", and running the script on ECS cluster as ECS tasks. The functionality is based on command executor SPI v2. It will be released in v0.10.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/CommandExecutorModule.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/CommandExecutorModule.java
@@ -4,8 +4,8 @@ import com.google.inject.Binder;
 import com.google.inject.Module;
 import com.google.inject.Scopes;
 import io.digdag.spi.CommandExecutor;
-import io.digdag.standards.command.kubernetes.DefaultKubernetesClientFactory;
-import io.digdag.standards.command.kubernetes.KubernetesClientFactory;
+import io.digdag.standards.command.ecs.DefaultEcsClientFactory;
+import io.digdag.standards.command.ecs.EcsClientFactory;
 
 public class CommandExecutorModule
     implements Module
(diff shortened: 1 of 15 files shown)


digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -0,0 +1,685 @@
...
+                // When RuntimeException is thrown by submitTask method, it will be handled and retried by BaseOperator, which is one of base
+                // classes of operator implementation.
+                final Task runTask = submitTask(commandContext, commandRequest, client, td); // ConfigException, RuntimeException
+                final ObjectNode currentStatus = createCurrentStatus(commandContext, commandRequest, clientConfig, runTask, awsLogs);
+                return createNextCommandStatus(commandContext, client, currentStatus);
+            }
+        }
+        catch (ConfigException e) {
+            logger.info("EcsCommandExecutor is not executed.", e);
+            return docker.run(commandContext, commandRequest); // fall back to DockerCommandExecutor
```
yoyama: I guess that fall back by ConfigException is a bit confused. 
And I have a question about fall back to DockerCommandExecutor is correct? Not SimpleCommandExecutor?

digdag-standards/src/main/java/io/digdag/standards/command/CommandExecutorModule.java
```diff
@@ -4,17 +4,19 @@
 import com.google.inject.Module;
 import com.google.inject.Scopes;
 import io.digdag.spi.CommandExecutor;
-import io.digdag.standards.command.kubernetes.DefaultKubernetesClientFactory;
-import io.digdag.standards.command.kubernetes.KubernetesClientFactory;
```
yoyama: We remove KubernetesCommandExecutor ?

muga: That good point. I don't remove Kubernetes command executor. But I'm now removing Kubernetes' one from the chain of command executor fall back. So, in this PR, we will have the fall back chain: Ecs -> Docker -> Simple.

I think that we'd better to make another mechanism for loading custom command executors and choose some of them like https://github.com/treasure-data/digdag/issues/901 separately.

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
muga: Yes, fall back to Docker, not Simple. so, the comment on my change is correct.

muga:
@yoyama thank you for reviewing.

> So after fixed it, this branch should be checked again. And I guess we need a document after merge.

Thanks, got it. I will take care of that. and documentation as well.

PR #1128, opened 2019-06-03
Fix missing projects command.

yoyama:
Fix missing projects command.


yoyama:
Merge to fix CI errors.

PR #1129, opened 2019-06-03
Missing migrate command

yoyama:
migrate command also disappeared in CLI. So add it.


yoyama:
Merge to fix CI errors.

PR #1130, opened 2019-06-03
Update release procedure

yoyama:
Minor fixes on release procedures.


yoyama:
@szyn Could you check this review if you have a change?
 This is only fix on release documentation.

PR #1134, opened 2019-06-04
Fix Travis CI build for v0_10

muga:
This PR fixes Travis CI build for v0_10 branch

### DIFF
--- a/ci/run_test_docker.sh
+++ b/ci/run_test_docker.sh
@@ -2,6 +2,7 @@
 
 docker run \
 -v /var/run/docker.sock:/var/run/docker.sock \
+-v /tmp:/tmp:rw \
 -w /digdag \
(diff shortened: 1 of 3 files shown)


muga:
@yoyama can you take a look when you get a chance? /tmp volume mount is necessary. because our integration tests are using directories and filse under /tmp as workflow projects. a Docker-in-Docker container in a Docker container uses the parent host's Docker daemon and hence, any volumes that are mounted in the "docker-in-docker" case is still referenced from the host, and not from the Docker container. therefore, the actual path mounted from the Docker container "cannot find the file" in the host.

ref. https://stackoverflow.com/questions/31381322/docker-in-docker-cannot-mount-volume

muga:
thank you for reviewing. merging..

PR #1138, opened 2019-06-06
Introduce http_call.max_response_content_size system config for http_call> operator

muga:
follow-up of https://github.com/treasure-data/digdag/pull/1082.

The size limit should be separated each other. Because those operators make different resource usage. For example, "http>" uses store param to store response content. but "http_call>" not uses store param. The limit for "http_call>" could be relaxed. For now, as the limitation of http response content size, "http_call>" follows and uses maxStoredResponseContentSize in "http>". This is not good idea.


muga:
Thanks! merging..

PR #1145, opened 2019-06-12
Fix logic to look up ECS task definition by tags in DefaultEcsClient

muga:
This PR fixes a logic to look up ECS task definition by tags in DefaultEcsClient.


muga:
thanks! merging..

PR #1146, opened 2019-06-12
Introduce metrics with micrometer. (v0_10)

yoyama:
- Integrate micrometer
- Support JMX (io.digdag domain)
- Customized by Guice module with SPI
- Targets of metrics as follows.
  - API endpoints
  - WorkflowExecutor
  - Agent and OperatorManager
  - Database


yoyama:
@muga I have not yet implemented test and document, but could you check roughly the code?
Following image is sample JMX metrics (not all).
Currently around 150 metrics are created. 
image

digdag-docs/src/metrics.md
```diff
@@ -0,0 +1,19 @@
+
+# Digdag metrics (experimental)
+
+Digdag support JMX and provided a few metrics. But it is not enough to monitor its performance and stability.
+To improve it, new metrics framework _digdag metrics_ has been introduced.
+These new metrics are provided via JMX.
+
+_Digdag metrics_ integrates micrometer
+There are meny metrics in _digdag merics_.
```
hiroyuki-sato: meny -> many
merics -> metrics?

yoyama: @hiroyuki-sato Thanks for your pointing out. I will fix them later.

digdag-server/src/main/java/io/digdag/server/metrics/DigdagTimedMethodInterceptor.java
```diff
@@ -0,0 +1,85 @@
...
+        String value = mkValue(timed, invocation);
+        String metricsName = metrics.mkMetricsName(category, value);
+        Timer.Sample sample = metrics.timerStart(category);
+        try {
+            return invocation.proceed();
+        } catch (Exception ex) {
+            throw ex;
+        } finally {
+            try {
+                Tags taskTags = timed.taskRequest() ? getTaskRequest(invocation) : Tags.empty();
```
muga: This method call could be invoked before timerStart call for elapsed time accuracy?

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
```diff
@@ -441,6 +446,7 @@ void expireLocks()
         catch (Throwable t) {
             logger.error("An uncaught exception is ignored. This lock expiration thread will be restarted.", t);
             errorReporter.reportUncaughtError(t);
+            metrics.increment(Category.DEFAULT, "uncaughtErrors");
```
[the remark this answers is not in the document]
yoyama: Thanks for your pointing out.
In JMX, DEFAULT is mapping to the domain "io.digdag" and used for "misc" metrics.
I set DEFAULT because I thought it was different from DB access.
But it is still under database package name space.
I will change to "DB".

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -317,7 +334,7 @@ protected TaskResult callExecutor(Path projectPath, String type, TaskRequest mer
 
         return operator.run();
     }
-
+
```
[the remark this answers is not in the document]
yoyama: Thanks! will fix it.

digdag-server/src/main/java/io/digdag/server/metrics/DigdagTimedMethodInterceptor.java
yoyama: Sorry, I may misunderstand your question. 
But timerStart is always call in L.35 as follows.

```
Timer.Sample sample = metrics.timerStart(category);
        try {
            return invocation.proceed();
```
So I guess the L.42 method call is never invoked before timerStart.

yoyama:
@muga Thanks for your comment.
I will try to fix the PR with your comments.

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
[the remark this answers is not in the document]
yoyama: Fixed it.

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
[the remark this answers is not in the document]
yoyama: Fixed it.

yoyama:
@muga Could you confirm again after CI pass. I updated as follows.

- DigdagEmbed set empty DigagaMetrics as default
- So remove injection of DigdagaMetrics from CLI and tests.
- Add comment to ErrorReporter class on deprecation in future.

I didn't fix your following comment. 
> Curious in ConfigEvalEngine#eval related metrics. Not only elapsed time but 2 arguments size as well :-)
I think that calculating the size of argument may affect to performance if argument is too big and it may also cause abort because of OutOfMemory.
If you have any good idea to calculate the arguments size safely and effectively, please give me advice.

digdag-core/src/main/java/io/digdag/core/database/DatabaseTaskQueueServer.java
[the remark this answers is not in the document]
muga: thanks 👍

digdag-server/src/main/java/io/digdag/server/metrics/DigdagTimedMethodInterceptor.java
muga: Sorry, my comment made your confusing. `getTaskRequest(invocation)` method call looks like heavy because it has for loop and instanceof operator. If the method could be called before `timerStart` method call, the metrics could be more accurate. But not so big deal I think.

muga:
@yoyama thanks! i didn't expect that they were done on this PR, but great job 👍 

> I think that calculating the size of argument may affect to performance if argument is too big and it may also cause abort because of OutOfMemory. If you have any good idea to calculate the arguments size safely and effectively, please give me advice.

No I don't have good idea, sorry. Your comment makes sense. we need to avoid the bottleneck for taking metrics. Please ignore my comment. 

Good to go? or wait for https://github.com/treasure-data/digdag/pull/1146#discussion_r309019182 Both of them are fine to me.

digdag-server/src/main/java/io/digdag/server/metrics/DigdagTimedMethodInterceptor.java
yoyama: Thanks! Got it. I will fix it before merge.

Issue #1131, opened 2019-06-04
Operations within "for_each>:" operator executed twice when retried


to-lz1:
DIgdag version: 0.9.37

We have `_retry` parameter and `for_each` operator in our workflow.
When workflow failed, workflow itself successfully restarted.
But each operation in `_do` executed twice, even if the error came from outside of `for_each` operator.

this .dig file will reproduce the issue.

```
timezone: Asia/Tokyo

_retry: 2

+test:
  _export:
    counts: [
      0, 1, 2
    ]

  +echo:
    for_each>:
      count: ${counts}
    _do:
      echo>: ${count}  

  +failure:
    for_each>:
      count: ${counts}
    _do:
      sh>: exit ${count}
```

```
2019-06-04 14:41:25 +0900: Digdag v0.9.37
2019-06-04 14:41:36 +0900 [WARN] (main): Using a new session time 2019-06-04T00:00:00+09:00.
2019-06-04 14:41:36 +0900 [INFO] (main): Starting a new session project id=1 workflow name=z session_time=2019-06-04T00:00:00+09:00
2019-06-04 14:41:37 +0900 [INFO] (0016@[0:default]+z+test+echo): for_each>: {count=[0,1,2]}
2019-06-04 14:41:37 +0900 [INFO] (0016@[0:default]+z+test+echo^sub+for-0=count=0=0): echo>: 0
0
2019-06-04 14:41:37 +0900 [INFO] (0016@[0:default]+z+test+echo^sub+for-0=count=1=1): echo>: 1
1
2019-06-04 14:41:37 +0900 [INFO] (0016@[0:default]+z+test+echo^sub+for-0=count=2=2): echo>: 2
2
2019-06-04 14:41:37 +0900 [INFO] (0016@[0:default]+z+test+failure): for_each>: {count=[0,1,2]}
2019-06-04 14:41:38 +0900 [INFO] (0016@[0:default]+z+test+failure^sub+for-0=count=0=0): sh>: exit 0
2019-06-04 14:41:38 +0900 [INFO] (0016@[0:default]+z+test+failure^sub+for-0=count=1=1): sh>: exit 1
2019-06-04 14:41:38 +0900 [ERROR] (0016@[0:default]+z+test+failure^sub+for-0=count=1=1): Task failed with unexpected error: Command failed with code 1
java.lang.RuntimeException: Command failed with code 1
[...]

PR #928, opened 2018-12-18
Fix #849. Add INITIAL_TASK flag when sub task generated.

yoyama:
To fix #849 I would like to set INITIAL_TASK flag when add sub task.
When call operator is processed, WorkFlowExecutor.taskSucceeded() add it as sub task. 
But sub task is not set as initial task, it is ignored when retry process happened.
When WorkFlowExecutor.retryGroupingTask() is called, only initial tasks are copied for retry.

PR #1153, opened 2019-06-18
[WIP] Fix retry degradtion #1131

yoyama:
The cause is PR #928 which fixed the retry issue in callee by call operator.
In that PR, INITIAL_TASK state flag is always set to generated sub tasks.
As result, generated sub tasks by operators such as for_each, if and other operator run twice.

ex) consider about operator A which  generates task subA.
If retry happens,  both A and subA are target of the retry because INITIAL_TASK state_flags.
The task A generate subA' again and executed.  subA is also retried.
So one retry generates two subA task.

This is cause of the issue.

To fix it, INITIAL_TASK is only added to the sub task of call operator.
In other operators, INITIAL_TASK is not set. (Same as revert #928)
Call operator's sub tasks still require INITIAL_TAKS because of the retry in the callee.
But same duplicate tasks also happen in call operator.
So for call operator, additional process are added.
For call operator task, after generated sub tasks are added, INITIAL_TASK of the call> task are removed
to avoid retry. 
And the parent task of generated sub tasks are changed from call> task to its parent task.
As result the call> task are eliminated from task chain and remove from retry process.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/CallOperatorFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/CallOperatorFactory.java
@@ -78,6 +78,7 @@ public class CallOperatorFactory
 
             return TaskResult.defaultBuilder(request)
                 .subtaskConfig(def)
+                .callSubTask(true)
                 .build();
         }
     }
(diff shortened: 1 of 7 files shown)


yoyama:
Fixed by another way in v0_10 branch. So closed.

PR #1162, opened 2019-06-25
Stop using imp module

chezou:
imp module [has been deprecated since Python 3.4](https://docs.python.org/3/library/imp.html), it should be replaced with other methods. While importlib.util.module_from_spec is preferred over types.ModuleType, module_from_spec requires dramatic change so this change uses types.ModuleType.

This PR removes the following warning:

```
.digdag/tmp/digdag-py-10-1406823779138478784/runner.py:4: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
  import imp
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,8 @@
 */build/
 */out/
 *.pyc
+venv*
+.venv*
 /dist/*.tar.gz
 /config/test_postgresql.properties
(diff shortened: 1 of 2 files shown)


muga:
thanks! this is great. i will take a look soon.

PR #1163, opened 2019-06-26
Remove duplicates when building resumed tasks map

eddsteel:
When resuming a failed workflow, many tasks with the same full name
may be added, if many retries of the same task failed. This will throw
an uncaught exception when many tasks are added with the same
key (full name), to the map of tasks to resume.

This fixes the problem by discarding any duplicate tasks at the map
building stage. The assumption is that any tasks with the same name
are retries of the same task, so we only want one.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/workflow/TaskControl.java
+++ b/digdag-core/src/main/java/io/digdag/core/workflow/TaskControl.java
@@ -140,7 +140,6 @@ public class TaskControl
                 .transform(index -> indexToId.get(index))
                 .or(parentTaskId);
             long id;
-
             if (resumingTaskMap.containsKey(wt.getFullName())) {
                 id = store.addResumedSubtask(attemptId, parentId,
                         wt.getTaskType(),
--- a/digdag-core/src/main/java/io/digdag/core/workflow/TaskControl.java
+++ b/digdag-core/src/main/java/io/digdag/core/workflow/TaskControl.java
@@ -121,7 +121,7 @@ public class TaskControl
 
         Map<String, ResumingTask> resumingTaskMap = resumingTasks
             .stream()
-            .collect(Collectors.toMap(t -> t.getFullName(), t -> t));
+            .collect(Collectors.toMap(t -> t.getFullName(), t -> t, (a, b) -> a));
 
         boolean firstTask = true;
         for (WorkflowTask wt : tasks) {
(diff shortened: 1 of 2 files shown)


digdag-core/src/main/java/io/digdag/core/workflow/TaskControl.java
```diff
@@ -121,7 +121,7 @@ private static long addTasks(TaskControlStore store,
 
         Map<String, ResumingTask> resumingTaskMap = resumingTasks
             .stream()
-            .collect(Collectors.toMap(t -> t.getFullName(), t -> t));
+            .collect(Collectors.toMap(t -> t.getFullName(), t -> t, (a, b) -> a));
```
muga: Looks good! May I ask you to add comment in the source code about why we need to add this fix. It would be helpful for future maintenance. It's fine to copy and paste from the description in your PR.

yoyama:
@muga @eddsteel I would like to confirm this PR later. 
Because the issue described in this PR is related to  #1131.
And I explained the background of issue in #1153 . Please check it.
I am afraid that if we deduplicate task without considering task dependency, task tree may break or not.

muga:
ok, got it. please take your time 👍

eddsteel:
> I am afraid that if we deduplicate task without considering task dependency, task tree may break or not.
Ah, interesting. Thanks for the background.

yoyama:
@eddsteel ( @muga ) I check the code around your fix.
I think this PR helps avoiding unexpected error by dup tasks becase of retry are listed in resumingTasks.
But this PR ignores the task tree and it may not be good to resume correctly.

For example, let's consider following workflow

```
+group1:
  _retry: 1
  +task1:
  +task2:
  +task3:
  +task4:
```

This WF does group retry 1 time. So be executed two times in max.

1. In first try, if +task1 and +task2, +task3 succeed and +task4 fail,  then group retry happen.
1. In second try, if +task1 succeed and +task2 fail, then workflow finally fails.

If we try to _retry with resume_ this WF, the resuming tasks are +task1, +task2, +task3. (dup is removed by your fix)
As result, only +task4 will be executed.
But I don't think it is correct behavior. At least in last try (=second try) only +task1 succeed, so resume from +task2 may be correct.
Furthermore this is group retry, so resume from +task1 is correct in my understand.
I guess these original code may not consider group retry.

I tested above behavior with this PR, but it may be my misunderstand.
Please confirm the tasks correctly run after your fix applied.

eddsteel:
Your description of correct behavior sounds right. My expectation was that this fix would only change a task appearing twice to appear once, not change a task appearing once to be removed. The root error is two identical names; I thought that could only happen if the task appears twice.

Would it be possible to create a unit test with your scenario, then define the expected outcome?

PR #1165, opened 2019-07-01
Fix the message of invalid ISO time format error

hiroyuki-sato:
Fixed `Invalid ISO time format` message

Reference: #1164

Before

```
> Invalid ISO time format: %s2016-01-01T01:01:01+0000 (json mapping)
```

After (Remove `%s`)

```
> Invalid ISO time format: 2016-01-01T01:01:01+0000 (json mapping)
```


hiroyuki-sato:
Thank you!

PR #1168, opened 2019-07-05
Prohibit overwrite runtime params

yoyama:
Digdag generates some parameters such as project_id,session_id and others.
These parameters should not be overwritten by --param because it will cause of unstable, security issues.
To do that I change evaluation order of parameters in OperatorManager. And runtime params overwrite user defined ones by --param.


yoyama:
@frsyuki  Thanks for your pointing out. I updated the document.
After CI pass, I will merge it later.

PR #1170, opened 2019-07-08
EcsCommandExecutor: Add getTaskDefinitionTags method to EcsClient

muga:
This PR adds `getTaskDefinitionTags` method in `EcsClient` for extensibility improvement. ECS task tags information could be useful. We'd better to provide a getter method of tags from ECS task definitions.


digdag-standards/src/test/java/io/digdag/standards/command/ecs/DefaultEcsClientTest.java
```diff
@@ -0,0 +1,54 @@
...
+    public void testGetTaskDefinitionTags()
+            throws Exception
+    {
+        final List<Tag> expected = ImmutableList.of(
+                new Tag().withKey("k1").withValue("v1"),
+                new Tag().withKey("k2").withValue("v2")
+        );
+        final ListTagsForResourceResult result = new ListTagsForResourceResult()
+                .withTags(expected);
+        doReturn(result).when(rawEcsClient).listTagsForResource(any(ListTagsForResourceRequest.class));
```
eddsteel: I think you could specify the ARN here, and then verify this was called with the expected parameters at the end of the test. It won't make much difference, though.

muga: Got it, thanks. Fixed https://github.com/treasure-data/digdag/pull/1170/commits/ac6ae52050ed5380c928043d0da8910fb0f86c0a

Issue #1092, opened 2019-04-26
[feature request] ignore file configuration like .gitignore


chezou:
This is a feature request to have a ignore file for a project.

**Expected use case:**
When developing a workflow with Python, Python generates `__pycache__` files. In Python development, it is a popular way to create a virtual environment under the project directory and sometimes it could be pushed to digdag server or could fail to push due to the large size of library dirs.

It is a usual way to develop workflow locally and push to server for production. So I'd be happy if we could have a ignore file configuration.

szyn:
This feature was released in v0.10.3. Please see also here: https://docs.digdag.io/releases/release-0.10.3.html
I will close this issue.

PR #1175, opened 2019-07-11
Add digdagignore

katsuyan:
I implemented this issue:  https://github.com/treasure-data/digdag/issues/1092

I create digdagignore like gitignore. It can ignore files when digdag push.
But, I don't use gitignore syntax. I use glob syntax.

It can ignore file in .digdagignore file like gitignore.
If first character is `!` then it reverse a process.
And, if first character `#` or ` ` then the line is skiped.

For example, you can try digdagignore in digdag project repository.

```
java -jar PathTo/digdag/pkg/digdag-0.9.39-SNAPSHOT.jar push sample --project examples/digdagignore/
2019-07-12 01:39:35 +0900: Digdag v0.9.39-SNAPSHOT
Creating .digdag/tmp/archive-5896468971676972804.tar.gz...
projectPath: PathTo/digdag/examples/digdagignore
  Archiving a.dig
  Skip b.dig
  Archiving a/c/e.txt
  Archiving a/c/d.txt
  Archiving a/d/a1.txt
  Archiving a/d/a2.txt
  Archiving a/d/a3.txt
  Skip a/d/a4.txt
  Skip a/b/b.txt
  Skip a/b/a.txt
Workflows:
  a.dig
Uploaded:
  id: 1
  name: sample
  revision: 42c88991-170e-4e5a-a503-9d4c6d7d82b2
  archive type: db
  project created at: 2019-07-11T16:39:37Z
  revision updated at: 2019-07-11T16:39:37Z

Use `digdag workflows` to show all workflows.
```

The reason why I use glob syntax is that gitignore syntax is complex to implement.
Glob syntax is simple and glob exists in java library.
And We must follow the git version when We use gitignore syntax.

Please comment if you like, and I will add tests and documents.


muga:
Hi @katsuyan, thank you for implementing this feature. I think that the feature really good 👍 I took a look your spec in the description and code roughly. I have the following comments.
- Can you move your change to `ProjectArchive` from `Archiver`? I think that the right place is `ProjectArchive#listFilesRecursively`. In your current implementation, `ignoreFileList` method is called every tar entry copying. it's not efficient for the performance. Moreover we ignore dot files during archiving the project. Your implementation could be integrated with that. If user-defined ignore file doesn't exist, it ignores dot files by default ignore file. It means that user-defined ignore file could override the behavior by default ignore file.
- Glob by raw regex is ok but what do you think that we  use Java based git library like [eclipse JGit](https://github.com/eclipse/jgit)? I'm not sure that we can use JGit for this purpose but, that's good option I think.
What do you think?

katsuyan:
Thank you very much for your kind comment, @muga.

- About the first comment

> It means that user-defined ignore file could override the behavior by default ignore file.

I did not care about the above. I understand that the right place is `ProjectArchive`.

- About the second comment

I didn't know Java based git library. I understand to use it will be easy for future maintenance.
If we can use it for this purpose, I think it's best way.
Thank you for the information about the roadmap.

I try to use JGit once.
Should I close this PR?

yoyama:
@katsuyan I would like to close this PR because of no activity long time.
I think this is a valuable functionality, so we can revisit in the future.

PR #1181, opened 2019-07-17
[doc] Add params option description in require> operator.

### LINKED ISSUE #1180: require> params option behavior
@yoyama, @muga Could you confirm `require>` behavior?

Please see [original post](https://twitter.com/sgswtky/status/1151392328227049472?s=20)

Does the following description is correct?
If so, I'll create PR. Because I don't find any description of this option.

`require>` operator doesn't pass a parameter to child task. If you want to set it, use `params` options like the following.

```yaml
+example:
  require>: child
  params:
    param_name1: ${parent_param_name}
```

[Sample workflow](https://github.com/hiroyuki-sato/digdag-support/tree/master/2019-07-17_twitter_require_params)

hiroyuki-sato:
This PR adds `params` option description in `require>` operator. Fix #1180.

### DIFF
--- a/digdag-docs/src/operators/require.md
+++ b/digdag-docs/src/operators/require.md
@@ -71,3 +71,16 @@
   ignore_failure: true
   ```
 
+* **params**: MAP
+
+  This operator doesn't pass a parameter to another workflow. `params` options set parameters.
+
+  Examples:
+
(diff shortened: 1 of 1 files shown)


hiroyuki-sato:
Thank you!

PR #1061, opened 2019-03-25
Feature add index to session attempts

yoyama:
Add an index to resolve performance issue on session_attempts.

To support _create index concurrently_, re-write DatabaseMigrator.migrate() and support no transaction mode in Migration.

Original migrate() support two mode.
(a) Initial mode : all migrations in a transaction
(b) Upgrade mode : each migration run in separate transaction

New migrate() only support (b). Because of support no transaction mode.
In multiple digdag server cluster, if digdag servers boot simultaneously, migrate() may fail.
To avoid it, migrateWithRetry() is implemented.

PR #1183, opened 2019-07-21
Change log level of Database Migrator

szyn:
This PR is related to #1061. That PR became to show messages about the migration status of the database each time the CLI is running. But these messages seem not always needed for users, so I updated these messages log level from info to debug.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseMigrator.java
@@ -84,9 +84,9 @@ public class DatabaseMigrator
         final int maxRetry = 3;
         for (int i = 0; i < maxRetry; i++) {
             try {
-                logger.info("Database migration started");
+                logger.debug("Database migration started");
                 migrate();
-                logger.info("Database migration successfully finished.");
+                logger.debug("Database migration successfully finished.");
                 return;
             }
             catch (RuntimeException re) {
@@ -122,7 +122,7 @@ public class DatabaseMigrator
             for (Migration m : migrations) {
                 Set<String> appliedSet = getAppliedMigrationNames(handle);
                 if (appliedSet.add(m.getVersion())) {
-                    logger.info("Applying database migration:" + m.getVersion());
+                    logger.debug("Applying database migration:" + m.getVersion());
                     numApplied++;
                     if (m.noTransaction()) {
                         // In no transaction we can't lock schema_migrations table


hiroyuki-sato:
Hello, @szyn 

Thank you for creating this PR.
This is personal opinion. I think log level need to change in specific mode ex. `digdag run`

* This message is useful on server mode (`digdag server`)
* I think user doesn't change log level during migration because they don't know when migration needs. 
* This message is not useful on CLI mode (`digdag run`) because it uses on-memory database and executes migration every time.

yoyama:
@szyn Thanks four your PR.
Could you explain in which situations, modes, commands do you feel it is needless?
In #1137 I had already removed useless migration. 
I understand sometimes migration message is noisy, but I think migration is very important event and it should be informed in production use properly. 
I don't check in detail, but basically I agree with @hiroyuki-sato 's proposal.
What is your opinion?

szyn:
@hiroyuki-sato @yoyama 
Thank you for your comments and for sharing your thoughts.
Yes, I agree with the @hiroyuki-sato opinion. Definitely, `digdag server` running with PostgreSQL shows these messages is useful. 
But I think it doesn't need to show for local and server running in the in-memory (H2) database mode because these modes need to migration at all times. 
How about output these messages when using the driver for PostgreSQL?

hiroyuki-sato:
Hello, @szyn and @yoyama 

How about `database.type = h2`? Does this mode use migration?
I know some users use that mode.

We agree with that `local` doesn't require that message and PostgreSQL requires it.
The difficulty is how to implement it 🤔 

@yoyama Do you have any idea?

yoyama:
I think we should provide migration info in H2 with disk.
@hiroyuki-sato pointed out, if it is used for production, migration info is important for them.
( I don't hope to use H2 for production, but it is another things.)
So if we change log level of migration by DB type, it should be restricted to H2 memory DB.
But I am wonder which use cases do you have in H2 memory DB?
Should we take care of H2 memory DB?
I guess we should change log level in 'run' command (== local mode) but I don't know another situation. @szyn Could you explain?

szyn:
Well, I didn't consider using H2 in production but I understand we also need to take care of it.
>I guess we should change log level in 'run' command (== local mode) but I don't know another situation. 

Sorry, I don't know the other situation 🙇. In any case, I agree with this opinion.

szyn:
I will close this PR because this is resolved by #1248.

PR #1185, opened 2019-07-22
Upgrade React from v15.4.2 to v16.8

szyn:
I upgraded React from v15.4.2 to v16.8 and also updated related packages. 
There are two important points. First, update to use prop-types package because of React v15.5 PropTypes is deprecated. Second, React-router v4 and up is a complete rewrite, so need to upgrade the source code related to routing.

References:
- https://reactjs.org/warnings/dont-call-proptypes.html
- https://github.com/ReactTraining/react-router/blob/432dc9cf2344c772ab9f6379998aa7d74c1d43de/packages/react-router/docs/guides/migrating.md


szyn:
Thank you!

PR #1186, opened 2019-07-22
Migrate to Bootstrap v4

szyn:
This PR is a change to introduce Bootstrap v4, which was already installed but not used. 
https://github.com/treasure-data/digdag/blob/5a7acce73c07c289f1980fa4118c979a8b183b04/digdag-ui/package.json#L20

Some UI adjustment is also included. As for the icon, v4 dropped the Glyphicons icon font, so I would like to propose using Font Awesome as instead.
Ref. https://getbootstrap.com/docs/4.0/migration/

### DIFF
--- a/digdag-ui/.bootstraprc
+++ b/digdag-ui/.bootstraprc
@@ -1,6 +1,6 @@
 loglevel: warn
 
-bootstrapVersion: 3
+bootstrapVersion: 4
 
 styleLoaders:
   - css-loader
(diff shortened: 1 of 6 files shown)


yoyama:
@szyn Unfortunately some files are conflicted because of merge your previous PR.
Could you fix them?

szyn:
@yoyama Thank you for pointing it out. I resolved the conflict. 
Could you take a look at this again?

szyn:
@yoyama Thank you for pointing it out. I resolved the conflict.
Could you take a look at this again?

yoyama:
@szyn Thank you for your PR.
I built and tested. It looks like be unstable.
Sometime I faced empty page.
For example, choose /projects then click a workflow link in "Session" section, it became empty page.
Could you check it?

szyn:
@yoyama Oh, good catch. And sorry about that. 
I need to update at #1185, but I overlooked that. I added the follow-up commit (b04e57a).

PR #1199, opened 2019-08-07
Add option to copy symbolic links pointing outside of project directory

frsyuki:
`digdag push` fails if the project includes a symbolic link that points
outside of the project directory. This is hard to avoid because the
archive package must be self-contained but such files pointed by the
symlinks won't be included in the archive.

This new option provides a solution to it. If the option is given, the
command copies files or directories instead of throwing an exception.

It might make sense to make this behavior default (with warning messages)
at a future major version upgrade.

This is especially useful to avoid a workaround as mentioned at
https://qiita.com/sonots/items/9cf643279da62fce5cdf


yoyama:
@sada It looks like all CI failed. Could you confirm them if you have a chance.

yoyama:
Rebase to fix CI errors.

PR #1206, opened 2019-08-19
Configurable digdag metrics

yoyama:
This PR update digag-metrics as follows

- Digag-metrics is disabled as default
- Configurable by server config
- Each category is also configured enable/disable.

### DIFF
--- a/digdag-server/src/main/java/io/digdag/server/metrics/DigdagMetricsConfig.java
+++ b/digdag-server/src/main/java/io/digdag/server/metrics/DigdagMetricsConfig.java
@@ -27,7 +27,7 @@ public class DigdagMetricsConfig
     {
         // fetch enabled plugin server.metrics.enable = jmx,...
         List<String> keys = config
-                .getOptional("server.metrics.enable", String.class)
+                .getOptional("metrics.enable", String.class)
                 .transform((s) ->
                         Arrays.asList(s.split(","))
                                 .stream()
--- a/digdag-server/src/main/java/io/digdag/server/metrics/DigdagMetricsConfig.java
+++ b/digdag-server/src/main/java/io/digdag/server/metrics/DigdagMetricsConfig.java
@@ -3,7 +3,7 @@ package io.digdag.server.metrics;
 import com.google.common.base.Optional;
 import io.digdag.client.config.Config;
 import io.digdag.client.config.ConfigException;
-import io.digdag.server.metrics.jmx.JmxPluginConfig;
+import io.digdag.server.metrics.jmx.JmxMonitorSystemConfig;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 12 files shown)


digdag-server/src/main/java/io/digdag/server/metrics/DigdagMetricsConfig.java
```diff
@@ -0,0 +1,62 @@
...
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class DigdagMetricsConfig
+{
+    Logger logger = LoggerFactory.getLogger(DigdagMetricsConfig.class);
```
muga: minor: should we make "private"?

digdag-server/src/main/java/io/digdag/server/metrics/DigdagMetricsConfig.java
```diff
@@ -0,0 +1,62 @@
...
+    Map<String, DigdagMetricsPluginConfig> plugins = new HashMap<>();
```
muga: ditto

digdag-server/src/main/java/io/digdag/server/metrics/DigdagMetricsConfig.java
muga: for the field.

muga:
Overall sounds good. Thank you for introducing that 👍 As we discussed in-person, let me summarize here, please confirm when you get a chance. 
- We'd better to revisit the appropriate place where "DigdagMetricsConfig" object is initialized now or later. In the current impl., that is in ServerConfig.java and it works.
- The current property name is "server.metrics.enable". But "metrics.enable" is more flexible in the future? for example, users may want to collect some metrics by local mode.
- The name of "DigdagMetricsPluginConfig" might make confusing because the "Plugin" comes from micrometer "plugin".
- In the future, will we allow customizing metrics collection by others (non "jmx" plugin) or not?

yoyama:
@muga Thanks for your reviews.

> We'd better to revisit the appropriate place where "DigdagMetricsConfig" object is initialized now or later. In the current impl., that is in ServerConfig.java and it works.

I removed DigdagMetricsConfig from ServerConfig.

>The current property name is "server.metrics.enable". But "metrics.enable" is more flexible in the future? for example, users may want to collect some metrics by local mode.

I fixed it.

>The name of "DigdagMetricsPluginConfig" might make confusing because the "Plugin" comes from micrometer "plugin".

Changed to MonitorSystemConfig. JMX, Datadog and others are called _monitor system_ in Micrometer site. So I named it.

>In the future, will we allow customizing metrics collection by others (non "jmx" plugin) or not?

Yes. 
Currently, DigdagEmbed and override some class is one of the way to do it.
We should support system plugin in future.

PR #1207, opened 2019-08-20
Improve exception handling in WorkflowExecutor.

yoyama:
This is just experimental implementation to confirm in CI.
WorkflowExecutor sometimes faces infinite loop with unexpected Exception handling.
To improve it, this PR catch all exception in runWhile()

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
+++ b/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
@@ -512,15 +512,16 @@ public class WorkflowExecutor
                 enqueueReadyTasks(queuer);
 
                 /**
-                 *  propagateSessionArchive() should always call.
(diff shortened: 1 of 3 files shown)


digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
```diff
@@ -539,6 +540,18 @@ public void runWhile(BooleanSupplier cond)
         }
     }
 
+    protected <R> R catching(Supplier<R> func, R defaultValue, String errorMessag)
+    {
+        try {
+            return func.get();
+        }
+        catch (Exception e) {
+            logger.warn(errorMessag);
```
muga: we'd better to collect metrics for that.

yoyama: Thanks.  I added it.

PR #1209, opened 2019-08-21
Split task log by size.

### LINKED ISSUE #1193: log doesn't refresh until task end
I'm running digdag server with the following script
`nohup digdag server -b 0.0.0.0 -n 9876 -o ~/digdag/db -O ~/digdag/log &`
while my workflow has a single task (long time)

But I noticed the digdag UI doesn't refresh log for the task until it's completed. Could I see the real time log from the web just like less +F command?
屏幕快照 2019-07-30 下午9 55 35

yoyama:
This PR is the purpose to resolve #1193 and replace PR #700. 

This PR splits task log by the configured size.
So long running task can also show the log by **digdag log -f**.
If server admin configured the size to small one, user can show the task log in near real time.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/log/BufferedRemoteTaskLogger.java
+++ b/digdag-core/src/main/java/io/digdag/core/log/BufferedRemoteTaskLogger.java
@@ -1,16 +1,14 @@
 package io.digdag.core.log;
 
 import java.time.Instant;
-import java.util.zip.GZIPOutputStream;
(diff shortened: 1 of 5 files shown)


yoyama:
@muga Thanks for your pointing out.
I guess UPLOAD_THRESHOLD of BufferedRemoteTaskLogger should be also configurable.
So If I will fix it, I try to refactor and improve both code.

PR #1219, opened 2019-09-06
Upgrade fluency-meter-registry

yoyama:
Upgrade fluency-meter-registry and some changes for refactoring and others.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -102,7 +102,7 @@ subprojects {
         compile 'org.slf4j:slf4j-api:1.7.25'
         compile group: 'io.micrometer', name: 'micrometer-core', version: '1.2.0'
         compile group: 'io.micrometer', name: 'micrometer-registry-jmx', version: '1.2.0'
-        compile(group: 'io.github.yoyama', name: 'fluency-meter-registory_2.12', version: '0.1.0') {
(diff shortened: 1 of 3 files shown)


muga:
@yoyama the change itself looks good 👍 Can you share the diff between 0.1.0 and 0.2.0 of fluency-meter-registry just in case?

yoyama:
@muga Thanks for pointing out.
Following the diff of 0.1.0 and 0.2.0.

https://github.com/yoyama/FluencyMeterRegistry/compare/rel-0.1.0...rel-0.2.0

It include class name, signature change and test framework change and others. So  a bit complicated.
Essential changes are as follows.

https://github.com/yoyama/FluencyMeterRegistry/blob/cd73da61611b791c31223c5dd2479ebbaf8579eb/src/main/scala/io/github/yoyama/micrometer/FluencyMeterRegistry.scala#L177-L190

Above code come from the digdag code as follows.
https://github.com/treasure-data/digdag/blob/v0_10/digdag-server/src/main/java/io/digdag/server/metrics/jmx/DigdagJmxMeterRegistry.java#L33-L59

These code remove metrics tag from Meter.Id  when Metrics instance created.

PR #1223, opened 2019-09-14
Change core.TaskCallbackApi to pass TaskRequest object to each interface method

muga:
In this PR we changed the signatures of interface methods in `io.digdag.core.TaskCallbackApi` to pass `TaskRequest` object to those methods. It would make more change tolerant.


muga:
@yoyama thank you! merging..

PR #1225, opened 2019-09-19
Fix bq output parameter.

zaimy:
Hello,

The `bq`, `bq_extract`, and `bq_load` operators are documented as accepting `bq.last_job_id` as an output parameter, but in the current implementation it is `bq.last_jobid`.

Since the td operator can use `td.last_job_id`, I thought `last_job_id` may be correct and fixed it like that.

### DIFF
--- a/digdag-docs/src/operators/bq.md
+++ b/digdag-docs/src/operators/bq.md
@@ -159,4 +159,4 @@ When you set those parameters, use [digdag secrets command](https://docs.digdag.
 
 * **bq.last_job_id** or **bq.last_jobid**
 
-  The id of the BigQuery job that performed this export. `bq.last_jobid` will be kept for compatibility and remove in near future release.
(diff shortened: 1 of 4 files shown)


yoyama:
@zaimy Thank you for your contribution!

I think I understand the issue and your fix is correct.

But we hope to keep compatibility as much as possible.
If someone is using bq.last_jobid in their workflows, the workflow will fail when they upgrade their Digdag.

So I would like to propose that keep bq.last_jobid and only add bq.last_job_id. 

In addition, it is better to describe as "bq.last_jobid will be kept for compatibility and remove in near future release" in the document.

zaimy:
Thank you check the PR!
I agree with your comment. I will add some commits. 😃

zaimy:
@yoyama Changed the Pull Request policy to make both old and new parameters available. Could you check it?

zaimy:
Thanks too!

PR #1229, opened 2019-10-08
Remove noisy stacktrace.

yoyama:
To avoid too many log are generated, remove printStackTrace from DigdagTimedMethodInterceptor.


yoyama:
To fix TravisCI error. I did rebase v0_10 which contain fix on it.
After CI pass, I will merge.

PR #1235, opened 2019-10-10
Change how to fetch ready tasks in WorkflowExecutor.

yoyama:
Change how to fetch ready tasks when WorkflowExecutor enqueue them.

WorkflowExecutor fetch ready tasks and then enqueue them.
This PR add option to customize it.

- executor.enqueue_random_fetch change the order to fetch tasks.
   If true, it fetches randomly
- executor.enqueue_fetch_size set the number of tasks to be fetched.

Optimize these parameters affect performance of WorkflowExecutor.
Default value is same as original. So without parameter setting there is no impact to behavior.


muga:
@yoyama Thank you for fixing! LGTM.

PR #1236, opened 2019-10-10
[doc]Add detailed usage of sla option

morihaya:
Adding the sla: option usage including  forced stop job sample.

### DIFF
--- a/digdag-docs/src/scheduling_workflow.rst
+++ b/digdag-docs/src/scheduling_workflow.rst
@@ -134,7 +134,6 @@ If you have a job that always finishes in 3 minutes, and you want to end it forc
     sla:
       # Kill after 3 minutes
       duration: 00:03:00
-      fail: true
       +kill:
         sh>: digdag kill ${attempt_id}
 
--- a/digdag-docs/src/scheduling_workflow.rst
+++ b/digdag-docs/src/scheduling_workflow.rst
@@ -105,6 +105,40 @@ Setting an alert if a workflow doesn't finish within expected time
     +long_running_job:
       sh>: long_running_job.sh
 
+In sla: directive, you can select either the ``time`` or ``duration`` option.
+
+=============================== ================================================== ==========================
+Syntax                          Description                                        Example
+=============================== ================================================== ==========================
+time>: ``HH:MM:SS``             Set this job must be completed by ``HH:MM:SS``     time>: 12:30:00
+duration>: ``HH:MM:SS``         Set this job must be completed during ``HH:MM:SS`` duration>: 00:05:00
+=============================== ================================================== ==========================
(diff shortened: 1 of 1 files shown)


yoyama:
@szyn Could you review this PR if you have a chance?

digdag-docs/src/scheduling_workflow.rst
```diff
@@ -105,6 +105,40 @@ Setting an alert if a workflow doesn't finish within expected time
...
+time>: ``HH:MM:SS``             Set this job must be completed by ``HH:MM:SS``     time>: 12:30:00
+duration>: ``HH:MM:SS``         Set this job must be completed during ``HH:MM:SS`` duration>: 00:05:00
+=============================== ================================================== ==========================
+
+Options
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This parameter supports fail: BOOLEAN and alert: BOOLEAN options. Setting fail: true makes the workflow failed. Setting alert: true sends an notification using above notification mechanism.
+
+* Setting ``fail: true`` makes the workflow failed.
```
szyn: This option makes the workflow status to fail, but trailing tasks keep running. I understand that this behavior sometimes causes confuse users. I want to confirm this behavior as intentional or not.

szyn: This option makes the workflow status to fail, but trailing tasks keep running. I understand that this behavior sometimes causes confuse users. I want to confirm this behavior as intentional or not.

digdag-docs/src/scheduling_workflow.rst
```diff
@@ -105,6 +105,40 @@ Setting an alert if a workflow doesn't finish within expected time
...
+* Setting ``alert: true`` sends an notification using above notification mechanism.
+
+If you have a job that always finishes in 3 minutes, and you want to end it forcibly after 3 minutes, write the following:
```
szyn: This might be a good example of workarounds for now, but there are problems such as not being able to use it in local mode. Thus I am not sure if it is good to introduce this example as an official procedure.
Also, as pointed out in the comment above, I think it's better to clarify the behavior before adding an example. For now, it might be a good idea to just add a description of the behavior.

digdag-docs/src/scheduling_workflow.rst
morihaya: I agree with that opinion.
I think it would be easier to cancel when the sla `duration` or `time` is exceeded.

digdag-docs/src/scheduling_workflow.rst
morihaya: Thank you for a very good indication. I think so too, so I deleted the sample.

PR #1237, opened 2019-10-11
Upgrade fluency-meter-registry to support old fluentd

yoyama:
To support old fluentd, I would like to upgrade fluency-meter-registry to 0.4.0.
Ver. 0.4.0 stop using EventTime.

The differences in fluency-meter-registr between previous version and 0.4.0 are as follows.

https://github.com/yoyama/FluencyMeterRegistry/compare/rel-0.2.0...rel-0.4.0

The related modification is as follows.

https://github.com/yoyama/FluencyMeterRegistry/compare/rel-0.2.0...rel-0.4.0#diff-f94b8cc5451f8031df3b29761290e71dR168

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,6 +45,9 @@ allprojects {
         maven {
             url "https://oss.sonatype.org/content/repositories/snapshots"
         }
+        maven {
+            url "https://dl.bintray.com/yoyama/maven"
+        }
     }
 
     jacoco {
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -45,6 +45,9 @@ allprojects {
         maven {
             url "https://oss.sonatype.org/content/repositories/snapshots"
         }
+        maven {
+            url "https://dl.bintray.com/yoyama/maven"
+        }
```
muga: so you stop using maven central?:-)

yoyama: Yes. Sonatype is too slow and frequently errors :(

yoyama:
@muga Can you review again? I need to fix an integration test for fluency metrics.

PR #1249, opened 2019-11-02
Add diff API to compare two revisions by REST API

frsyuki:
This change adds `/api/projects/{id}/compare?from=&to=` API which
compares two revisions and returns diff of them. This API is useful when
a session failed after pushing a new revision and investigate which
change made the impact.


yoyama:
I would like to close the PR because of suspended long time.
But an interesting PR and may be useful. We will re-open it if needed in the future.

PR #1251, opened 2019-11-05
[WIP] Update queries related to attempts

szyn:
The main purpose is to improve the performance of the queries used in the `attempts` related methods. Make use of the `project_id` in the `session_attempts` table instead of the `sessions` table.


szyn:
I will close this PR because I need more investigation.

PR #1253, opened 2019-11-08
upgrade kubernetes client library version

TrsNium:
# Motivation
I want to support the latest kubernetes version. 
But compile error when upgrade kubernetes-client version to 4.6.2 due to dependency.
So some libraries exclude to avoid compilation failures due to dependencies and jacksonVersion and jacksonDatabindVersion upgrade  to 2.9.10.

This PR is necessary to solve the following issues
https://github.com/treasure-data/digdag/issues/1254

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -123,8 +123,8 @@ subprojects {
     }
 
     ext {
-        jacksonVersion = "2.9.9"
-        jacksonDatabindVersion = "2.9.9"
+        jacksonVersion = "2.9.10"
+        jacksonDatabindVersion = "2.9.10"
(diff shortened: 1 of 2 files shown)


sonots:
write comments why you needed to exclude some libraries

sonots:
@yoyama @muga he is a member of my team, and we are planning to migrate our workflow engine from Google Cloud Composer to digdag, and this is the first step. Could you take a look at this?

muga:
@TrsNium @sonots sorry to be late. i take a look

TrsNium:
@muga @szyn 
Thank you for review

PR #1256, opened 2019-11-11
Add Graal JavaScript engine in addition to Nashorn

frsyuki:
New option `eval.js-engine-type` chooses engine type:

* eval.js-engine-type=nashorn: use Nashorn
* eval.js-engine-type=graal: use Graal.js
* eval.js-engine-type=nashorn-graal-check: use Nashorn but also use
Graal and check the result. If the result is different from Nashorn's
result, show a error message to logger.

Default engine type is Nashorn if JVM is older than 11. Otherwie,
Graal.js.

Graal.js enables nashorn-compatible mode to minimize impact.

In addition to the change, JavaScript parameter evaluation
supports nested parens in JavaScript if engine is Graal.

Config eval engine had a problem where JavaScript code can't use useful
functional methods such as Array.map because code couldn't include `}`
character. With this change, `}` is assumed as a part of JavaScript code
if number of `{` and `}` matches. We can see similar strategy in Ruby
(`%{...}` syntax).

This change is enabled only for Graal JavaScript engine so that it
doesn't make unexpected backward incompatibility.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
@@ -1,30 +1,23 @@
 package io.digdag.core.agent;
 
-import java.util.List;
-import java.util.Map;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.IOException;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.charset.Charset;
-import javax.script.ScriptEngine;
-import javax.script.ScriptEngineManager;
-import javax.script.ScriptException;
-import javax.script.Invocable;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.io.CharStreams;
 import com.google.inject.Inject;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-import jdk.nashorn.api.scripting.NashornScriptEngineFactory;
 import io.digdag.client.config.Config;
+import io.digdag.client.config.ConfigException;
 import io.digdag.spi.TemplateEngine;
 import io.digdag.spi.TemplateException;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.util.Map;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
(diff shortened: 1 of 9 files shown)


digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
```diff
@@ -1,44 +1,38 @@
...
     private static Logger logger = LoggerFactory.getLogger(ConfigEvalEngine.class);
 
-    private static final String DIGDAG_JS_RESOURCE_PATH = "/io/digdag/core/agent/digdag.js";
-
-    private static final String[] LIBRARY_JS_RESOURCE_PATHS = {
-        "/io/digdag/core/agent/moment.min.js"
+    // array of (name, resourceName)
+    private static final String[][] LIBRARY_JS_RESOURCES = {
+        new String[] { "digdag.js", "/io/digdag/core/agent/digdag.js", "digdag.js" },
+        new String[] { "moment.min.js", "/io/digdag/core/agent/moment.min.js", "moment.min.js" },
```
yoyama: Comment says each row has two column (name, resource), but there are 3 columns.
3rd column is used ?

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
```diff
@@ -51,94 +45,116 @@ private static String readResource(String resourceName)
...
+        // Return NASHORN if runtime Java VM is older than 11. Otherwise,
+        // by default, return GRAAL. This is because org.graalvm.js:js depends
+        // on JVMCI (JEP 243) which is available only from Java 11.
+        String javaSpecVer = System.getProperty("java.specification.version");
+        int major = Integer.parseInt(javaSpecVer.split("[^0-9]")[0]);
+        if (major < 11) {
+            return JsEngineType.NASHORN;
         }
-        return (Invocable) jsEngine;
+        return JsEngineType.GRAAL;
```
yoyama: If we change the default JS engine based on Java version, what about logging to inform it?

yoyama:
@frsyuki  Sorry for my late response.
I am still working on the review, but I would like to leave some comments before I will be off from tomorrow.



1. I tried to create shadow jar with _"./gradlew clean cli -PwithoutUi"_, but it take very long time in _:digdag-cli:shadowJar_ .
   I waited for over 30 min, but it did not finish. When I did same thing with v0_10 branch, it took under 2 min. Is there any change to improve the situation?
   My JDK version is as follows.
   ```
   $ java -version
   java version "1.8.0_201"
   Java(TM) SE Runtime Environment (build 1.8.0_201-b09)
   Java HotSpot(TM) 64-Bit Server VM (build 25.201-b09, mixed mode)
   ```
   I guess memory of gradle may be not enough. But I set org.gradle.jvmargs in gradle.properties. Not yet resolved.
1. I left some comments in code on minor things.
1. About nested paren support, I hope this functionality should be enable/disable in system config if users choose graal. Our JDK version support plan is not clear. But in your investigation,  nashorn in Java11 is too slow. So if we would decide to stop support of Java8 in the future, users will be forced to use graal. In this case incompatible syntax may be affected to users workflow. So it is better to be configurable.

yoyama:
#1361 is follow up this PR and it has been merged. So close.

PR #1266, opened 2019-11-25
Improve ConfigEvalEngine performance by remove unnecessarry calls

yoyama:
This PR improve performance of operator processing by reducing call of Javascript engine processing.
Current code always call invokeTemplate().
But if there is no evaluated code in the text, we don't need to call it.
This PR checks text contains '$' char and only call invokeTemplate() if exists.

I tested with the following dig and measured the time with output _digdag_ attempt <id>_ .

|   secs  | 1  | 2  | 3  | Avg.  |
|:-:      |:-: |:-: |:-: |:-:    |
|original |17  |12  |13  | 14.0  |
|modified |14  | 9  | 9  | 10.67 |

As above, the PR is faster than original in this dig.
Performance depends on the dig.
But in most case it improve the performance.

- This PR improve the performance of operator by skip invokeTemplate() if needless
- Will improve CPU utilization
- Even though we move to GraalJS, this code(or way) is still effective.

```
+task1:
  for_each>:
    fruit: [apple, orange, banana]
    color: [white, black, blue, green, red, yellow]
    verb: [eat, throw, look]
  _parallel: true
  _do:
    echo>: ${verb} ${color} ${fruit}

+task2:
  for_each>:
    fruit: [apple, orange, banana]
    color: [white, black, blue, green, red, yellow]
    verb: [eat, throw, look]
  _parallel: true
  _do:
    echo>: aaaaa
```

As additional test, I measured with +task1 only WF. In this case, task always have '$' the performance should be near equivalent.

|   secs  | 1  | 2  | 3  | Avg.  |
|:-:      |:-: |:-: |:-: |:-:    |
|original |10  |7  | 6 | 7.67  |
|modified |11  |7  | 5  |7.67  |

The result is as expected. In not good condition for this PR,  performance is almost all same.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
@@ -5,6 +5,7 @@ import java.util.Map;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.IOException;
+import java.util.regex.Pattern;
 import javax.script.ScriptEngine;
 import javax.script.ScriptException;
 import javax.script.Invocable;
@@ -233,9 +234,11 @@ public class ConfigEvalEngine
         }
     }
 
+    private static final Pattern requireInvokdetemplatePattern = Pattern.compile("(?m)^.*[$].*$", Pattern.MULTILINE);
+
     @VisibleForTesting
     protected boolean requireInvokdetemplate(String code) {
-        return code != null && code.matches("^.*[$].*$");
+        return code != null && requireInvokdetemplatePattern.matcher(code).find();
     }
 
     @Override
--- a/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
@@ -18,7 +18,6 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.io.CharStreams;
 import com.google.inject.Inject;
 import io.digdag.metrics.DigdagTimed;
-import io.digdag.spi.metrics.DigdagMetrics;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import jdk.nashorn.api.scripting.NashornScriptEngineFactory;
@@ -63,7 +62,6 @@ public class ConfigEvalEngine
 
     private final ObjectMapper jsonMapper;
     private final NashornScriptEngineFactory jsEngineFactory;
-    private DigdagMetrics metrics;
 
     @Inject
     public ConfigEvalEngine()
@@ -113,8 +111,7 @@ public class ConfigEvalEngine
             throw new TemplateException("Failed to serialize parameters to JSON", ex);
         }
         try {
-            String evaluated = (String) templateInvocable.invokeFunction("template", code, context);
-            return evaluated;
+            return (String) templateInvocable.invokeFunction("template", code, context);
         }
         catch (ScriptException ex) {
             String message;
@@ -145,13 +142,10 @@ public class ConfigEvalEngine
             this.params = params;
         }
 
-        // ToDo Lombok Getter(lazy=true)
         private Invocable lazyGetTemplateInvocable()
         {
             if (templateInvocable == null) {
-                synchronized(this) {
-                    templateInvocable = newTemplateInvocable(params);
-                }
+                templateInvocable = newTemplateInvocable(params);
             }
             return templateInvocable;
         }
@@ -219,7 +213,7 @@ public class ConfigEvalEngine
                 scopedParams.set(pair.getKey(), pair.getValue());
             }
             String resultText = null;
-            if (requireInvokdetemplate(code)) {
+            if (requireInvokedTemplate(code)) {
                 resultText = invokeTemplate(lazyGetTemplateInvocable(), code, scopedParams);
             }
             else {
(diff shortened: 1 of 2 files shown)


digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
```diff
@@ -213,12 +234,25 @@ private JsonNode evalValue(ObjectNode local, String code)
         }
     }
 
+    private static final Pattern requireInvokdetemplatePattern = Pattern.compile("(?m)^.*[$].*$", Pattern.MULTILINE);
```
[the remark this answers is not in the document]
frsyuki: `requireInvokdetemplatePattern.contains("$")` is even simpler and safer. but using expression seems also fair idea. which do you prefer?

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
```diff
@@ -109,7 +113,8 @@ private String invokeTemplate(Invocable templateInvocable, String code, Config p
             throw new TemplateException("Failed to serialize parameters to JSON", ex);
         }
         try {
-            return (String) templateInvocable.invokeFunction("template", code, context);
+            String evaluated = (String) templateInvocable.invokeFunction("template", code, context);
+            return evaluated;
```
frsyuki: If there is a significant reason why this change is needed, please add comments here. If it's just styling matter, that's fine.

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
```diff
@@ -213,12 +234,25 @@ private JsonNode evalValue(ObjectNode local, String code)
         }
     }
 
+    private static final Pattern requireInvokdetemplatePattern = Pattern.compile("(?m)^.*[$].*$", Pattern.MULTILINE);
+
+    @VisibleForTesting
+    protected boolean requireInvokdetemplate(String code) {
```
frsyuki:
```suggestion
    protected boolean requireInvokdeTemplate(String code)
    {
```

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
```diff
@@ -63,6 +63,7 @@ private static String readResource(String resourceName)
 
     private final ObjectMapper jsonMapper;
     private final NashornScriptEngineFactory jsEngineFactory;
+    private DigdagMetrics metrics;
```
frsyuki: this variable is not used. please delete it as well as `import` lines

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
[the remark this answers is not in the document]
yoyama: I prefer your simple idea. Thanks!

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
```diff
@@ -132,13 +137,23 @@ private String invokeTemplate(Invocable templateInvocable, String code, Config p
...
         {
             this.params = params;
-            this.templateInvocable = newTemplateInvocable(params);
+        }
+
+        // ToDo Lombok Getter(lazy=true)
+        private Invocable lazyGetTemplateInvocable()
+        {
+            if (templateInvocable == null) {
+                synchronized(this) {
```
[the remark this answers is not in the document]
yoyama: Thank you for your advice. I will fix it.

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
```diff
@@ -132,13 +137,23 @@ private String invokeTemplate(Invocable templateInvocable, String code, Config p
...
+        private Invocable templateInvocable = null;
         private final ImmutableList<String> noEvaluatedKeys = ImmutableList.of("_do",  "_else_do");
 
         public Context(Config params)
         {
             this.params = params;
-            this.templateInvocable = newTemplateInvocable(params);
+        }
+
+        // ToDo Lombok Getter(lazy=true)
```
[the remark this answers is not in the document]
yoyama: I removed it.

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
yoyama: Sorry, I reverted it.

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
yoyama: Thanks. Removed it.

digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
```diff
@@ -203,7 +212,13 @@ private JsonNode evalValue(ObjectNode local, String code)
             for (Map.Entry<String, JsonNode> pair : ImmutableList.copyOf(local.fields())) {
                 scopedParams.set(pair.getKey(), pair.getValue());
             }
-            String resultText = invokeTemplate(templateInvocable, code, scopedParams);
+            String resultText = null;
+            if (requireInvokedTemplate(code)) {
```
frsyuki: This is minor but you may reconsider this method name.
boolean methods used in digdag code are https://github.com/treasure-data/digdag/search?q=%22public+boolean%22&unscoped_q=%22public+boolean%22

yoyama: Thank you for your advice. I agree and  I will fix it.

yoyama: Fixed it.

PR #1273, opened 2019-12-04
[Doc] Add _check and _error example

hiroyuki-sato:
A user told me there is no example of `_check` usages.
This PR adds how to use `_check` and `_error` tasks

@yoyama, @szyn What do you think?

Ref: [Twitter](https://twitter.com/_osna329_/status/1201863710056144898?s=20) Japanese


hiroyuki-sato:
Thank you!

PR #1280, opened 2019-12-16
Introduce timeout to s3_wait> operator.

yoyama:
This PR introduces timeout for s3_wait> operator.
The following two options are introduced.

- **timeout**  : set timeout duration.
- **ignore_timeout_error** : if true, the task will finish successfully.

example:
```
+task1:
  s3_wait>: <buckect>/<file>
  timeout: 120s
  ignore_timeout_error: true
```

### Interval calculation
This PR changed the interval calculation.
Interval start 5 seconds and gradually increase and  5 minutes is max.
The current formula is as follows.
```
 5sec * 2^(n-1)    (n: num of iteration)
```
It is too much increase if timeout parameter is not so large.
For example if timeout is 180s, the task will be time out after n=6 because total interval is 155. But next interval in n=6 is 160. So the task wait time will be 315seconds. 
It is over 130 seconds.

|               | n=1  | 2   | 3     | 4   | 5  |6  |   7|
|:-:           |:-:     |:-:   |:-:    |:-:   |:-: |:-: |  :-: |
| interval |  5     | 10  | 20  | 40 | 80|160|320|
|total       |  0     | 5    | 15   |35  |75 |155|315|

To mitigate, this PR changed exponential base value from 2 to 1.2.
In this case, total wait time will be 197 seconds and the over time is 17.

### DIFF
--- a/digdag-docs/src/operators/s3_wait.md
+++ b/digdag-docs/src/operators/s3_wait.md
@@ -81,6 +81,21 @@ For more information about SSE-C, See the [AWS S3 Documentation](http://docs.aws
   An optional flag to control whether to use path-style or virtual hosted-style access when accessing S3.
   *Note:* Enabling `path_style_access` also requires specifying a `region`.
 
+* **timeout**: TIMEOUT
+
+  Set timeout.
+
+  Examples: wait 120 seconds
+
+  ```
+  timeout: 120s
+  ```
+
+* **ignore_timeout_error**: true/false (default:false)
+
+  If ignore_timeout_error is set to true, the task will finish successfully.
+  *Note* s3 variable is not accessible.
+
 ## Output Parameters
 
 * **s3.last_object**
--- a/digdag-docs/src/operators/s3_wait.md
+++ b/digdag-docs/src/operators/s3_wait.md
@@ -94,7 +94,18 @@ For more information about SSE-C, See the [AWS S3 Documentation](http://docs.aws
 * **continue_on_timeout**: true/false (default:false)
 
   If continue_on_timeout is set to true, the task will finish successfully on timeout.
-  s3.last_object.result is set to false.
+  s3.last_object is empty in this case. Empty check is required in following tasks if access to s3.last_object.
+  
+  ```
+  +task1:
+    s3_wait>: bucket/object
+    timeout: 60s
+    continue_on_timeout: true
+  +task2:
+    if>: ${s3.last_object}
+    _do:
+      echo>: "No timeout"
+  ```
 
 ## Output Parameters
 
--- a/digdag-docs/src/operators/s3_wait.md
+++ b/digdag-docs/src/operators/s3_wait.md
@@ -91,10 +91,10 @@ For more information about SSE-C, See the [AWS S3 Documentation](http://docs.aws
   timeout: 120s
   ```
 
-* **ignore_timeout_error**: true/false (default:false)
+* **continue_on_timeout**: true/false (default:false)
 
-  If ignore_timeout_error is set to true, the task will finish successfully.
-  *Note* s3 variable is not accessible.
+  If continue_on_timeout is set to true, the task will finish successfully on timeout.
+  s3.last_object.result is set to false.
 
 ## Output Parameters
 
--- a/digdag-docs/src/operators/s3_wait.md
+++ b/digdag-docs/src/operators/s3_wait.md
@@ -95,7 +95,7 @@ For more information about SSE-C, See the [AWS S3 Documentation](http://docs.aws
 
   If continue_on_timeout is set to true, the task will finish successfully on timeout.
   s3.last_object is empty in this case. Empty check is required in following tasks if access to s3.last_object.
-  
+
   ```
   +task1:
     s3_wait>: bucket/object
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
@@ -199,7 +199,7 @@ public class S3WaitOperatorFactory
                         .build();
             }
             catch (PollingTimeoutException e) {
-                logger.debug(e.toString());
+                logger.debug("s3_wait timed out: {} (making the task {})", (ignoreTimeoutError ? "continued" : "failed"), , e.toString());
                 if (ignoreTimeoutError) {
                     return TaskResult.defaultBuilder(request)
                             .resetStoreParams(ImmutableList.of(ConfigKey.of("s3", "last_object")))
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
@@ -214,12 +214,12 @@ public class S3WaitOperatorFactory
             Config params = request.getConfig().getFactory().create();
             Config object = params.getNestedOrSetEmpty("s3").getNestedOrSetEmpty("last_object");
             if (objectMetadata.isPresent()) {
-                object.set("result", true);
+                object.set("stored", true);
                 object.set("metadata", objectMetadata.get().getRawMetadata());
                 object.set("user_metadata", objectMetadata.get().getUserMetadata());
             }
             else { // Timeout happen
-                object.set("result", false);
+                object.set("stored", false);
             }
             return params;
         }
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
@@ -45,6 +45,7 @@ public class S3WaitOperatorFactory
     private static Logger logger = LoggerFactory.getLogger(S3WaitOperatorFactory.class);
 
     private static final DurationInterval POLL_INTERVAL = DurationInterval.of(Duration.ofSeconds(5), Duration.ofMinutes(5));
+    private static final double NEXT_INTERVAL_EXP_BASE = 1.2; // interval time will increase 5sec *1.2**N until 5min.
 
     private final AmazonS3ClientFactory s3ClientFactory;
     private final Map<String, String> environment;
@@ -177,6 +178,7 @@ public class S3WaitOperatorFactory
             try {
                 ObjectMetadata objectMetadata = pollingWaiter(state, "EXISTS")
                         .withPollInterval(POLL_INTERVAL)
+                        .withNextIntervalExpBase(NEXT_INTERVAL_EXP_BASE)
                         .withTimeout(timeout)
                         .withWaitMessage("Object '%s/%s' does not yet exist", bucket.get(), key.get())
                         .await(pollState -> pollingRetryExecutor(pollState, "POLL")
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
@@ -28,6 +28,7 @@ import io.digdag.spi.TaskResult;
 import io.digdag.standards.operator.DurationInterval;
 import io.digdag.standards.operator.state.PollingTimeoutException;
 import io.digdag.standards.operator.state.TaskState;
+import io.digdag.util.DurationParam;
 import io.digdag.util.Durations;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -103,9 +104,8 @@ public class S3WaitOperatorFactory
             Optional<String> key = params.getOptional("key", String.class);
             Optional<String> versionId = params.getOptional("version_id", String.class);
             Optional<Boolean> pathStyleAccess = params.getOptional("path_style_access", Boolean.class);
-            Optional<Duration> timeout = params.getOptional("timeout", String.class)
-                                            .transform( (t) -> Durations.parseDuration(t));
-            boolean ignoreTimeoutError = params.getOptional("ignore_timeout_error", Boolean.class).or(false);
+            Optional<Duration> timeout = params.getOptional("timeout", DurationParam.class).transform(DurationParam::getDuration);
+            boolean continueOnTimeout = params.getOptional("continue_on_timeout", Boolean.class).or(false);
 
             if (command.isPresent() && (bucket.isPresent() || key.isPresent()) ||
                     !command.isPresent() && (!bucket.isPresent() || !key.isPresent())) {
@@ -178,7 +178,6 @@ public class S3WaitOperatorFactory
             try {
                 ObjectMetadata objectMetadata = pollingWaiter(state, "EXISTS")
                         .withPollInterval(POLL_INTERVAL)
-                        .withNextIntervalExpBase(NEXT_INTERVAL_EXP_BASE)
                         .withTimeout(timeout)
                         .withWaitMessage("Object '%s/%s' does not yet exist", bucket.get(), key.get())
                         .await(pollState -> pollingRetryExecutor(pollState, "POLL")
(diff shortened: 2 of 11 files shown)


digdag-docs/src/operators/s3_wait.md
```diff
@@ -81,6 +81,21 @@ For more information about SSE-C, See the [AWS S3 Documentation](http://docs.aws
...
+
+  Examples: wait 120 seconds
+
+  ```
+  timeout: 120s
+  ```
+
+* **ignore_timeout_error**: true/false (default:false)
+
+  If ignore_timeout_error is set to true, the task will finish successfully.
```
frsyuki: Please re-consider better naming. If timeout shouldn't make the workflow failed, the user probably thinks timeout is not an error. From the metal model, the user is not ignoring error, but they don't think it is an error.

digdag-docs/src/operators/s3_wait.md
```diff
@@ -81,6 +81,21 @@ For more information about SSE-C, See the [AWS S3 Documentation](http://docs.aws
...
+
+  Set timeout.
+
+  Examples: wait 120 seconds
+
+  ```
+  timeout: 120s
+  ```
+
+* **ignore_timeout_error**: true/false (default:false)
```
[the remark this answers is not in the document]
frsyuki: What is the use case of setting `true` here? Depending on it, shouldn't s3_wait set a store_param when timeout is ignored so that following tasks can know whether the file exists or not? Some workflows may want to branch using `if> ${store_parameter_set_by_s3_wait}`?
Please consider naming together. An example is `continue_on_timeout` but please think.

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
```diff
@@ -102,7 +149,7 @@ public PollingWaiter withPollInterval(DurationInterval retryInterval)
         if (!result.isPresent()) {
             int iteration = pollState.params().get(ITERATION, int.class, 0);
             pollState.params().set(ITERATION, iteration + 1);
-            int interval = (int) Math.min(pollInterval.min().getSeconds() * Math.pow(2, iteration), pollInterval.max().getSeconds());
+            int interval = (int) Math.min(pollInterval.min().getSeconds() * Math.pow(nextIntervalExpBase, iteration), pollInterval.max().getSeconds());
```
frsyuki: With exp=1.2, timeout is:
1s, 1.2s, 1.44s, 1.73, ..., 197s (i=29), 237s (i=30)
And cumulative waiting time is:
1s, 2.2s, 3.64s, ..., 19.7m (i=29), 23.7m (i=30)

If user sets 20 minute of timeout, s3_waits waits 23.7 minute, where it should fail after 20 minutes. 3 minutes, 15% of difference is a significant & unexpected.
If we shorten `nextIntervalExpBase`, s3_wait makes too many polling. This is a trade-off.

But if we modify this `interval` based remaining timeout wait, we don't have have the trade-off. If remaining time is 0.3 minute, wait 0.3 minute instead of waiting for 5 minute. This implementation seems easy. I don't think it's good idea to leave such easy improvement.

Please add it, and remove this change https://github.com/treasure-data/digdag/pull/1280/files#diff-5371699d0c423b5e397f66775898325fR48 which is making negative impact.

digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
```diff
@@ -141,12 +147,13 @@ public TaskResult run()
 
             Aws.configureServiceClient(s3Client, endpoint, regionName);
 
-            S3ClientOptions clientOptions = new S3ClientOptions();
-            if (pathStyleAccess.isPresent()) {
-                clientOptions.setPathStyleAccess(pathStyleAccess.get());
+            {
+                S3ClientOptions.Builder builder = S3ClientOptions.builder();
```
frsyuki: Removing warning is very good idea 👍

digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
```diff
@@ -168,27 +175,38 @@ public TaskResult run()
...
-                    .storeParams(storeParams(objectMetadata))
-                    .build();
+                                }));
+                return TaskResult.defaultBuilder(request)
+                        .resetStoreParams(ImmutableList.of(ConfigKey.of("s3", "last_object")))
+                        .storeParams(storeParams(objectMetadata))
+                        .build();
+            }
+            catch (PollingTimeoutException e) {
+                logger.debug(e.toString());
```
frsyuki: Please keep in mind that this debug message is visible to users. At least we should add why a message is shown. Let's keep high quality of logging so that we don't get exhausted due to too many bad logs...

```suggestion
                logger.debug("s3_wait timed out: {} (making the task {})", (ignoreTimeoutError ? "continued" : "failed"), , e.toString());
```

Or if you think the message is really for development purpose only, then use trace level instead. Examples: https://github.com/treasure-data/digdag/search?q=logger.trace&unscoped_q=logger.trace

digdag-docs/src/operators/s3_wait.md
```diff
@@ -81,6 +81,21 @@ For more information about SSE-C, See the [AWS S3 Documentation](http://docs.aws
...
+
```
frsyuki: PollingWaiter is used also by EmrOperatorFactory, BqJobRunner, and GcsWaitOperatorFactory. At least GcsWaitOperatorFactory should have the same option. Consistency across different but similar operators is important for usability.
What is your strategy to expand the same option to those other operators?

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
```diff
@@ -87,9 +123,20 @@ public PollingWaiter withPollInterval(DurationInterval retryInterval)
...
         TaskState pollState = state.nestedState(stateKey);
-
+        Long startTime = pollState.params().get(START_TIME, Long.class, null);
+        if (startTime == null) {
+            startTime = System.currentTimeMillis();
+            pollState.params().set(START_TIME, startTime);
+        }
         TaskState operationState = pollState.nestedState(OPERATION);
 
+        //Check timeout
```
frsyuki:
```suggestion
        // Check timeout
```

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
```diff
@@ -87,9 +123,20 @@ public PollingWaiter withPollInterval(DurationInterval retryInterval)
...
+        Long now = System.currentTimeMillis();
```
frsyuki: Don't use boxed type when unnecessary
```suggestion
        long now = System.currentTimeMillis();
```

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
```diff
@@ -87,9 +123,20 @@ public PollingWaiter withPollInterval(DurationInterval retryInterval)
     public <T> T await(Operation<Optional<T>> f)
     {
         TaskState pollState = state.nestedState(stateKey);
-
+        Long startTime = pollState.params().get(START_TIME, Long.class, null);
```
frsyuki: Digdag's code base highly takes advantage of `java.time` package. `Instant` is highly preferred than System.currentTimeMillis.
* Existent use of `System.currentTimeMillis()`:  https://github.com/treasure-data/digdag/search?q=currentTimeMillis&unscoped_q=currentTimeMillis
* Existent use of `Instant.now()`: https://github.com/treasure-data/digdag/search?q=%22Instant.now%22&unscoped_q=%22Instant.now%22

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
```diff
@@ -87,9 +123,20 @@ public PollingWaiter withPollInterval(DurationInterval retryInterval)
...
+        if (startTime == null) {
+            startTime = System.currentTimeMillis();
+            pollState.params().set(START_TIME, startTime);
+        }
         TaskState operationState = pollState.nestedState(OPERATION);
 
+        //Check timeout
+        Long now = System.currentTimeMillis();
+        if (timeout.isPresent() && timeout.get().toMillis() <= now - startTime) {
+            logger.debug("Timeout happened. startTime:{}, timeout:{}", startTime, timeout.get());
```
frsyuki: Please reconsider whether this log is for users or you only, while s3_wait operator itself has its own logging. If you only, then use `trace` or remove.

digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
```diff
@@ -100,6 +103,9 @@ public TaskResult run()
             Optional<String> key = params.getOptional("key", String.class);
             Optional<String> versionId = params.getOptional("version_id", String.class);
             Optional<Boolean> pathStyleAccess = params.getOptional("path_style_access", Boolean.class);
+            Optional<Duration> timeout = params.getOptional("timeout", String.class)
+                                            .transform( (t) -> Durations.parseDuration(t));
```
frsyuki: parseDuration throws a RuntimeException which is unexpected.
You can use DurationParam instead which implements error handling already: https://github.com/treasure-data/digdag/blob/6547542a274ac1a30c88c82e509ac0742d6d872e/digdag-plugin-utils/src/main/java/io/digdag/util/DurationParam.java
Usage of it: https://github.com/treasure-data/digdag/search?q=DurationParam&unscoped_q=DurationParam

digdag-docs/src/operators/s3_wait.md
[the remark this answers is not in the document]
frsyuki: And AbstractWaitOperatorFactory (which is a base class of td_wait and td_wait_table).

digdag-docs/src/operators/s3_wait.md
frsyuki: And AbstractWaitOperatorFactory (which is a base class of td_wait and td_wait_table).

digdag-docs/src/operators/s3_wait.md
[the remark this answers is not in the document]
yoyama: The reason I introduced this parameter is demand from internal.
But I will check it in detail.

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
yoyama: I agree with the this implementation has still impact.
>If remaining time is 0.3 minute, wait 0.3 minute instead of waiting for 5 minute.
I will reconsider based on your idea. In this case users need to wait for until last minute of timeout. In addition, we need to consider the next task will not start precisely, it will be delayed.

digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java
yoyama: Thanks! Will fix it.

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
yoyama: Got it.  Will fix it.

digdag-docs/src/operators/s3_wait.md
yoyama: My thought is the impact of modification to be minimized to avoid unexpected degradation.
I added the parameters, so other operators can change easily later in my understand.
But I understand your opinion.

> Consistency across different but similar operators is important for usability.

frsyuki: ok. Do you mean that the same option should be added to other similar operators but not now? If so, it sounds good. Which operators do you expect to have the option?

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
frsyuki: Next interval should be 55s at the case.
Please also consider jitter though.

digdag-docs/src/operators/s3_wait.md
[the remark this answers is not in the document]
frsyuki: Better understanding of users is very good to write better. If you have more public channels to get information about users and use cases, you are good OSS developer :-)

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
yoyama: How about the following way.
Is this acceptable for you?

**interval = max(5sec,  min( (now- start time), (5sec * 2^N), 5min )**

ex1) timeout: 60s
N=0  : interval = 5s       ( max(5s, min(60s, 5s, 5m)) )
N=1  : interval =  10s     ( max(5s, min(55s, 10s, 5m)) )
N=2  : interval =  20s     ( max(5s, min(45s, 20s, 5m)) )
N=3  : interval =  40s     ( max(5s, min(25s, 40s, 5m)) )
N=4  : interval =  5s       ( max(5s, min(5s, 80s, 5m)) )
N=5  : last check then timeout

ex2) timeout: 60m(3600s)
N=0  : interval = 5s       ( max(5s, min(3600s, 5s, 5m)) )
N=1  : interval =  10s     ( max(5s, min(3555s, 10s, 5m)) )
N=2  : interval =  20s     ( max(5s, min(3545s, 20s, 5m)) )
N=3  : interval =  40s     ( max(5s, min(3525s, 40s, 5m)) )
N=4  : interval =  80s     ( max(5s, min(3485s, 80s, 5m)) )
N=5  : interval =  160s   ( max(5s, min(3405s, 160s, 5m)) )
N=6  : interval =  5m     ( max(5s, min(3245s, 320s, 5m)) )
N=7  : interval =  5m     ( max(5s, min(2945s, 320s, 5m)) )
....
N=15  : interval = 5m      ( max(5s, min(545s, 320s, 5m)) )
N=16  : interval =  245s  ( max(5s, min(245s, larger than 5m, 5m)) )
N=17  : last check then timeout

digdag-docs/src/operators/s3_wait.md
yoyama: Yes. I guess gcs_wait> should also hove same option. 
td_wait> and td_wait_table> may be also good with timeout option.

digdag-docs/src/operators/s3_wait.md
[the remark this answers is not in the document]
yoyama: @frsyuki I would like to introduce s3.last_object.stored : true/false.

I investigated the option name for this behavior. But not good name found.
In Airflow, sensor has _soft_fail_ parameter but downstream tasks will be skipped.
In GitHub Actions, they have _jobs.<job_id>.steps.continue-on-error_ .
It is near to _continue_on_timeout_ which you proposed as example.
So I would like to use the name _continue_on_timeout_

digdag-standards/src/main/java/io/digdag/standards/operator/state/PollingWaiter.java
yoyama: After discussion, I will modify as follows:
- max(5s, min(timeout - (now - start time), 5min)   (if timeout set)
- original formula (if timeout not set)

digdag-docs/src/operators/s3_wait.md
frsyuki: Sounds good. For gcs_wait>, it's good idea to add a comment somewhere. It would say something like...`// timeout 
 and continue_on_timeout should be implemented later here. See https://github.com/treasure-data/digdag/pull/1280/files#r358679909 for details`. (Otherwise we will forget)

yoyama: I filed a ticket in #1281 and our internal ticket system for gcs_wait> timeout.

yoyama:
@frsyuki Could you review again. I updated the code based on the discussion with you.

PR #1284, opened 2019-12-20
[wip]get credential from service account

TrsNium:


TrsNium:
I could not get project id from the method

PR #1286, opened 2019-12-20
Add @Deprecated annotation to deprecated JAX-RS resources for Swagger

frsyuki:
Swagger API document displays endpoints with `@Deprecated` annotation
with grayed out.


yoyama:
@frsyuki  I guess it is better to leave comments why these APIs are deprecated.

frsyuki:
Added some more comments.
ApiOperation is added for #1292.

PR #1287, opened 2019-12-21
Add swagger documents

frsyuki:
This PR adds swagger annotations and link to https://swagger.digdag.io.

New `swaggerYaml` task of build.gradle dumps results of `/api/swagger.yaml` to `digdag-docs/src/_static/swagger.yaml` so that it's available at https://docs.digdag.io/_static/swagger.yaml. When we update Swagger API document, we should run `./gradlew swaggerYaml` task to update the file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ venv*
 .venv*
 /dist/*.tar.gz
 /config/test_postgresql.properties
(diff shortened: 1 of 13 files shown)


frsyuki:
> static html API
This is nice. I created #1292.

PR #1288, opened 2019-12-23
Filter project revisions API by name

chezou:
~This PR introduces Revision API for a specific Project. With the current implementation, there's no way to fetch the latest or specfic revision from REST API. This prevents to get user information associated with revision with the specific project.~

This PR introduces filter by name for Revisions API associated with a specific Project. With the current implementation, we don't have a way to get user info with revision without fetching revisions, but this endpoint fetches fixed number (100) of revisions and we can't fetch older revision than 100.

### DIFF
--- a/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
+++ b/digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
@@ -115,8 +115,8 @@ public class ProjectResource
     // GET  /api/projects                                # list projects
     // GET  /api/projects?name=<name>                    # lookup a project by name, or return an empty array
     // GET  /api/projects/{id}                           # show a project
-    // GET  /api/projects/{id}/revision?name=<name>      # lookup a revision of a project by name
     // GET  /api/projects/{id}/revisions                 # list revisions of a project from recent to old
+    // GET  /api/projects/{id}/revisions?name=<name>     # lookup a revision of a project by name, or return an empty array
     // GET  /api/projects/{id}/workflows                 # list workflows of the latest revision of a project
     // GET  /api/projects/{id}/workflows?revision=<name> # list workflows of a past revision of a project
     // GET  /api/projects/{id}/workflows?name=<name>     # lookup a workflow of a project by name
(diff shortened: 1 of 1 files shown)


chezou:
I know that other API endpoints are shifting to index access, so let me know if this new endpoint `api/projects/{id}/revision` isn't consistent with other APIs.

frsyuki:
`<singular>?name=` is deprecated REST API style.
`/api/projects/{id}/revisions?name=` is consistent. It returns empty list when no matching revision is found.

frsyuki:
What's the use case? Just checking existence of a revision?

chezou:
The original intention was to know the latest revision's author, which is required by our customer using Python digdag client. After sending this PR I realized that `/api/projects/{id}/revisions` API returns fixed number (100) of revisions and seems to be ok to fetch the latest author from revisions  API.
Another use case could be fetching old existing revision to know the user info, but it seems nice to have thing.

chezou:
Changed to filter revision by name.

frsyuki:
@yoyama can I ask you to review? I think this is good

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
```diff
@@ -315,13 +317,27 @@ public RestProject getProject(@PathParam("id") int projId)
...
+            List<StoredRevision> revs;
+            if (revName != null) {
+                try {
+                    StoredRevision rev = ps.getRevisionByName(proj.getId(), revName); // check NotFound first
+                    revs = ImmutableList.of(rev);
+                }
+                catch (ResourceNotFoundException ex) {
+                    revs = ImmutableList.of();
+                }
+            } else {
```
yoyama: Can we change the style as follows.
```
}
else {
```

yoyama:
@chezou Thank you for your contribution. 
I made a comment on minor thing.
After fix it, I will approve and merge.

digdag-server/src/main/java/io/digdag/server/rs/ProjectResource.java
chezou: Thanks, fixed 2aa853a

PR #1292, opened 2019-12-23
Serve REST API document as a static HTML at docs.digdag.io/api/

frsyuki:
digdag-docs/src/api/index.html file uses RapiDoc (MIT License) hosted at
unpkg.com. Sphinx config is changed to serve static files at /api.


digdag-docs/src/conf.py
```diff
@@ -63,7 +63,7 @@ def setup(app):
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
```
chezou: Seems that this keeps copying html and yaml file into `_static` directory under `build/html` on my local environment. Is it as you expected?

chezou: Tried to use `html_extra_path` but it can't copy directories.

yoyama:
@frsyuki  Thank you so much for this PR. It is the just way  I am thinking.
I tried to build the doc. But it looks like  there is no api directory in digdag-docs/build/html/.
Is this OK? If I copied the src/api  to html/  I can display the API.

digdag-docs/src/conf.py
chezou: @frsyuki @yoyama  Finally, I got way to bundle `api` directory setting as:

```suggestion
html_static_path = ['_static']
html_extra_path = ['_extra']
```

with creating `src/_extra` and move current `src/api` directory under `src/_extra/` like:

```
$ tree src/_extra
src/_extra
└── api
    ├── index.html
    └── swagger.yaml
```

frsyuki:
oh thank you!...I updated. do you think this is good?

yoyama:
After @chezou confirm, I will merge.

PR #1263, opened 2019-11-14
Support GCS

TrsNium:
## Purpose
Store project archives to object storage(gcs)
Since the s3 version is already implemented, the gcs version was implemented with reference to it.

## How to run the test
~To run the test, you need to prepare a JsonKey for a Service Account with `roles / storage.admin` privileges.
You will also need to export the JsonKey path to an environment variable at runtime.~

## Other supplementary matters
~In s3 StorageFactory, when put an object and it succeeds, it looks like getting a hash(MD5) of the contents of the object.
In gcs, the hash value of object cannot be obtained at the time of put, and I am troubled with the implementation(
So `putReturnsMd5 test` fails).
I would like advice on this point~

PR #1298, opened 2020-01-08
[Master to v0_10] Support GCS (#1263)

frsyuki:
cherry-pick 9ce9aab


yoyama:
In master (0.9.x), google-auth-library-* are implicitly upgrade to 0.18.0
```com.google.auth:google-auth-library-credentials:0.4.0 -> 0.18.0
```
But in v0_10 (0.10.x), it is still kept as 0.4.0 and it causes library conflict and failed.
So upgrade it explicitly.

yoyama:
#1312 is included.

yoyama (approved):
I checked it in our dev. environment.

PR #1312, opened 2020-01-08
support to generate signedUrl for gcs

TrsNium:
Supports `getDirectDownloadHandle` and` getDirectUploadHandle` methods as well as `s3`.
Test code is not added because `LocalStorageHelper` cannot write for  signedURL test


digdag-storage-gcs/src/main/java/io/digdag/storage/gcs/GCSStorage.java
```diff
@@ -134,6 +141,30 @@ public void list(String objectPrefix, FileListing callback)
         callback.accept(objectSummaryList);
     }
 
+    @Override
+    public Optional<DirectDownloadHandle> getDirectDownloadHandle(String object)
+    {
+        final long secondsToExpire = config.get("direct_download_expiration", Long.class, 10L*60);
```
frsyuki: `config.get` should be done in advance so that ConfigException happens early. Otherwise ConfigException may happen when a REST API called.

PR #1315, opened 2020-01-09
add command executor provider

TrsNium:
I made `CommandExecutorProvider` to select a command executor with `agent.command_executor.type`.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/CommandExecutorModule.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/CommandExecutorModule.java
@@ -1,6 +1,7 @@
 package io.digdag.standards.command;
 
 import com.google.inject.Binder;
+import com.google.inject.Module;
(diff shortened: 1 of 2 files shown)


frsyuki:
@szyn we can make CommandExecutor pluggable (system plugin) using the same approach with #1325 
Especially around https://github.com/treasure-data/digdag/pull/1325/files#diff-268ff650c4f1a5613111cae9872803dfR13-R42
We also need to wrap CommandExecutor in CommandExecutorFactory so that we can lookup CommandExecutor by name like this: https://github.com/treasure-data/digdag/pull/1325/files#diff-ed375f65ccee84dfc2b4850e79f2adf2R10-R11
Making it pluggable is helpful also for us internally because we can override methods of CommandExecutor and inject.

szyn:
Thank you. That's good to know, I will try to make CommandExecutor pluggable.

TrsNium:
Close because the maintainer will take over

PR #1324, opened 2020-01-16
[Master to v0_10] Authenticator SPI

frsyuki:
cherry-pick 66de22c8f5885b10215d913af3403b7fa2641b65 6a95cf019321793b08d3f65823a00ad03e29a896
With #1325.

PR #1325, opened 2020-01-16
Make basic authenticator default

frsyuki:
DO NOT MERGE THIS to master branch. This PR is included in #1324 for v0_10 branch.
Just close this PR once #1324 is merged to v0_10 branch.

It's because this change includes a backward incompatible change in SPI, and not intended to be merged in v0.9.x series.

---

Jwt authenticator is undocumented and not expected to be used. Using it
as the default authenticator doesn't make sense.
This change uses the basic authenticator (added since #1098) as the
default authenticator.

If no configuration is set, behavior is kept same (no authentications
happen).

Current BasicAuthenticator does not support multi-tenancy (multiple
site_id) but it can be added in the future.

This change includes documentation and test changes so that system
plugin mechanism works intentionally for AuthenticatorFactory SPI.

This change is expected to be merged for v0.10 which deprecates admin
API.


yoyama:
This has been included in #1324 . Close.

PR #1326, opened 2020-01-16
[WIP] Graal-js follow up

yoyama:
For test.
Rebase to latest v0_10.
Upgrade shadow plugin to avoid no stop issue.


yoyama:
This follow up conflicted with current v0_10.
So close.

PR #1349, opened 2020-02-03
Fix td_result_export> documents and tests

frsyuki:
`td_result_export>` operator takes parameters from the top-level, not
from the `td_result_export>` key. This fixes tests and documents to
reflect the actual implementation behavior. This doesn't intend to
confirm appropriacy of current implementation behavior.

Test was working because `job_id` parameter was set as a session parameter.
It was working inappropriately.


naritta:
Thank you for fixing. I'll consider about appropriate implementation later.

PR #1356, opened 2020-02-12
Mitigate AWS API rate limit errors in EcsCommandExecutor.

yoyama:
There is a rate limit in calling AWSLogs.getLogEvents() and if it happen, task will fail.
To mitigate the issue, catch the exception and retry getting logs.
Increase retry interval with jitter to disperse API access.


yoyama:
@naritta Thanks for your review and good point. It may be better to reuse RetryExecutor, but requirement is a bit different on interval handling(linear small increasing, jitter). So I don't use it.
But it may be better to refactor in future.

naritta:
I see, thanks for responding, it makes sense.

PR #1359, opened 2020-02-15
Fix KubernetesCommandExecutor for throw an exception if pod exceeds expire of TemporalConfigStorage.

TrsNium:

### DIFF
--- a/digdag-spi/src/main/java/io/digdag/spi/Storage.java
+++ b/digdag-spi/src/main/java/io/digdag/spi/Storage.java
@@ -35,4 +35,8 @@ public interface Storage
     {
         return Optional.absent();
     }
+
+    Long getDirectDownloadExpiration();
+
+    Long getDirectUploadExpiration();
 }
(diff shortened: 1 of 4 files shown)


TrsNium:
https://github.com/treasure-data/digdag/blob/ec9ba0e8882d6334cb271f22890716bd77be81b4/digdag-standards/src/main/java/io/digdag/standards/command/KubernetesCommandExecutor.java#L183
Fail if inTemporalConfigStorage has expired.

https://github.com/treasure-data/digdag/blob/ec9ba0e8882d6334cb271f22890716bd77be81b4/digdag-standards/src/main/java/io/digdag/standards/command/KubernetesCommandExecutor.java#L213
Fail if outTemporalConfigStorage has expired.

yoyama:
@TrsNium I would like to close this PR because we stop working on `v0_10` branch.
The `v0_10` branch has been merged to `master`.
So please reopen a PR on `master` branch if need.

PR #1362, opened 2020-02-19
Fix to retry when TDClient returns 401, 403 and get apikey again

naritta:
Currently we have unstable function to get apikey, so sometimes it fails.
To avoid this, I want to fix to retry when td client return 4xx.
To implement this, I think there are 2 ways like:
・Fix to retry when 4xx and update apikey before retrying.
・create TD operator each time before each operation
https://github.com/treasure-data/digdag/blob/48939b1fdcc26aa52309436af539209007170053/digdag-standards/src/main/java/io/digdag/standards/operator/td/TdDdlOperatorFactory.java#L95

latter is simpler but very redundant because it has to create TD operator for all operation, so I implemented in former way, but I think both has disadvantage. (In current codebase, I think these 2 way is most simple.)

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/td/BaseTDOperator.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/td/BaseTDOperator.java
@@ -1,0 +1,120 @@
+package io.digdag.standards.operator.td;
+
+import com.google.common.base.Throwables;
+import com.treasuredata.client.TDClient;
+import com.treasuredata.client.TDClientException;
+import com.treasuredata.client.TDClientHttpException;
+import io.digdag.spi.SecretProvider;
+import io.digdag.util.RetryExecutor;
+import io.digdag.util.RetryExecutor.RetryGiveupException;
+import org.eclipse.jetty.http.HttpStatus;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.Callable;
+
+import static io.digdag.util.RetryExecutor.retryExecutor;
+
+public class BaseTDOperator
+{
+    private static final Logger logger = LoggerFactory.getLogger(BaseTDOperator.class);
+
+    private static final int INITIAL_RETRY_WAIT = 500;
+    private static final int MAX_RETRY_WAIT = 2000;
+    private static final int MAX_RETRY_LIMIT = 3;
+    public static final int AUTH_MAX_RETRY_LIMIT = 1;
+
+    TDClient client;
+    SecretProvider secrets;
+
+    static final RetryExecutor defaultRetryExecutor = retryExecutor()
+            .withInitialRetryWait(INITIAL_RETRY_WAIT)
+            .withMaxRetryWait(MAX_RETRY_WAIT)
+            .withRetryLimit(MAX_RETRY_LIMIT)
+            .retryIf((exception) -> !isDeterministicClientException(exception));
+
+    public void updateApikey(SecretProvider secrets)
+    {
+        String apikey = TDClientFactory.getApikey(secrets);
+        client = client.withApiKey(apikey);
+    }
+
+    public RetryExecutor authenticatinRetryExecutor() {
+        return retryExecutor()
+                .withInitialRetryWait(INITIAL_RETRY_WAIT)
+                .withMaxRetryWait(MAX_RETRY_WAIT)
+                .withRetryLimit(AUTH_MAX_RETRY_LIMIT)
+                .onRetry((exception, retryCount, retryLimit, retryWait) -> {
+                    logger.debug("apikey will be tried to update by retrying");
+                    updateApikey(secrets);
+                })
+                .retryIf((exception) -> isAuthenticationErrorException(exception));
+    }
+
+    public <T> T callWithRetry(Callable<T> op)
+    {
+        try {
+            return defaultRetryExecutor.run(() -> {
+                try {
+                    return authenticatinRetryExecutor().run(() -> op.call());
+                } catch (RetryGiveupException ex) {
+                    throw Throwables.propagate(ex.getCause());
+                }
+            });
+        }
+        catch (RetryGiveupException ex) {
+            throw Throwables.propagate(ex.getCause());
+        }
+    }
+
+    static boolean isDeterministicClientException(Exception ex)
+    {
+        if (ex instanceof TDClientHttpException) {
+            int statusCode = ((TDClientHttpException) ex).getStatusCode();
+            switch (statusCode) {
+                case HttpStatus.TOO_MANY_REQUESTS_429:
+                case HttpStatus.REQUEST_TIMEOUT_408:
+                    return false;
+                default:
+                    // return true if 4xx
+                    return statusCode >= 400 && statusCode < 500;
+            }
+        }
+        return isFailedBeforeSendClientException(ex);
+    }
+
+    static boolean isFailedBeforeSendClientException(Exception ex)
+    {
+        if (ex instanceof TDClientException) {
+            // failed before sending HTTP request or receiving HTTP response
+            TDClientException.ErrorType errorType = ((TDClientException) ex).getErrorType();
+            switch (errorType) {
+                case INVALID_CONFIGURATION:  // failed to read td.conf, failed to pares integer in properties set to TDClientBuilder, etc.
+                case INVALID_INPUT:          // early table name validation fails, failed to format request body in json, etc.
+                    return true;
+                default:
+                    // other cases such as PROXY_AUTHENTICATION_FAILURE, SSL_ERROR, REQUEST_TIMEOUT, INTERRUPTED, etc.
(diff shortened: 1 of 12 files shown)


digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -408,17 +415,27 @@ private static int exponentialBackoffInterval(DurationInterval pollInterval, int
...
+    static boolean isAuthenticationErrorException(TDClientHttpException ex)
+    {
+        int statusCode = ex.getStatusCode();
+        switch (statusCode) {
+            case HttpStatus.TOO_MANY_REQUESTS_429:
+            case HttpStatus.REQUEST_TIMEOUT_408:
+                return false;
+            default:
+                // return true if 4xx
+                return statusCode >= 400 && statusCode < 500;
```
LeenSun: Will it be better to use httpstatus code instead of 400 ~ 500 ? because not all 4xx are authentication failure

naritta: Thanks for pointing out! I fixed here. https://github.com/treasure-data/digdag/pull/1362/commits/73cdc06b428726998469a599acc253af00472d2e

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -418,7 +442,43 @@ static boolean isDeterministicClientException(Exception ex)
...
+    }
+
+    static boolean isAuthenticationErrorException(TDClientHttpException ex)
+    {
+        int statusCode = ex.getStatusCode();
+        switch (statusCode) {
+            case HttpStatus.UNAUTHORIZED_401:
+            // This is not for authentication basically, but it may be 403 for auth token error. https://tools.ietf.org/html/rfc6750
+            case HttpStatus.FORBIDDEN_403:
+                return true;
```
yoyama: I am not sure 401 and 403 are only authentication errors.
But you have investigated it and as result you set condition, it is ok.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -418,7 +442,43 @@ static boolean isDeterministicClientException(Exception ex)
                     return statusCode >= 400 && statusCode < 500;
             }
         }
-        else if (ex instanceof TDClientException) {
+        return isFailedBeforeSendClientException(ex);
+    }
+
+    static boolean isNeedNotRetryException(Exception ex)
```
yoyama: Not serious, but method name may be difficult to understand. If possible rename is better.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -418,7 +442,43 @@ static boolean isDeterministicClientException(Exception ex)
...
+    {
+        if (ex instanceof TDClientHttpException) {
+            int statusCode = ((TDClientHttpException) ex).getStatusCode();
+            switch (statusCode) {
+                case HttpStatus.TOO_MANY_REQUESTS_429:
+                case HttpStatus.REQUEST_TIMEOUT_408:
+                case HttpStatus.UNAUTHORIZED_401:
+                case HttpStatus.FORBIDDEN_403:
+                    return false;
```
yoyama: I am not sure this condition is correct or not.
If this behavior is different from original, we should clarify the change.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -67,7 +67,13 @@ public static TDOperator fromConfig(BaseTDClientFactory clientFactory, SystemDef
...
+        return new TDOperator(client, database, secrets);
+    }
+
+    public void updateApikey(SecretProvider secrets)
+    {
+        String apikey = TDClientFactory.getApikey(secrets);
+        client = client.withApiKey(apikey);
     }
 
     static final RetryExecutor defaultRetryExecutor = retryExecutor()
```
yoyama: This defaultRetryExecutor is still used in TDJobOperator. Is this OK?

naritta: Thanks, I fixed TDJobOperator also here.
https://github.com/treasure-data/digdag/pull/1362/commits/87cf2ba7c7b3232be7228e896d2e0892a03094cf

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
naritta: I fixed to `isDeterministicException` opposite to `isDeterministicClientException` because it includes non-client(authentication) but deterministic. Please tell me if you have better idea.
https://github.com/treasure-data/digdag/pull/1362/commits/972d0b788f7c0ac101c9f9453472fc4ffbbb6db6

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDJobOperator.java
```diff
@@ -7,59 +7,92 @@
...
+        this.secrets = secrets;
+    }
+
+    void updateApikey(SecretProvider secrets)
+    {
+        String apikey = TDClientFactory.getApikey(secrets);
+        client = client.withApiKey(apikey);
+    }
+
+    RetryExecutor authenticatinRetryExecutor() {
```
yoyama: This method is similar to TDOperator.authenticatinRetryExecutor().
But some options are not set.
Could you explain difference between them ?
Can't we merge and reuse same logic?

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDJobOperator.java
```diff
@@ -7,59 +7,92 @@
...
+            .retryIf((exception) -> isAuthenticationErrorException(exception));
     }
 
     String getJobId()
     {
         return jobId;
     }
 
-    TDJobSummary getJobStatus()
+    private <T> T callWithRetry(Callable<T> op)
```
yoyama: Ditto. It seems complet same logic in TDOperator.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDJobOperator.java
```diff
@@ -7,59 +7,92 @@
...
+    {
+        String apikey = TDClientFactory.getApikey(secrets);
+        client = client.withApiKey(apikey);
+    }
+
+    RetryExecutor authenticatinRetryExecutor() {
+        return defaultRetryExecutor
+            .withRetryLimit(AUTH_MAX_RETRY_LIMIT)
+            .onRetry((exception, retryCount, retryLimit, retryWait) -> {
+                logger.warn("apikey will be tried to update by retrying");
```
yoyama: Logging is good. But _warn_ is the best? In my understand it is not problem so debug or info may be better.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDJobOperator.java
naritta: `updateApikey` update client member in each TDOperator and TDJobOperator, and td-client doesn't have overwrite method for itself, and there is only `withApiKey` which returns new client, I thought I have to update each client member in each class, so I couldn't merge this method with passing client.
Sorry if I'm wrong.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDJobOperator.java
naritta: I couldn't make `authenticatinRetryExecutor` static or merged with above reason, so I couldn't make this like that also.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDJobOperator.java
naritta: I fixed here. https://github.com/treasure-data/digdag/pull/1362/commits/f824f2e8ebed8a9642c28e46caaed67c655b4ae9

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDJobOperator.java
yoyama: Got it.  There is a possibility to fix by using base class. But anyway I understand.

digdag-standards/src/main/java/io/digdag/standards/operator/td/BaseTDOperator.java
```diff
@@ -0,0 +1,120 @@
...
+                    return true;
+                default:
+                    return false;
+            }
+        }
+        else {
+            return false;
+        }
+    }
+}
```
yoyama: No new line?

PR #1366, opened 2020-02-25
Add config option eval.extended-syntax in GraalJS

yoyama:
Currently extended syntax (nested {}) is always enabled in graal.
But it is better to have an option to enable/disable it for compatibility enhancement.
This PR introduces _eval.extended-syntax_ (true|false)  option.
The default is true. So the behavior is same.

This PR also includes unit test for extended syntax and document.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/ConfigEvalEngine.java
@@ -62,7 +62,7 @@ public class ConfigEvalEngine
 
     private static ImmutableList<String> NO_EVALUATE_PARAMETERS = ImmutableList.of("_do",  "_else_do");
 
-    private static enum JsEngineType
+    enum JsEngineType
     {
         NASHORN("nashorn"),
         GRAAL("graal"),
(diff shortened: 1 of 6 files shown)


frsyuki:
Doesn't this spoil eval.js-engine-type=nashorn-graal-check...?
Assuming that all existent workflows assume Nashorn, we need following path to migrate them safely:

1. Set eval.js-engine-type=nashorn-graal-check option (New digdag users should just use the default, eval.js-eval-type=graal).
2. Check reported by the checking
3. If some workflows show significant error reports, fix them to be compatible with Graal.
4. Remove eval.js-engine-type option (use graal).

But with this new option...

1. Set eval.js-engine-type=nashorn-graal-check option...and eval.extended-syntax=false?.
2. Check reported by the checking
3. If some workflows show significant error reports, fix them to be compatible with Graal.
4. Remove eval.js-engine-type option (graal)
5. Set eval.extended-syntax=true

At step 5, workflows will start using extended syntax without migration period. We can't migrate the workflows safely.

yoyama:
@frsyuki Our highest priority is migrating to GraalJS with Java11 smoothly.
I understand extended syntax is convenient. So I keep enable as default in "graal" mode.
But if it causes critical incompatibility in our service, we cannot enable it right now. 
I will do test with false and if no issue happen, I will also do test with true. Then make decision.

Even if we can't enable it as result of test, we can test and improve after migration.
I think nashorn-graal-check is excellent idea and will expand to more general way in the future. 
For example if we can set main=graal, check=graal+extended_syntax, we can check compatibility of extended_syntax. If we want to introduce something incompatibility feature, this way will work.

Anyway I would like to divide large things and apply gradually to avoid critical incidents.

frsyuki:
OK, I see. So,

* (A) You think two smaller migrations with faster deliver of the first migration could be better, while I think one-time migration is better even though the first delivery delays.
* (B) You think extended-syntax could have significantly big impact on backward compatibility, while I don't.

(B) needs data to verify but I agree that verifying it using a working code is reasonable.
(A) depends on how we think about customer/user's manual work involved for the migration.

PR #1368, opened 2020-02-28
Logging dig difinition fetched by http_call> operator

petitviolet:
To make debugging easier, logging digdag definition which is fetched by `http_call>` operator.
Also added a test case for `HttpCallOperatorFactory` to confirm added logging works as expected.

For example, the log looks like:
```
11:09:25.035 [Test worker] INFO  i.d.s.o.HttpCallOperatorFactory - Fetched definition for http_call>:
+sample:
  echo>: 'hello'
```

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/HttpCallOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/HttpCallOperatorFactory.java
@@ -133,6 +133,8 @@ public class HttpCallOperatorFactory
 
             // parse content based on response media type
             String digFileSource = reformatDigFile(content, response.getMediaType(), mediaTypeOverride);
+            // write workflow definition to log
+            logger.info("Fetched definition for http_call>:\n{}", digFileSource);
             // write to http_call.dig file
             Path workflowPath = writeDigFile(digFileSource);
 
--- a/digdag-standards/src/test/java/io/digdag/standards/operator/HttpCallOperatorFactoryTest.java
+++ b/digdag-standards/src/test/java/io/digdag/standards/operator/HttpCallOperatorFactoryTest.java
@@ -1,0 +1,82 @@
+package io.digdag.standards.operator;
+
+import com.github.tomakehurst.wiremock.junit.WireMockRule;
+import io.digdag.client.config.Config;
+import io.digdag.client.config.ConfigUtils;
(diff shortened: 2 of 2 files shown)


digdag-standards/src/main/java/io/digdag/standards/operator/HttpCallOperatorFactory.java
```diff
@@ -133,6 +133,8 @@ public TaskResult runTask()
 
             // parse content based on response media type
             String digFileSource = reformatDigFile(content, response.getMediaType(), mediaTypeOverride);
+            // write workflow definition to log
+            logger.info("Fetched definition for http_call>:\n{}", digFileSource);
```
yoyama: I am afraid of increasing server log.  Current max_stored_response_content_size is 262KB.
So this logging record is near to 262KB in maximum.

petitviolet: @yoyama Thank you for comment.

Even if `config.http_call.max_response_content_size` is too large, I assume log string size could not be a problem, because the log string will be out through `TaskLogger` and its implementations have [some mechanisms to handle huge amount of log strings](https://github.com/treasure-data/digdag/blob/d3fe54b9dfa3a924fceafd31febbdc25851560b8/digdag-core/src/main/java/io/digdag/core/log/BufferedRemoteTaskLogger.java#L63-L65).
In addition, other factories possible to write large volume logs, for instance, `TdOperatorFactory` [writes query statement](https://github.com/treasure-data/digdag/blob/7309575d66fe297e91edf35dce4cba094ecf2a8c/digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java#L236) that may be large.

yoyama:
@petitviolet Thank you for your explanation.
I agree with that task log may not be affected.
I have a concern on not task log but server log.
If I don't misunderstand,  this log may be sent to server log too. In this side I have a concern on the size.
I also agree with another operator also make large log.
But I am not sure about it.
Please wait for a while. I will re-think it.

yoyama:
@petitviolet Sorry for my late reply.
May I ask you to change the PR target from master to v0_10 if you want to use this in our service?

petitviolet:
@yoyama I've updated this PR target to `v0_10`, and also rebased my branch. Thanks!

PR #1378, opened 2020-03-16
Fix TD operator for creating table/database to retry when it does not exist

naritta:


digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -137,6 +137,10 @@ public void ensureDatabaseCreated(String name)
             throws TDClientException
     {
         runWithRetry(() -> client.createDatabase(name), TDClientHttpConflictException.class);
+        // Check database existing and retry only once if it doesn't exit by session error(WM-763)
+        if (!client.existsDatabase(name)){
+            runWithRetry(() -> client.createDatabase(name), TDClientHttpConflictException.class);
```
yoyama: I confirmed to  you, createDatabase() will not fail even though already exist it.
So it is ok.

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -150,6 +154,10 @@ public void ensureTableCreated(String tableName)
     {
         // TODO set include_v=false option
         runWithRetry(() -> client.createTable(database, tableName), TDClientHttpConflictException.class);
+        // Check table existing and retry only once if it doesn't exit by session error(WM-763)
+        if (!client.existsTable(database, tableName)){
+            runWithRetry(() -> client.createTable(database, tableName), TDClientHttpConflictException.class);
```
yoyama: Ditto.

PR #1380, opened 2020-03-23
Fix to update apikey when TD job failed for 401

naritta:
Result export will run after issuing query and finished in worker side, so td-digdag-server can’t get 401 directly, so it doesn’t update apikey even if it’s retried.
So I fixed here to update apikey when job failed for 401 even if it’s job executed in worker like result export.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
@@ -309,6 +309,12 @@ public class TDOperator extends BaseTDOperator
             }
             throw errorPollingException(state, key, jobState, retryInterval);
         }
+        catch (TaskExecutionException ex) {
+            if (ex.getMessage().contains("HTTP request execution failed with code 401")) {
+                updateApikey(secrets);
+            }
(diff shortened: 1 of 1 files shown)


frsyuki:
What does this patch fix...? The patch calls updateApikey which updates “TDClient client” field of this TDOperator instance. Then, it immediately throws `throw Throwables.propagate(ex)`. When is the updated `client` used?

digdag-standards/src/main/java/io/digdag/standards/operator/td/TDOperator.java
```diff
@@ -309,6 +309,12 @@ public TDJobOperator runJob(TaskState state, String key, DurationInterval pollIn
             }
             throw errorPollingException(state, key, jobState, retryInterval);
         }
+        catch (TaskExecutionException ex) {
+            if (ex.getMessage().contains("HTTP request execution failed with code 401")) {
+                updateApikey(secrets);
+            }
+            throw Throwables.propagate(ex);
```
frsyuki: TaskExecutionException is RuntimeException. You don't need Throwables.propagate. Just `throw ex`.

PR #1381, opened 2020-03-24
Add validation for schedule

naritta:
For now, there is no validation for schedule.
So I added that and it fails when there is invalid keys.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java
+++ b/digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java
@@ -101,7 +101,7 @@ public class ProjectControl
     {
         ImmutableList.Builder<ScheduleWithScheduler> schedules = ImmutableList.builder();
         for (StoredWorkflowDefinition def : defs) {
-            Optional<Scheduler> sr = srm.tryGetScheduler(revision, def);
+            Optional<Scheduler> sr = srm.tryGetScheduler(revision, def, true);
             if (sr.isPresent()) {
                 ScheduleTime firstTime = sr.get().getFirstScheduleTime(currentTime);
                 Schedule schedule = Schedule.of(def.getName(), def.getId(), firstTime.getRunTime(), firstTime.getTime());
--- a/digdag-core/src/main/java/io/digdag/core/schedule/SchedulerManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/schedule/SchedulerManager.java
@@ -1,11 +1,18 @@
 package io.digdag.core.schedule;
 
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 import java.util.Map;
 import java.time.ZoneId;
+import java.util.function.Consumer;
+
 import com.google.inject.Inject;
 import com.google.common.base.Optional;
 import com.google.common.collect.ImmutableMap;
+import io.digdag.core.agent.CheckedConfig;
+import io.digdag.core.agent.EditDistance;
 import io.digdag.spi.Scheduler;
 import io.digdag.spi.SchedulerFactory;
 import io.digdag.client.config.Config;
@@ -14,9 +21,23 @@ import io.digdag.core.repository.Revision;
 import io.digdag.core.repository.WorkflowDefinition;
 import io.digdag.core.repository.StoredWorkflowDefinition;
 import io.digdag.core.repository.StoredWorkflowDefinitionWithProject;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static io.digdag.core.schedule.ScheduleExecutor.BUILT_IN_SCHEDULE_PARAMS;
+
+class UnusedConfigException extends ConfigException
+{
+    UnusedConfigException(String message)
+    {
+        super(message);
+    }
+}
 
 public class SchedulerManager
 {
+    private static final Logger logger = LoggerFactory.getLogger(SchedulerManager.class);
+
     private static Optional<Config> tryGetScheduleConfig(WorkflowDefinition def)
     {
         return def.getConfig().getOptional("schedule", Config.class);
@@ -43,26 +64,37 @@ public class SchedulerManager
     public Optional<Scheduler> tryGetScheduler(Revision rev, WorkflowDefinition def)
     {
         return tryGetScheduleConfig(def).transform(it ->
-                    getScheduler(it, def.getTimeZone())
-                );
+                getScheduler(it, def.getTimeZone(), false)
+        );
     }
 
     public Optional<Scheduler> tryGetScheduler(StoredWorkflowDefinitionWithProject def)
     {
         return tryGetScheduleConfig(def).transform(it ->
-                    getScheduler(it, def.getTimeZone())
+                getScheduler(it, def.getTimeZone(), false)
+        );
+    }
+
+    public Optional<Scheduler> tryGetScheduler(Revision rev, WorkflowDefinition def, boolean throwUnusedKeys)
+    {
+        return tryGetScheduleConfig(def).transform(it ->
+                    getScheduler(it, def.getTimeZone(), throwUnusedKeys)
                 );
     }
 
     // used by ScheduleExecutor which is certain that the workflow has a scheduler
     Scheduler getScheduler(StoredWorkflowDefinition def)
     {
-        return getScheduler(getScheduleConfig(def), def.getTimeZone());
+        return getScheduler(getScheduleConfig(def), def.getTimeZone(), false);
     }
 
-    private Scheduler getScheduler(Config schedulerConfig, ZoneId workflowTimeZone)
+    private Scheduler getScheduler(Config schedulerConfig, ZoneId workflowTimeZone, boolean throwUnusedKeys)
     {
-        Config c = schedulerConfig.deepCopy();
+        Set<String> shouldBeUsedKeys = new HashSet<>(schedulerConfig.getKeys());
+        // Track accessed keys using UsedKeysSet class
+        CheckedConfig.UsedKeysSet usedKeys = new CheckedConfig.UsedKeysSet();
+
+        Config c = new CheckedConfig(schedulerConfig, usedKeys);
 
         String type;
         if (c.has("_type")) {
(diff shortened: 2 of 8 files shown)


frsyuki:
This works as long as we have only standard schedules ("hourly>", "daily>", "weekly>", "monthly>", "minutes_interval>", "seconds_interval>", "cron>"). However, digdag's scheduler is designed to be pluggable. Can't we have the validation without spoiling it? To make it possible, ProjectResource.putProject needs to validate pushed workflow WITHOUT using static keys (VALID_SCHEDULE_KEYS).

A direction is commented at [ProjectControl.updateSchedules](https://github.com/treasure-data/digdag/blob/v0_10/digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java#L112-L114):
```
        // TODO validate workflows and sessions
        //   * compile workflow
        //   * validate SubtaskMatchPattern
```

ProjectControl.updateSchedules is called by ProjectControl.insertWorkflowDefinitions, and it's called by ProjectResource.putProject. It means that `PUT /api/projects` returns an error if ProjectControl.updateSchedules throws an exception. Therefore, an alternative direction is adding validation to ProjectControl.updateSchedules so that it throws an exception if schedule definition includes unknown keys.

Now, ProjectControl.updateSchedules is creating a `Scheduler` instance at https://github.com/treasure-data/digdag/blob/0b1ca9c237b37504d1b34b38b3f6350b81c6f2e7/digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java#L104. `Scheduler` is either of CronScheduler, DailyScheduler, etc. implemented in digdag-standards. [SchedulerManager is creating a Scheduler instance](https://github.com/treasure-data/digdag/blob/v0_10/digdag-core/src/main/java/io/digdag/core/schedule/SchedulerManager.java#L43-L90).

As you can see SchedulerManager.getScheduler there, the code is similar to [this code in OperatorManager](https://github.com/treasure-data/digdag/blob/v0_10/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java#L227-L278). It gets `"_type"` from config, gets SchedulerFactory, and call SchedulerFactory.newScheduler to create a new `Scheduler`. A big difference is that SchedulerManager.getScheduler doesn't have checking of unused keys. Checking of unused keys is implemented in [OperatorManager here](https://github.com/treasure-data/digdag/blob/v0_10/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java#L262-L278).

naritta:
Thank you for reviewing! I'll fix it. 🙇

frsyuki:
So, how about doing this?

1. Change SchedulerFactory.tryGetScheduler and .getScheduler to get another argument - `Consumer<Set<String>> unusedKeysCallback`.
2. SchedulerFactory.getScheduler calls unusedKeysCallback.accept if some config keys are not used.
3. ProjectControl.updateSchedules calls tryGetScheduler with unusedKeysCallback=(unusedKeys) -> throw new SomeSortOfException.

Item 2 can be implemented by mimicking the same code with [OperatorManager here](https://github.com/treasure-data/digdag/blob/v0_10/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java#L262-L278):

1. Create `usedKeys`: CheckedConfig.UsedKeysSet usedKeys = new CheckedConfig.UsedKeysSet()
2. Wrap Config with CheckedConfig (Note: CheckedConfig implements Config).
3. Give the CheckedConfig to factory.newScheduler.
4. Thanks to CheckedConfig, `usedKeys` contains list of used keys now.
5. Compare usedKeys with shouldBeUsedKeys like here: https://github.com/treasure-data/digdag/blob/v0_10/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java#L273-L278
6. shouldBeUsedKeys is all keys of Config like this: https://github.com/treasure-data/digdag/blob/v0_10/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java#L225

naritta:
@frsyuki Thank you very much for clear explanation. 
I fixed like above, so could you review when you have time?

digdag-core/src/main/java/io/digdag/core/schedule/SchedulerManager.java
```diff
@@ -82,10 +94,47 @@ private Scheduler getScheduler(Config schedulerConfig, ZoneId workflowTimeZone)
             c.set("_command", command);
         }
 
+        for(String param : BUILT_IN_SCHEDULE_PARAMS){
+            if (param.equals("skip_on_overtime"))
```
frsyuki, line 98: You don't need these `if` statements. You can use following instead:
```suggestion
            usedKeys.add(param);
```

digdag-core/src/main/java/io/digdag/core/schedule/SchedulerManager.java
```diff
@@ -82,10 +94,47 @@ private Scheduler getScheduler(Config schedulerConfig, ZoneId workflowTimeZone)
...
+
+            List<String> suggestions = EditDistance.suggest(key, candidateKeys, 0.50);
+            if (suggestions.isEmpty()) {
+                throw new ConfigException(buf.toString());
+            }
+            else {
+                buf.append("> Did you mean '");
+                buf.append(suggestions);
+                buf.append("'");
+                throw new ConfigException(buf.toString());
```
frsyuki, line 136: Throwing exception here is dangerous because:

* Schedules already saved in DB might already have unused keys. Those schedules are currently running in production.
* If here throws an exception, such schedules will stop working (ScheduleExecutor.runSchedule is calling SchedulerManager.getScheduler).
* From user's point of view, they can't notice that their schedules have stopped (because there are no notifications). It's unexpected from the users.
* From our point of view, ScheduleExecutor.runSchedule starts logging exceptions. It's not UncaughtException, but almost.

So, unused keys checking must be done only when a new workflow is uploaded. And that's good enough.

frsyuki: To make it possible, unused keys checking in this getScheduler method needs to be optional.
A possible way is adding `boolean checkUnusedKeys` argument to this getScheduler method. Alternative way is adding `Optional<Consumer<Map<String, List<String>>>> unusedKeysCallback` argument to this getScheduler method (do the unused keys checking if unusedKeysCallback.isPresent()).
Either way works. I think either is fine.

frsyuki: Another way I came up right now is adding a class like `UnusedConfigExceptions extends ConfigException`.
...either way works.

digdag-tests/src/test/java/acceptance/SchedulerIT.java
```diff
@@ -42,4 +44,30 @@ public void testScheduler()
...
+            throws Exception
+    {
+        try {
+            copyResource("acceptance/scheduler/invalid_op.dig", root().resolve("test.dig"));
+            main("scheduler", "--project", root().toString());
+            fail();
+        } catch (Throwable ie) {
+            assertThat(ie.getCause().getCause().toString(), Matchers.containsString("Parameter 'wrong>' is not used at schedule."));
+        }
+    }
```
frsyuki, line 72: can you add some more tests?:
* A test that verifies REST API response code. I think API response code should be 422. It's very bad if it's 5xx. If API response code is unexpected, additional code will be necessary around ProjectResource.validateWorkflowAndSchedule.
* what if unknown scheduler type is set (like `no_such_scheduler>: abc`)

frsyuki: you can copy testing mechanism from ServerErrorStatusCodeIT to this SchedulerIT.

digdag-core/src/main/java/io/digdag/core/schedule/SchedulerManager.java
naritta: I fixed to use unusedKeysCallback and UnusedConfigExceptions, but it seems like same as `boolean checkUnusedKeys`, so sorry if I'm misunderstanding your intension.
https://github.com/treasure-data/digdag/pull/1381/commits/db6f9434eea7980ff8fafdc3da1f04ac5359ab3b

digdag-core/src/main/java/io/digdag/core/schedule/SchedulerManager.java
naritta: Thanks, I fixed to use that.
https://github.com/treasure-data/digdag/pull/1381/commits/e9740105d2abb8eae39c1bab699ee0816fdf4de8

digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java
```diff
@@ -101,7 +103,7 @@ private void updateSchedules(
     {
         ImmutableList.Builder<ScheduleWithScheduler> schedules = ImmutableList.builder();
         for (StoredWorkflowDefinition def : defs) {
-            Optional<Scheduler> sr = srm.tryGetScheduler(revision, def);
+            Optional<Scheduler> sr = srm.tryGetScheduler(revision, def, Optional.of(getUnusedKeysCallback()));
```
frsyuki, line 106: if we throw UnusedConfigException, this doesn't have to be a callback. A boolean flag is good enough.
Callback is useful if ProjectResource.putProject creates a callback object because ProjectResource.putProject can decide the type of the exception class to throw.
I think boolean flag is good enough.

digdag-tests/src/test/java/acceptance/SchedulerIT.java
naritta: I added tests for them here. 
https://github.com/treasure-data/digdag/pull/1381/commits/cca6e1f7ee3f7507b0948448ae5212676acf0e19

digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java
naritta: I see, thanks, then I fixed to use boolean for that.
https://github.com/treasure-data/digdag/pull/1381/commits/74d9f4e42d7cbbcf8f163dce4edb3d361bd68a5b

naritta:
@frsyuki I fixed like the comment and added more tests, so please tell me if there are problem. :man-bowing:

digdag-tests/src/test/java/acceptance/SchedulerIT.java
```diff
@@ -42,4 +54,71 @@ public void testScheduler()
...
+
+    @Test
+    public void testInvalidScheduler()
+            throws Exception
+    {
+        try {
+            copyResource("acceptance/scheduler/invalid.dig", root().resolve("test.dig"));
+            main("scheduler", "--project", root().toString());
+            fail();
+        } catch (Throwable ie) {
```
frsyuki, line 66:
```suggestion
        }
        catch (Throwable ie) {
```

digdag-tests/src/test/java/acceptance/SchedulerIT.java
```diff
@@ -42,4 +54,71 @@ public void testScheduler()
...
+
+    @Test
+    public void testInvalidOpScheduler()
+            throws Exception
+    {
+        try {
+            copyResource("acceptance/scheduler/invalid_op.dig", root().resolve("test.dig"));
+            main("scheduler", "--project", root().toString());
+            fail();
+        } catch (Throwable ie) {
```
frsyuki, line 79:
```suggestion
        }
        catch (Throwable ie) {
```

digdag-core/src/main/java/io/digdag/core/schedule/SchedulerManager.java
```diff
@@ -82,10 +114,44 @@ private Scheduler getScheduler(Config schedulerConfig, ZoneId workflowTimeZone)
...
         if (factory == null) {
             throw new ConfigException("Unknown scheduler type: " + type);
         }
+
+        if (!usedKeys.isAllUsed()) {
+            shouldBeUsedKeys.removeAll(usedKeys);
+            if (!shouldBeUsedKeys.isEmpty()) {
+                for (String key: shouldBeUsedKeys) {
+                    if (throwUnusedKeys)
+                        throw new UnusedConfigException(getWarnUnusedKey(key, usedKeys));
```
frsyuki, line 131: Showing logging in this loop is good but throwing exception has a downside. Because throwing an exception will leave from this loop, users can know only one unused key at a time.
How about moving `throw` before the loop if `throwUnusedKeys` is set? We can concatenate messages of `getWarnUnusedKey` with `\n` as the message of UnusedConfigException.

naritta:
Thanks for reviewing! 🙇

PR #1390, opened 2020-04-08
Add validation for sla operator

naritta:
For now, there is no validation for pushing workflows with sla. 
This PR validation for sla operator.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java
+++ b/digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java
@@ -13,6 +13,8 @@ import io.digdag.core.schedule.ScheduleStoreManager;
 import io.digdag.core.schedule.SchedulerManager;
 import io.digdag.core.schedule.ScheduleExecutor;
 import io.digdag.client.config.ConfigException;
+import io.digdag.core.session.SessionMonitor;
+import io.digdag.core.workflow.SlaCalculator;
 import io.digdag.spi.ScheduleTime;
 import io.digdag.spi.Scheduler;
 import java.util.stream.Collectors;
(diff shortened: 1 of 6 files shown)


digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java
```diff
@@ -69,6 +71,7 @@ public void insertRevisionArchiveData(int revId, byte[] data)
     {
         List<StoredWorkflowDefinition> list = insertWorkflowDefinitionsWithoutSchedules(revision, defs);
         updateSchedules(revision, list, srm, currentTime);
+        validateSla(defs);
```
yoyama, line 74: Just in case, if validation fail, updateSchedules() has been cancelled ? 
If this method is only called in transaction and updateSchedules() has no side effect except Pg, I guess it is OK.

naritta: Thank you for checking, I think I should move this before inserting actually, so I fixed it.
https://github.com/treasure-data/digdag/pull/1390/commits/a1ea9c6584d5e08776adb5056851eaec415f0dc0

digdag-tests/src/test/resources/acceptance/sla/invalid_config.dig
```diff
@@ -0,0 +1,7 @@
+timezone: UTC
+
+sla:
+  duration2: 00:00:05
```
yoyama, line 4: Just in case. If sla contain both `duration:` and 'duration2:'  is it error ?

digdag-core/src/main/java/io/digdag/core/workflow/SlaCalculator.java
```diff
@@ -13,22 +17,53 @@
...
 import io.digdag.client.config.Config;
 import io.digdag.client.config.ConfigException;
+import io.digdag.core.agent.CheckedConfig;
+import io.digdag.core.agent.EditDistance;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import static java.util.Locale.ENGLISH;
 
+class UnusedConfigException extends ConfigException
```
yoyama, line 27: This exception is defined as package private.
I am not sure how and where this exception is handled.
If it is only processed in same package, it is ok.

naritta:
Thank you for reviewing, I added more tests and I'll merge this.

PR #1392, opened 2020-04-13
Implement /api/schedules/{id}/enable_by_mode

szyn:
This is the PR for adding a new endpoint `/api/schedules/{id}/enable_by_mode`, which is accepting mode when enabling the schedule. 
If a mode is specified, the schedule will be updated based on the mode. 
If a mode is not specified, re-run all past sessions (= same as current behavior)

For example:
- mode: schedule => Start running as scheduled during the most recent session
- mode: next_schedule => Start running as scheduled during the next session

Note that specifying the `next_schedule` might be different from the actual next schedule, the `schedule` will be the actual one.
Please see also [my comment](https://github.com/treasure-data/digdag/blob/f3e822640cfccc8e3637243331338116d9df3829/digdag-server/src/main/java/io/digdag/server/rs/ScheduleResource.java#L333-L339).

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/client/EnableSchedule.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/client/EnableSchedule.java
@@ -1,13 +1,16 @@
 package io.digdag.cli.client;
 
+import com.beust.jcommander.Parameter;
 import com.google.common.base.Optional;
 import io.digdag.cli.SystemExitException;
 import io.digdag.client.DigdagClient;
 import io.digdag.client.api.Id;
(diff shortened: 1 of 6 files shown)


yoyama:
I guess we had already discussed on adding new end point is good or not.
But could you clarify the reason for it again in your comment?
Especially if mode parameter is not set, the behavior is equivalent by your comment. 
So it seems we can merge the functionality to original end point "/api/schedules/{id}/enable" .

szyn:
@yoyama Thank you for pointing out. As you said, this functionality can be merged into `/api/schedules/{id}/enable`. I will close this one and create another PR.

PR #1398, opened 2020-04-21
Add attempt status response

dzannotti:
Add flag of attempt status, so we remove logic from cli

### DIFF
--- a/digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
+++ b/digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
@@ -10,13 +10,11 @@ import org.immutables.value.Value;
 
 @Value.Immutable
 @JsonDeserialize(as = ImmutableRestSessionAttempt.class)
-public interface RestSessionAttempt
-{
+public interface RestSessionAttempt {
     Id getId();
 
     @Value.Default
-    default int getIndex()
-    {
+    default int getIndex() {
         // index has default value only to keep backward compatibility.
         // index is added since 01714615088e315ba401f4deca41b6dd58bb349b.
         return 0;
@@ -24,7 +22,7 @@ public interface RestSessionAttempt
 
     IdAndName getProject();
 
-    //Optional<String> getRevision();
+    // Optional<String> getRevision();
 
     NameOptionalId getWorkflow();
 
@@ -61,8 +59,7 @@ public interface RestSessionAttempt
         return "running";
     }
 
-    static ImmutableRestSessionAttempt.Builder builder()
-    {
+    static ImmutableRestSessionAttempt.Builder builder() {
         return ImmutableRestSessionAttempt.builder();
     }
 }
--- a/digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
+++ b/digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
@@ -50,15 +50,15 @@ public interface RestSessionAttempt
 
     default String getStatus() {
         if (this.getCancelRequested()) {
-            return "KILLED";
+            return "killed";
         }
         if (this.getSuccess()) {
-            return "SUCCESS";
+            return "success";
         }
-        if (!this.getSuccess() && this.getDone()) {
-            return "ERROR";
+        if (this.getDone()) {
+            return "error";
         }
-        return "RUNNING";
+        return "running";
     }
 
     static ImmutableRestSessionAttempt.Builder builder()
--- a/digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
+++ b/digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
@@ -48,6 +48,19 @@ public interface RestSessionAttempt
 
     Optional<Instant> getFinishedAt();
 
+    default String getStatus() {
+        if (this.getCancelRequested()) {
+            return "KILLED";
+        }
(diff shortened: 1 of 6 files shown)


digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
```diff
@@ -48,6 +48,19 @@ default int getIndex()
 
     Optional<Instant> getFinishedAt();
 
+    default String getStatus() {
```
dzannotti, line 51: would an enum be better here?

digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
```diff
@@ -48,6 +48,19 @@ default int getIndex()
...
+        if (this.getCancelRequested()) {
+            return "KILLED";
+        }
+        if (this.getSuccess()) {
+            return "SUCCESS";
+        }
+        if (!this.getSuccess() && this.getDone()) {
```
komamitsu, line 58: `!this.getSuccess()` isn't needed?

digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
komamitsu: What would the enum be like?

digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
```diff
@@ -48,6 +48,19 @@ default int getIndex()
 
     Optional<Instant> getFinishedAt();
 
+    default String getStatus() {
+        if (this.getCancelRequested()) {
+            return "KILLED";
+        }
+        if (this.getSuccess()) {
+            return "SUCCESS";
```
komamitsu, line 56: If you don't have a strong opinion for using an uppercase string as a return value, why don't you return a lowercase string to keep consistency with the existing behaviour?

digdag-docs/src/_extra/api/swagger.yaml
```diff
@@ -1239,6 +1239,13 @@ definitions:
       cancelRequested:
         type: "boolean"
         default: false
+      status:
+        type: "string"
+        enum:
+        - "SUCCESS"
```
komamitsu, line 1245: Don't need to change these to lowercases?

digdag-client/src/main/java/io/digdag/client/api/RestSessionAttempt.java
```diff
@@ -48,6 +48,19 @@ default int getIndex()
 
     Optional<Instant> getFinishedAt();
 
+    default String getStatus() {
+        if (this.getCancelRequested()) {
```
komamitsu, line 52: Ah, it could happen that a task finishes while it received a cancel request right before the finish...? @yoyama 

In this case, I think it should show `success` / `error`.

komamitsu: I checked recent 5000000 records of `session_attempts` table in our service

```
 state_flags | count
-------------+--------
           6 | 588183
           2 |  21313
           3 |    635
           0 |    136
```

6 means `success`, 3 means `killed`, 2 means `error` and 0 means `running`. So for now, `success` status that already received a cancel request doesn't exist. But I think it won't be strongly guaranteed on server side design in the future and the change in this PR should be a bit more robust for future changes to some extents. So I'd like to ask you to move `if (this.getCancelRequested()) {...}` into `if (this.getDone()) {...}`.

BTW, I talked with @yoyama about this logic and we thought `cancel requested` and another state should be orthogonal as much as possible. So `success (cancel requested)` or `running (cancel requested)` might need to be added in the future.

dzannotti: @komamitsu is this what you meant?

komamitsu: I meant it would be better if the condition is like
```
       if (this.getSuccess()) {
            return "success";
        }
        if (this.getDone()) {
            if (this.getCancelRequested()) {
                return "killed";
            }
            else {
                return "error";
            }
        }
        return "running";
```

digdag-client/src/test/java/io/digdag/client/api/ModelCompatibilityTest.java
```diff
@@ -67,4 +67,96 @@ private String removeField(String json, String field)
...
+        RestSessionAttempt attempt = RestSessionAttempt.builder()
+            .id(Id.of("1"))
+            .index(3)
+            .project(IdAndName.of(Id.of("1"), "p"))
+            .workflow(NameOptionalId.of("w", Optional.of(Id.of("1"))))
+            .sessionId(Id.of("1"))
+            .sessionUuid(UUID.randomUUID())
+            .sessionTime(OffsetDateTime.now(ZoneId.of("UTC")))
+            .retryAttemptName(Optional.absent())
+            .done(false)
```
komamitsu, line 130: In current implementation, `killed` is printed out only when the session attempt has been done...?

komamitsu: Ah, found this one https://ci.appveyor.com/project/treasure-data/digdag/builds/33292140/job/ndb3n2q3x25i9wj3#L967.

Let me tweak it...

komamitsu:
@dzannotti Thanks!

PR #1399, opened 2020-04-21
Add validation for sla operator

naritta:
https://github.com/treasure-data/digdag/pull/1390 has bugs for checking subtasks by removing key in for each itelator. So I fixed not to use remove there.
I moved subtask check to here
https://github.com/treasure-data/digdag/pull/1399/files#diff-58d878f293bcecaeab95c78eb1279581R66-R69
from
https://github.com/treasure-data/digdag/pull/1390/files#diff-58d878f293bcecaeab95c78eb1279581R104-R107
And added some more tests.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java
+++ b/digdag-core/src/main/java/io/digdag/core/repository/ProjectControl.java
@@ -13,6 +13,8 @@ import io.digdag.core.schedule.ScheduleStoreManager;
 import io.digdag.core.schedule.SchedulerManager;
 import io.digdag.core.schedule.ScheduleExecutor;
 import io.digdag.client.config.ConfigException;
+import io.digdag.core.session.SessionMonitor;
+import io.digdag.core.workflow.SlaCalculator;
 import io.digdag.spi.ScheduleTime;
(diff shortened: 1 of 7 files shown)


naritta:
This PR (https://github.com/treasure-data/digdag/pull/1390) has a bug, because there is `remove` in for-each. https://github.com/treasure-data/digdag/pull/1390/files#diff-58d878f293bcecaeab95c78eb1279581R106
We cannot use element remove in for-each in Java. 
https://stackoverflow.com/questions/1196586/calling-remove-in-foreach-loop-in-java

So I fixed to remove elements which should be removed (here subtask) by adding them to usedKeys.
https://github.com/treasure-data/digdag/pull/1399/files#diff-58d878f293bcecaeab95c78eb1279581R66-R69
And I added additional tests for that. https://github.com/treasure-data/digdag/pull/1399/files#diff-59274f82dd5e523c8926ef3a15c17cc0R143
There is no changes from previous PR (https://github.com/treasure-data/digdag/pull/1390) for other things.

yoyama:
@naritta Could you fix CI error?

yoyama:
Close because of no activity long time.
We will re-visit if needed.

PR #1401, opened 2020-04-23
Add support for skipping past sessions on /api/schedules/{id}/enable

szyn:
Add support for skipping past sessions on existing endpoint `/api/schedules/{id}/enable`.
We decided to merge this functionality with the existing endpoints. #1392

Make this endpoint to accept `skipSchedule` and `nextTime` requests when enabling the schedule.

- skipSchedule: false -> re-run all past sessions between paused and resumed. (= same as current behavior)
- skipSchedule: true -> the schedule will be skipped (updated) to currentTime (default) / nextTime (optional, ISO8601 format). It means don't run all previously scheduled sessions.

And I have a plan to update CLI and document as another PR.

### DIFF
--- a/digdag-client/src/main/java/io/digdag/client/DigdagClient.java
+++ b/digdag-client/src/main/java/io/digdag/client/DigdagClient.java
@@ -26,6 +26,7 @@ import io.digdag.client.api.RestRevisionCollection;
 import io.digdag.client.api.RestSchedule;
 import io.digdag.client.api.RestScheduleCollection;
 import io.digdag.client.api.RestScheduleBackfillRequest;
+import io.digdag.client.api.RestScheduleEnableRequest;
 import io.digdag.client.api.RestScheduleSkipRequest;
 import io.digdag.client.api.RestScheduleSummary;
 import io.digdag.client.api.RestSecret;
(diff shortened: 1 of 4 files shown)


digdag-cli/src/main/java/io/digdag/cli/client/EnableSchedule.java
```diff
@@ -97,7 +101,9 @@ private void enableSchedule(Id scheduleId)
             throws IOException, SystemExitException
     {
         DigdagClient client = buildClient();
-        client.enableSchedule(scheduleId);
+        Optional<String> mode = modeString == null ? Optional.absent() : Optional.of(modeString);
```
yoyama, line 104: Just trivial things, but you may use Optional.fromNullable()

digdag-cli/src/main/java/io/digdag/cli/client/EnableSchedule.java
```diff
@@ -15,10 +16,15 @@
...
 {
+    @Parameter(names = {"-m", "--mode"})
+    String modeString = null;
+
     @Override
     public SystemExitException usage(String error)
     {
         err.println("Usage: " + programName + " enable <schedule-id> | <project-name> [name]");
+        err.println("  Options:");
+        err.println("    -m, --mode MODE                  update schedules with this mode for skipping sessions (default: none)");
```
yoyama, line 27: It is better to something explain about MODE. In addition we may need document for it.

digdag-cli/src/main/java/io/digdag/cli/client/EnableSchedule.java
szyn: Thank you, that's good to know. But I decided not to use this mode, so I'm going to mark as resolve this conversation.

digdag-cli/src/main/java/io/digdag/cli/client/EnableSchedule.java
szyn: Ditto.

PR #1405, opened 2020-05-15
Implement project_name param in require>.

yoyama:
Introduce **project_name** parameter in **require>** operator.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
@@ -165,7 +165,7 @@ public class InProcessTaskCallbackApi
     public StoredSessionAttempt startSession(
             OperatorContext context,
             int siteId,
-            int projectId,
+            ProjectIdentifier projectIdentifier,
             String workflowName,
             Instant instant,
             Optional<String> retryAttemptName,
@@ -177,7 +177,10 @@ public class InProcessTaskCallbackApi
 
                             //TODO check permission by AccessController for 'require' operator over another project.
 
-                            StoredProject proj = projectStore.getProjectById(projectId);
+                            StoredProject proj = projectIdentifier.byId() ?
+                                projectStore.getProjectById(projectIdentifier.getId()) :
+                                projectStore.getProjectByName(projectIdentifier.getName());
+
                             StoredWorkflowDefinitionWithProject def =
                                     projectStore.getLatestWorkflowDefinitionByName(proj.getId(), workflowName);
 
--- a/digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
@@ -6,6 +6,7 @@ import io.digdag.client.config.Config;
 import io.digdag.client.config.ConfigElement;
 import io.digdag.client.config.ConfigException;
 import io.digdag.client.config.ConfigFactory;
+import io.digdag.core.agent.TaskCallbackApi.ProjectIdentifier;
 import io.digdag.core.repository.ResourceLimitExceededException;
 import io.digdag.core.repository.ResourceNotFoundException;
 import io.digdag.core.session.StoredSessionAttempt;
@@ -13,13 +14,11 @@ import io.digdag.spi.Operator;
 import io.digdag.spi.OperatorContext;
 import io.digdag.spi.OperatorFactory;
 import io.digdag.spi.TaskExecutionException;
-import io.digdag.spi.TaskRequest;
 import io.digdag.spi.TaskResult;
 import io.digdag.util.BaseOperator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.nio.file.Path;
 import java.time.Instant;
 
 import static java.util.Locale.ENGLISH;
@@ -66,16 +65,17 @@ public class RequireOperatorFactory
         {
             Config config = request.getConfig();
             String workflowName = config.get("_command", String.class);
-            int projectId = config.get("project_id", int.class);
             Instant instant = config.get("session_time", Instant.class);
             boolean ignoreFailure = config.get("ignore_failure", boolean.class, false);
             Optional<String> retryAttemptName = config.getOptional("retry_attempt_name", String.class);
             Config overrideParams = config.getNestedOrGetEmpty("params");
             try {
+                ProjectIdentifier projectIdentifier = makeProjectIdentifier();
+
                 StoredSessionAttempt attempt = callback.startSession(
                         context,
                         request.getSiteId(),
-                        projectId,
+                        projectIdentifier,
                         workflowName,
                         instant,
                         retryAttemptName,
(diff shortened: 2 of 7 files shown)


digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -103,5 +103,34 @@ public TaskResult runTask()
...
+         */
+        private ProjectIdentifier makeProjectIdentifier()
+        {
+            Config config = request.getConfig();
+            Config localConfig = request.getLocalConfig();
+            int projectId = config.get("project_id", int.class);
+            Optional<Integer> projectIdParam = localConfig.getOptional("project_id", Integer.class);
+            Optional<String> projectNameParam = localConfig.getOptional("project_name", String.class);
+
+            if (projectIdParam.isPresent() && projectNameParam.isPresent()) {
```
komamitsu, line 120: Do you think it's reasonable to do the same check in compile time when a workflow is pushed?

digdag-core/src/main/java/io/digdag/core/agent/TaskCallbackApi.java
```diff
@@ -34,10 +34,30 @@ void retryTask(TaskRequest request, AgentId agentId,
...
+            ProjectIdentifier projectIdentifier,
             String workflowName,
             Instant instant,
             Optional<String> retryAttemptName,
             Config overrideParams)
-        throws ResourceNotFoundException, ResourceLimitExceededException;
+            throws ResourceNotFoundException, ResourceLimitExceededException;
+
+    // Identifier of Project: id or name
+    class ProjectIdentifier
```
komamitsu, line 45: Probably we can add `static` keyword to this

digdag-docs/src/operators/require.md
```diff
@@ -60,6 +60,17 @@
...
+
+  Name of another project. You can kick another project's workflow by setting this parameter.
+
+  Examples:
+
+  ```
+  require>: another_project_wf
+  project_name: another_project
+  ```
+
```
komamitsu, line 73: Maybe it's worth explaining what will happen if the target project doesn't exist.

digdag-tests/src/test/resources/acceptance/require/parent_another_project.dig
```diff
@@ -0,0 +1,7 @@
++require_by_project_id:
+  require>: child_another_project
+  project_id: ${child_project_id}
```
komamitsu, line 3: This string `${child_project_id}` seems as a template parameter, but it's just replaced in the test code, right? How about following an existing case
https://github.com/treasure-data/digdag/blob/c00384a7f96a373a458d805d5c5a8ef8459a5ad1/digdag-tests/src/test/resources/acceptance/require/child.dig#L2 ? I don't insist on it, though 😺

digdag-core/src/main/java/io/digdag/core/agent/TaskCallbackApi.java
yoyama: In interface we don't need public and static keyword in my understand.
IntelliJ suggests as it and jls may also explain it.

https://stackoverflow.com/questions/2400828/inner-class-within-interface
https://docs.oracle.com/javase/specs/jls/se8/html/jls-9.html#jls-9.5

digdag-docs/src/operators/require.md
yoyama: Got it. Will add.

digdag-tests/src/test/resources/acceptance/require/parent_another_project.dig
yoyama: Good point. I found another example which used ${} and followed it.
https://github.com/treasure-data/digdag/blob/master/digdag-tests/src/test/java/acceptance/RetryIT.java#L168-L170

I hesitated to judge which is better and don't have strong opinion.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Hmm I am not sure but I will check it later.
If require> has already validation in compile, I will add it. But if not, it may be another improvement in my understand. I will check it later.

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
```diff
@@ -177,7 +177,10 @@ public StoredSessionAttempt startSession(
 
                             //TODO check permission by AccessController for 'require' operator over another project.
 
-                            StoredProject proj = projectStore.getProjectById(projectId);
+                            StoredProject proj = projectIdentifier.byId() ?
```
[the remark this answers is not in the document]
yoyama: I am not sure adding new method to ProjectStore is good or not. It is only combination of existing method.
The reason of ProjectIdentifier is transactional processing in startSession().
if RequireOperator converts project_name to project_id, I don't need to change startSession() signature and don't need to define ProjectIdentifier.
My concerns in this case are
 1. It is not a transaction
 2. RequireOperator need to inject ProjectStore
     (require> is defined in digdag-core and special operator, so may be acceptable)
What do you think about it ?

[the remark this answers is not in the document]
komamitsu: Current change in this PR looks good. So my weak suggestion is either of these:
1. Just add a comment like `This code should be moved to ProjectStore when the same code is required somewhere else`
2. Move the code to ProjectStore

Either is fine with me 😉

digdag-core/src/main/java/io/digdag/core/agent/TaskCallbackApi.java
komamitsu: Oh, sorry. I thought the file is for `class`.

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskCallbackApi.java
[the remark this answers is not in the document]
yoyama: I added a comment in ProjectIdentifier as of discussion in person.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Currently it seems no validation on operators in check. So I would like to skip it.

digdag-docs/src/operators/require.md
yoyama: I updated the document.

digdag-tests/src/test/resources/acceptance/require/parent_another_project.dig
yoyama: I changed the name of template parameters as disucssion.

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +189,55 @@ public void testIgnoreProjectIdParam()
...
+                "-e", server.endpoint());
+        assertThat(pushParentStatus.errUtf8(), pushParentStatus.code(), is(0));
+
+        CommandStatus startStatus = main("start",
+                "-c", config.toString(),
+                "-e", server.endpoint(),
+                "parent_another", "parent_another_project",
+                "--session", "now"
+        );
+        assertThat(startStatus.errUtf8(), startStatus.code(), is(0));
```
komamitsu, line 237: Does this test check if the child project workflow has run as expected?

yoyama: Oh, it is not enough.
Need to check the result of attempt. (must be success).
I will add check code.

komamitsu: Thanks!

yoyama: To make test robust, I separated the project_id and project_name test and check attempt result both parent and its child.

- testRequireToAnotherProjectById()
- testRequireToAnotherProjectByName()

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +193,89 @@ public void testIgnoreProjectIdParam()
...
+                "-c", config.toString(),
+                "-e", server.endpoint());
+        assertThat(pushChildStatus.errUtf8(), pushChildStatus.code(), is(0));
+
+        // extract child project id
+        Matcher m = Pattern.compile(".*\\s+id:\\s+(\\d+).*").matcher(pushChildStatus.outUtf8());
+        assertThat(m.find(), is(true));
+        String childProjectId = m.group(1);
+
+        // Push parent project with project_id: xxx
```
komamitsu, line 236: Which to use depends on `useProjectId`?

yoyama: Thanks, I will fix the comment.

yoyama:
#1409 include this fix. So close.

PR #1409, opened 2020-05-26
require> enhancement

### LINKED ISSUE #712: require operator does not honor past session's success/failure when retrying
I am using [require operator](http://docs.digdag.io/operators/require.html) to check the status of the previous session of the same workflow. The workflow should -
1. continue if the last session was successful,
2. fail if the last session failed,
3. start last session if the last session does not exist, and,
4. wait for the last session's completion if it's pending.

The above four conditions work perfectly when the session is started by the scheduler or is started using `digdag start` command without the `--retry` option. 

But when I retry a session (either successful or failed) using `digdag retry` or `digdag start --retry` or `digdag backfill`, the require operator starts the last session (condition 3 above) regardless of the status of the last session. In other words, last session is started again even if its status is successful or failed.

yoyama:
This PR is follow up PR to #1405 and enhance require> operator.

- Introduce rerun_on: (none,failed,all).
- Ignore retry_attempt_name from out of operator. (fix #712 )
- exponential backoff for task retry (max 10 secs)

To exclude #1405 fix from file changes, following view may be good.
https://github.com/treasure-data/digdag/pull/1409/files/48b9ae47b4554f8b5b1fa0fadcc1a28842cd07b7..ef1f5fdeb1b8f248d281b5fb989f714b5b75649e

Following is the flow chart of require>
WM_815_Google_Slides

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
@@ -84,8 +84,6 @@ public class RequireOperatorFactory
 
             Config overrideParams = config.getNestedOrGetEmpty("params");
 
-            Optional<StoredSessionAttempt> attempt = Optional.absent();
-            Optional<SessionAttemptConflictException> sessionAttemptConflictException = Optional.absent();
             Optional<ProjectIdentifier> projectIdentifier = Optional.absent();
             /**
              *  First of all, try to start attempt by startSession()
@@ -108,58 +106,51 @@ public class RequireOperatorFactory
             try {
                 projectIdentifier = Optional.of(makeProjectIdentifier());
 
-                attempt = Optional.of(callback.startSession(
+                callback.startSession(
                         context,
                         request.getSiteId(),
                         projectIdentifier.get(),
                         workflowName,
                         instant,
                         retryAttemptName,
-                        overrideParams));
+                        overrideParams);
+                throw nextPolling(request.getLastStateParams().deepCopy().set("require_kicked", true));
             }
             catch (SessionAttemptConflictException ex) {
-                sessionAttemptConflictException = Optional.of(ex);
+                return processAttempt(ex.getConflictedSession(), lastStateParams, rerunOn, ignoreFailure);
             }
             catch (ResourceNotFoundException ex) {
                 throw new TaskExecutionException(String.format(ENGLISH, "Dependent workflow does not exist. %s, workflowName:%s",
-                        projectIdentifier.transform((p)->p.toString()).or(""), workflowName ));
+                        projectIdentifier.transform(ProjectIdentifier::toString).or(""), workflowName));
             }
             catch (ResourceLimitExceededException ex) {
                 throw new TaskExecutionException(ex);
             }
+        }
 
-            if (sessionAttemptConflictException.isPresent()) {
-                StoredSessionAttempt conflictedAttempt = sessionAttemptConflictException.get().getConflictedSession();
-                if (conflictedAttempt.getStateFlags().isDone()) {
-                    // A flag to distinguish whether the attempt is kicked by require> or previous attempt.
-                    boolean requireKicked = lastStateParams.get("require_kicked", boolean.class, false);
-                    if (!requireKicked &&  (
-                            rerunOn == OptionRerunOn.ALL ||
-                            rerunOn == OptionRerunOn.FAILED && !conflictedAttempt.getStateFlags().isSuccess())) {
+        private TaskResult processAttempt(StoredSessionAttempt attempt, Config lastStateParams, OptionRerunOn rerunOn, boolean ignoreFailure)
+        {
+            if (attempt.getStateFlags().isDone()) {
+                // A flag to distinguish whether the attempt is kicked by require> or previous attempt.
+                boolean requireKicked = lastStateParams.get("require_kicked", boolean.class, false);
+                if (!requireKicked &&  (
+                        rerunOn == OptionRerunOn.ALL ||
+                                rerunOn == OptionRerunOn.FAILED && !attempt.getStateFlags().isSuccess())) {
 
-                        //To force run, set flag gen_retry_attempt_name and do polling
-                        throw nextPolling(lastStateParams.deepCopy().set("rerun_on_retry_attempt_name", UUID.randomUUID().toString()));
-                    }
-                    if (!ignoreFailure && !conflictedAttempt.getStateFlags().isSuccess()) {
-                        // ignore_failure is false and the attempt is in error state. Make this operator failed.
-                        throw new TaskExecutionException(String.format(ENGLISH,
+                    //To force run, set flag gen_retry_attempt_name and do polling
+                    throw nextPolling(lastStateParams.deepCopy().set("rerun_on_retry_attempt_name", UUID.randomUUID().toString()));
+                }
+                if (!ignoreFailure && !attempt.getStateFlags().isSuccess()) {
+                    // ignore_failure is false and the attempt is in error state. Make this operator failed.
+                    throw new TaskExecutionException(String.format(ENGLISH,
                             "Dependent workflow failed. Session id: %d, attempt id: %d",
-                            conflictedAttempt.getSessionId(), conflictedAttempt.getId()));
-                    }
-                    return TaskResult.empty(cf);
+                            attempt.getSessionId(), attempt.getId()));
                 }
-                else {
-                    // Wait for finish running attempt
-                    throw nextPolling(request.getLastStateParams().deepCopy());
-                }
-            }
-            else if (attempt.isPresent()) { // startSession succeeded and created new attempt
-                throw nextPolling(request.getLastStateParams().deepCopy().set("require_kicked", true));
+                return TaskResult.empty(cf);
             }
             else {
-                throw new RuntimeException(String.format(ENGLISH,
-                        "Unexpected condition happened in require> operator.  %s, workflowName:%s",
-                        projectIdentifier.transform((p)->p.toString()).or(""), workflowName));
+                // Wait for finish running attempt
+                throw nextPolling(request.getLastStateParams().deepCopy());
             }
         }
 
--- a/digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
@@ -6,6 +6,7 @@ import io.digdag.client.config.Config;
 import io.digdag.client.config.ConfigElement;
 import io.digdag.client.config.ConfigException;
 import io.digdag.client.config.ConfigFactory;
+import io.digdag.core.agent.TaskCallbackApi.ProjectIdentifier;
 import io.digdag.core.repository.ResourceLimitExceededException;
 import io.digdag.core.repository.ResourceNotFoundException;
 import io.digdag.core.session.StoredSessionAttempt;
@@ -13,13 +14,11 @@ import io.digdag.spi.Operator;
 import io.digdag.spi.OperatorContext;
 import io.digdag.spi.OperatorFactory;
 import io.digdag.spi.TaskExecutionException;
-import io.digdag.spi.TaskRequest;
 import io.digdag.spi.TaskResult;
 import io.digdag.util.BaseOperator;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.nio.file.Path;
 import java.time.Instant;
 
 import static java.util.Locale.ENGLISH;
@@ -66,16 +65,17 @@ public class RequireOperatorFactory
         {
             Config config = request.getConfig();
             String workflowName = config.get("_command", String.class);
-            int projectId = config.get("project_id", int.class);
             Instant instant = config.get("session_time", Instant.class);
             boolean ignoreFailure = config.get("ignore_failure", boolean.class, false);
             Optional<String> retryAttemptName = config.getOptional("retry_attempt_name", String.class);
             Config overrideParams = config.getNestedOrGetEmpty("params");
             try {
+                ProjectIdentifier projectIdentifier = makeProjectIdentifier();
+
                 StoredSessionAttempt attempt = callback.startSession(
                         context,
                         request.getSiteId(),
-                        projectId,
+                        projectIdentifier,
                         workflowName,
                         instant,
                         retryAttemptName,
@@ -102,6 +102,35 @@ public class RequireOperatorFactory
             catch (ResourceLimitExceededException ex) {
                 throw new TaskExecutionException(ex);
             }
+        }
+
+
+        /**
+         * Make ProjectIdentifier from parameters.
+         * @return
+         */
+        private ProjectIdentifier makeProjectIdentifier()
+        {
+            Config config = request.getConfig();
+            Config localConfig = request.getLocalConfig();
+            int projectId = config.get("project_id", int.class);
+            Optional<Integer> projectIdParam = localConfig.getOptional("project_id", Integer.class);
+            Optional<String> projectNameParam = localConfig.getOptional("project_name", String.class);
+
+            if (projectIdParam.isPresent() && projectNameParam.isPresent()) {
+                throw new ConfigException("Both project_id and project_name can't be set");
+            }
+            ProjectIdentifier projectIdentifier;
+            if (projectNameParam.isPresent()) {
+                projectIdentifier = ProjectIdentifier.ofName(projectNameParam.get());
+            }
+            else if (projectIdParam.isPresent()) {
+                projectIdentifier = ProjectIdentifier.ofId(projectIdParam.get());
+            }
+            else {
+                projectIdentifier = ProjectIdentifier.ofId(projectId);
+            }
+            return projectIdentifier;
         }
     }
 }
--- a/digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
@@ -10,6 +10,7 @@ import io.digdag.core.agent.TaskCallbackApi.ProjectIdentifier;
 import io.digdag.core.repository.ResourceLimitExceededException;
 import io.digdag.core.repository.ResourceNotFoundException;
 import io.digdag.core.session.StoredSessionAttempt;
+import io.digdag.core.workflow.SessionAttemptConflictException;
 import io.digdag.spi.Operator;
 import io.digdag.spi.OperatorContext;
 import io.digdag.spi.OperatorFactory;
@@ -20,6 +21,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.time.Instant;
+import java.util.UUID;
 
 import static java.util.Locale.ENGLISH;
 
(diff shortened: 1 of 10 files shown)


digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -65,42 +68,157 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
                         retryAttemptName,
-                        overrideParams);
+                        overrideParams));
+            }
+            catch (SessionAttemptConflictException ex) {
+                sessionAttemptConflictException = Optional.of(ex);
+            }
+            catch (ResourceNotFoundException ex) {
+                throw new TaskExecutionException(String.format(ENGLISH, "Dependent workflow does not exist. %s, workflowName:%s",
+                        projectIdentifier.transform((p)->p.toString()).or(""), workflowName ));
```
komamitsu, line 125: There is an unexpected space between `workflowName` and `)`.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -64,46 +68,109 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
+             *  First of all, try to start attempt by startSession()
+             *  If no attempt exists (no conflict), it return new StoredSessionAttempt.
+             *    - set state param "require_kicked" to true.
+             *    - task is retried to wait for done by nextPolling.
+             *  If something errors happen, it will throw following exceptions.
+             *    - ResourceNotFoundException ... workflow ,project name(or id) are wrong. -> processed as deterministic error
+             *    - ResourceLimitExceededException ... this exception should be deterministic error
+             *    - SessionAttemptConflictException ... this is not error.
+             *      - If the conflict attempt is still running, wait for until done by nextPolling.
+             *      - If done, check the state param "require_kicked" and whether the attempt is kicked by this require> or not.
```
komamitsu, line 100: What about the case the conflict attempt failed like `If failed, xxx`?

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
komamitsu: [minor] `(p)->p.toString()` can be replaced with `ProjectIdentifier::toString`

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -64,46 +68,109 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
+            }
+            catch (ResourceLimitExceededException ex) {
+                throw new TaskExecutionException(ex);
+            }
 
-                boolean isDone = attempt.getStateFlags().isDone();
-                if (isDone) {
-                    if (!ignoreFailure && !attempt.getStateFlags().isSuccess()) {
+            if (sessionAttemptConflictException.isPresent()) {
+                StoredSessionAttempt conflictedAttempt = sessionAttemptConflictException.get().getConflictedSession();
```
komamitsu, line 132: How about moving this block to `catch (SessionAttemptConflictException ex) { ... }` block? Maybe we can remove the Optional?

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -64,46 +68,109 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
-                    // TODO use exponential-backoff to calculate retry interval
-                    throw TaskExecutionException.ofNextPolling(1, ConfigElement.copyOf(request.getLastStateParams()));
+                    // Wait for finish running attempt
+                    throw nextPolling(request.getLastStateParams().deepCopy());
                 }
             }
-            catch (ResourceNotFoundException ex) {
-                throw new ConfigException(ex);
+            else if (attempt.isPresent()) { // startSession succeeded and created new attempt
+                throw nextPolling(request.getLastStateParams().deepCopy().set("require_kicked", true));
```
komamitsu, line 157: How about moving this line to right after calling `callback.startSession()`  in the above `try { ... }` block? Maybe we can remove the Optional?

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Of course, we can do.  My intention of this way is that firstly get result of startSession() including exception (especially SessionAttemptConflictException is not error), then the logic goes ahead based on the result.
if we move the logic in `catch (SessionAttemptConflictException ex) { ... }`, catch {} become large.  
In this case it is better to define another method and call it.
What is your thought?

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: My intention is same as above. But this code is small. 
So I agree with move to `try{..}`.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Do you mean, if the attempt which is kicked by require> failed,  what happen?
In this case the task will fail as usual.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Thanks! I will fix them.

yoyama: Fixed it. Thanks!

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Based on discussion in person, I fixed it.
As result, the code became simple. Thank you for your advice! 🙂

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Fixed it. Thanks!

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -65,43 +68,149 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
+
             String workflowName = config.get("_command", String.class);
-            int projectId = config.get("project_id", int.class);
             Instant instant = config.get("session_time", Instant.class);
             boolean ignoreFailure = config.get("ignore_failure", boolean.class, false);
-            Optional<String> retryAttemptName = config.getOptional("retry_attempt_name", String.class);
+            OptionRerunOn rerunOn = OptionRerunOn.of(config.get("rerun_on", String.class, "none"));
+
+            //retry_attempt_name is set from local config only.
+            Optional<String> retryAttemptName = localConfig.getOptional("retry_attempt_name", String.class);
```
komamitsu, line 80: I'm wondering if this scope reduction of `retry_attempt_name` can affect existing workflows...

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -65,43 +68,149 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
+        }
+    }
+
+    enum OptionRerunOn
+    {
+        NONE("none"),
+        FAILED( "failed"),
+        ALL ("all");
+
+        private final String name;
```
komamitsu, line 201: This isn't needed?

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -65,43 +68,149 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
             }
         }
+
+        private TaskResult processAttempt(StoredSessionAttempt attempt, Config lastStateParams, OptionRerunOn rerunOn, boolean ignoreFailure)
+        {
+            if (attempt.getStateFlags().isDone()) {
+                // A flag to distinguish whether the attempt is kicked by require> or previous attempt.
+                boolean requireKicked = lastStateParams.get("require_kicked", boolean.class, false);
+                if (!requireKicked &&  (
+                        rerunOn == OptionRerunOn.ALL ||
```
komamitsu, line 137: When OptionRerunOn.ALL is set, the workflow should be executed regardless it was executed via `require` operator or not, right? If so, `rerunOn == OptionRerunOn.ALL` should not depend on `requireKicked`?

komamitsu: Ah, `require_kicked` is marked on the current request not a previous attempt...

komamitsu: Ah, got it. I had a wrong idea. It's for `Check attempt kicker` on the diagram.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -65,43 +68,149 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
+            }
+        }
+
+        private TaskExecutionException nextPolling(Config stateParams)
+        {
+            int iteration = stateParams.get("retry", int.class, 0);
+            stateParams.set("retry", iteration+1);
+            int interval = (int) Math.min(1 * Math.pow(2, iteration), MAX_TASK_RETRY_INTERVAL);
+
+            return TaskExecutionException.ofNextPolling( interval, ConfigElement.copyOf(stateParams));
```
komamitsu, line 163: A space on the left of `interval`

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -65,43 +68,149 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
+            else {
+                // Wait for finish running attempt
+                throw nextPolling(request.getLastStateParams().deepCopy());
+            }
+        }
+
+        private TaskExecutionException nextPolling(Config stateParams)
+        {
+            int iteration = stateParams.get("retry", int.class, 0);
+            stateParams.set("retry", iteration+1);
```
komamitsu, line 160: Spaces on the both sides of `+` would be better?

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
```diff
@@ -65,43 +68,149 @@ private RequireOperator(OperatorContext context, TaskCallbackApi callback)
...
+                    // ignore_failure is false and the attempt is in error state. Make this operator failed.
+                    throw new TaskExecutionException(String.format(ENGLISH,
+                            "Dependent workflow failed. Session id: %d, attempt id: %d",
+                            attempt.getSessionId(), attempt.getId()));
+                }
+                return TaskResult.empty(cf);
+            }
+            else {
+                // Wait for finish running attempt
+                throw nextPolling(request.getLastStateParams().deepCopy());
```
komamitsu, line 153: Maybe duplicated deep copy?

digdag-docs/src/operators/require.md
```diff
@@ -71,3 +88,8 @@
   ignore_failure: true
   ```
 
+  require> evaluates *ignore_failure* at last of its process. If rerun_on is set and require> run new attempt, the result of new attempt is checked.
+
+## Notes
+- require> has been changed to ignore inherited *retry_attempt_name* parameter. 
+  `digdag retry ` command generates unique retry_attempt_name to run, but it is not passed to require>.
```
komamitsu, line 95: Unnecessary space on the right of `digdag retry`

digdag-docs/src/operators/require.md
```diff
@@ -71,3 +88,8 @@
   ignore_failure: true
   ```
 
+  require> evaluates *ignore_failure* at last of its process. If rerun_on is set and require> run new attempt, the result of new attempt is checked.
+
+## Notes
+- require> has been changed to ignore inherited *retry_attempt_name* parameter.
```
komamitsu, line 94: Adding `since v0.10` or something may be helpful for users, I think.

yoyama: Good point. But I am planning back port to 0.9.xx.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Ah, correct. Currently not used. Will remove it. Thanks!

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Yes. There is possibility as discussion in person.
Let me discuss again in a meeting.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: May be not duplicated, but I should use lastStateParam argument in my understand.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
komamitsu: Just noticed you've already mentioned it in the document :)

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Thanks! Fixed it.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: Thanks! Fixed it.

digdag-core/src/main/java/io/digdag/core/agent/RequireOperatorFactory.java
yoyama: I updated it.

digdag-docs/src/operators/require.md
yoyama: Fixed it. Thanks!

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +203,159 @@ public void testIgnoreProjectIdParam()
...
+                "-e", server.endpoint());
+        assertThat(pushParentStatus.errUtf8(), pushParentStatus.code(), is(0));
+
+        CommandStatus startStatus = main("start",
+                "-c", config.toString(),
+                "-e", server.endpoint(),
+                "parent_another", "parent_another_project",
+                "--session", "now"
+        );
+        assertThat(startStatus.errUtf8(), startStatus.code(), is(0));
```
komamitsu, line 250: Just posted a comment https://github.com/treasure-data/digdag/pull/1405/files#r435112975

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +203,159 @@ public void testIgnoreProjectIdParam()
...
+            RestSessionAttemptCollection childAttempts = testRerunOnParam(sessionTime, "failed", false);
+            assertThat("Number of child's attempt must be one. (== require> kick)", childAttempts.getAttempts().size(), is(1));
+        }
+    }
+
+    private RestSessionAttemptCollection testRerunOnParam(String sessionTime, String rerunOn, boolean previousChildRunFail) throws InterruptedException
+    {
+        //Start a child
+        String childFailParam = previousChildRunFail ? "yes" : "no";
+        CommandStatus childStatus = startAndWait(true,"require", "child", "--session", sessionTime,
```
komamitsu, line 310: [minor] Space

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +203,159 @@ public void testIgnoreProjectIdParam()
...
+                "-p", "child_fail=" + childFailParam);
+        assertThat(isAttemptSuccess(childStatus), not(previousChildRunFail));
+
+        // Start a parent with same session time.
+        CommandStatus rerunOnNoneStatus = startAndWait(false,"require", "parent", "--session", sessionTime,
```
komamitsu, line 315: Same as above

digdag-tests/src/test/java/acceptance/RequireIT.java
komamitsu, line 310: I'm wondering why this method call always passes `ignoreAttemptFailure=true` (sorry, a bit too difficult for me 😅.)

Could I ask you the followings?
1. Add a comment for each case about why target attempt is expected to fail
2. If possible, make this method call pass `attemptShouldFail` or something deterministic instead of ` ignoreAttemptFailure`

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +203,159 @@ public void testIgnoreProjectIdParam()
...
+                success = true;
+                break;
+            }
+            else if (attemptsStatus.outUtf8().contains("status: error")) {
+                break;
+            }
+            Thread.sleep(1000);
+        }
+        if (!ignoreAttemptFailure) {
+            assertThat(success, is(true));
```
komamitsu, line 353: Oh, the same check is done above like `assertThat(isAttemptSuccess(childStatus), not(previousChildRunFail))`?
So we can remove this to make things simpler?

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +203,159 @@ public void testIgnoreProjectIdParam()
...
+            throws Exception
+    {
+        // Create new project
+        CommandStatus initStatus = main("init",
+                "-c", config.toString(),
+                projectDir.toString());
+        assertThat(initStatus.errUtf8(), initStatus.code(), is(0));
+
+        Path childOutFile = projectDir.resolve("child.out").toAbsolutePath().normalize();
+        prepareForChildWF(childOutFile);
```
komamitsu, line 264: Is this line needed?

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +203,159 @@ public void testIgnoreProjectIdParam()
...
+                "require",
+                "-c", config.toString(),
+                "-e", server.endpoint(),
+                "-r", "4711");
+        assertThat(pushStatus.errUtf8(), pushStatus.code(), is(0));
+
+        // test for rerun_on: none
+        {
+            String sessionTime = "2020-05-28 12:34:01";
+            RestSessionAttemptCollection childAttempts = testRerunOnParam(sessionTime, "none", false);
```
komamitsu, line 280: How about adding a test case of the combination of `none` and true?

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +203,159 @@ public void testIgnoreProjectIdParam()
...
+            assertThat("Number of child's attempt must be one. (== require> skip the call)", childAttempts.getAttempts().size(), is(1));
+        }
+
+        // test for rerun_on: all. require> kick child always.
+        {
+            String sessionTime = "2020-05-28 12:34:11";
+            RestSessionAttemptCollection childAttempts = testRerunOnParam(sessionTime, "all", false);
```
komamitsu, line 287: How about adding a test case of the combination of `all` and true?

digdag-tests/src/test/java/acceptance/RequireIT.java
komamitsu: After having read all codes in `testRerunOnParam()` I understand all things. So probably you can ignore the above comment from me.

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: Thanks. I will remove 263 and 264.

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: Thanks for your pointing out. I will remove them and argument ignoreAttemptFailure.

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: I will add some additional comments.

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: I fixed it in #1405 and if it become OK, rebase to this PR.

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: I added it. Thanks!

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: Thanks, I added it. Now all conditions are covered.

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: Fixed it, thanks.

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: Fixed it and refactoring. Thanks for your pointing out.

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: Done. Thanks!

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: Fixed it. (Removed the param because of refactoring)

digdag-tests/src/test/java/acceptance/RequireIT.java
yoyama: Fixed it. (Removed the param because of refactoring)

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -185,6 +203,176 @@ public void testIgnoreProjectIdParam()
...
+    /**
+     * Test for rerun_on
+     *
+     * @param sessionTime
+     * @param rerunOn              parameter for rerun_on:  "none" or "fail" or "all"
+     * @param previousChildRunFail if true, first child workflow failed.
+     * @return
+     * @throws InterruptedException
+     */
+    private RestSessionAttemptCollection testRerunOnParam(String sessionTime, String rerunOn, boolean previousChildRunFail, boolean expectParentSucceed) throws InterruptedException
```
komamitsu, line 328: 💯

digdag-tests/src/test/resources/acceptance/require/parent_rerun_on.dig
```diff
@@ -0,0 +1,6 @@
+
```
komamitsu, line 1: Empty line?

yoyama:
@komamitsu  Thank you so much for your long reviews.
It is very helpful. After fix of #1405 is approved, I will rebase it then merge this.

PR #1418, opened 2020-06-18
updated scheduling doc to inform using week number can be used for we…

kohki1234:
I added the description below to indicate a week number can be used as well instead of DDD,HH:MM:SS.

For weekly scheduling, you can specify a week number like this "2,10:32:32". 2 is Tuesday in this example.


hiroyuki-sato:
Hello, @kohki1234 

Thank you for creating the PR.

This note describes the usage of session time. I think It is better to write in another place about the week of the day description.
What do you think to change `weekly>: The,10:32:32` instead of `weekly>: `2,10:32:32`?
This day of the week style easier than number style.

And also this page doesn't use a number of the week style except for this part.

kohki1234:
@hiroyuki-sato 
Thank you for reviewing and giving feedback! 
I initially thought that adding an explanation that week of day number can be used, but I totally agree that writing like `weekly>: Tue,10:32:32 ` is more readable. 

Will create another PR

PR #1419, opened 2020-06-18
Fix resume mode to work correctly when group retry happened

szyn:
When resuming a workflow where a group retry has occurred, it will be restarted from the first task in the group. Because if a group retry has occurred, it is difficult to determine which task to resume, it might be a good idea to retry the entire group. 

e.g., The user wants to retry failed the following workflow that named sample, and it will be resumed from the task `+sample+group1+a`.

sample.dig
```yaml
+group1:
  _retry: 2
  +a:
    echo>: a
  +b:
    echo>: b
  +c:
    fail>: error! ## This task will fail
```

Related PR: https://github.com/treasure-data/digdag/pull/1163


digdag-tests/src/test/java/acceptance/GroupRetryIT.java
```diff
@@ -0,0 +1,208 @@
...
+        // Wait for the attempt to fail
+        assertThat(joinAttempt(client, originalAttemptId).getSuccess(), is(false));
+
+        assertOutputExists(originalAttemptId + "1-1", true);
+        assertOutputExists(originalAttemptId + "1-2a", true);
+        assertOutputExists(originalAttemptId + "1-2b", true);
+        assertOutputExists(originalAttemptId + "1-2c", false);
+
+        // Push a new revision
+        pushRevision("acceptance/group_retry/retry-2.dig", "retry");
```
yoyama, line 80: Project name and workflow name and command are all same name **"retry"** ?
It ma be a bit confused for me.

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
```diff
@@ -302,7 +302,16 @@ public Response startAttempt(RestSessionAttemptRequest request)
                 .getSessionStore(getSiteId())
                 .getTasksOfAttempt(attemptId);
 
+        List<Long> groupRetryErrorTaskIds = tasks.stream()
+                .filter(task -> task.getState() == TaskStateCode.GROUP_ERROR && task.getStateParams().has("retry_count"))
```
yoyama, line 306: May I ask you the purpose to check "retry_count" ?

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
```diff
@@ -344,7 +353,11 @@ private ArchivedTask matchTaskPattern(String pattern, List<ArchivedTask> tasks)
                     tasks
                             .stream()
                             .collect(
-                                    Collectors.toMap(t -> t.getFullName(), t -> t)
+                                    // Avoid 500 error due to duplicate key ArchivedTask
+                                    // if a group retry happened, IllegalStateException will be thrown
+                                    // because there are multiple tasks of the same name.
+                                    // Take the first one to make sure it is appropriately resumed.
+                                    Collectors.toMap(t -> t.getFullName(), t -> t, (a, b) -> a)
```
yoyama, line 360: May I ask you the reason you choose first one ?

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
szyn: Yes, it's to determine if a group retry has occurred or not.
Whenever a group retry occurs, there is a `retry_count` in stateParams.

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
szyn: There is no specific reason. Either is fine because of the same name.

digdag-tests/src/test/java/acceptance/GroupRetryIT.java
szyn: I agree with you.. sorry for confusing, let me update project and workflow name.

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
yoyama: Got it. Thanks for your explanation.

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
yoyama: Ah you need only the name. Got it. Thanks!

yoyama:
Basically LGTM.
But if possible, could you add more complicated test cases ?
For example.
- nested group task
- combination of call>

digdag-tests/src/test/java/acceptance/GroupRetryIT.java
```diff
@@ -0,0 +1,583 @@
...
+
+        // Wait for the attempt to fail again
+        assertThat(joinAttempt(client, retry1).getSuccess(), is(false));
+
+        assertOutputExists(retry1 + "1-1", true);
+        assertOutputExists(retry1 + "0-1-2a", true);
+        assertOutputExists(retry1 + "0-1-2b", true);
+        assertOutputExists(retry1 + "0-1-2c", false);
+
+        // Retry with the latest fixed revision & resume failed
```
yoyama, line 227: resume failed ?

digdag-tests/src/test/java/acceptance/GroupRetryIT.java
```diff
@@ -0,0 +1,583 @@
...
+
+        // Wait for the attempt to fail again
+        assertThat(joinAttempt(client, retry1).getSuccess(), is(false));
+
+        assertOutputExists(retry1 + "1-1", true);
+        assertOutputExists(retry1 + "1-2a", true);
+        assertOutputExists(retry1 + "1-2b", true);
+        assertOutputExists(retry1 + "1-2c", false);
+
+        // Retry with the latest fixed revision & resume failed
```
yoyama, line 353: ditto. it seems you may expect to succeed.

digdag-tests/src/test/java/acceptance/GroupRetryIT.java
```diff
@@ -0,0 +1,583 @@
...
+                    "--resume",
+                    String.valueOf(originalAttemptId));
+            assertThat(retryStatus.errUtf8(), retryStatus.code(), is(0));
+            retry2 = getAttemptId(retryStatus);
+        }
+
+        // Wait for the attempt to success
+        assertThat(joinAttempt(client, retry2).getSuccess(), is(true));
+
+        assertOutputExists(retry2 + "2-1", true);   // no-skipped, +group1 has _retry
```
yoyama, line 369: I am not sure that this is really expected behavior or not.

digdag-tests/src/test/java/acceptance/GroupRetryIT.java
```diff
@@ -0,0 +1,583 @@
...
+        // Wait for the attempt to fail again
+        assertThat(joinAttempt(client, retry1).getSuccess(), is(false));
+
+        assertOutputExists(retry1 + "1-1", true);
+        assertOutputExists(retry1 + "1-2a", true);
+        assertOutputExists(retry1 + "1-2b", true);
+        assertOutputExists(retry1 + "1-2c", false);
+
+
+        // Retry with the latest fixed revision & resume failed
```
yoyama, line 478: ditto ?

digdag-tests/src/test/resources/acceptance/group_retry/retry-with-call-3.dig
```diff
@@ -0,0 +1,21 @@
+_export:
+  outdir: ${outdir}
+
++step1:
+  _export:
+    val: 2-1
```
yoyama, line 6: 3-1 is better ?

digdag-tests/src/test/java/acceptance/GroupRetryIT.java
szyn: resume failed actually means retry failed tasks. My understanding that it comes from here: https://github.com/treasure-data/digdag/blob/1f277a06b9ada4c9bbf1c6884f9e3ff20ca1af52/digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java#L220-L224

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
```diff
@@ -302,7 +302,19 @@ public Response startAttempt(RestSessionAttemptRequest request)
                 .getSessionStore(getSiteId())
                 .getTasksOfAttempt(attemptId);
 
+        // Collect tasksIds in which a group retry occurred.
+        // Need to filter the status of the task with "group_error" and "retry_count" in stateParams.
```
komamitsu, line 306: >         // Need to filter the status of the task with "group_error" and "retry_count" in stateParams.

This might be a silly question... but what does it mean if the group error task has no `retry_count`?

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
```diff
@@ -344,7 +356,11 @@ private ArchivedTask matchTaskPattern(String pattern, List<ArchivedTask> tasks)
                     tasks
                             .stream()
                             .collect(
-                                    Collectors.toMap(t -> t.getFullName(), t -> t)
+                                    // Avoid 500 error due to duplicate key ArchivedTask
```
komamitsu, line 360: `duplicated key of ArchivedTask` is a bit better?

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
szyn: It means that group error has occurred, but no group retry has occurred. I added the same description to the codebase :)

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
komamitsu: When are duplicated key tasks generated?

digdag-tests/src/test/resources/acceptance/group_retry/retry-1.dig
```diff
@@ -0,0 +1,10 @@
++step1:
+  sh>: touch ${outdir}/${attempt_id}1-1.out
++step2:
+  _retry: 2
```
komamitsu, line 4: All the group tasks in this test have ` _retry: 2`, but with https://github.com/treasure-data/digdag/pull/1419/files#diff-55f697f2343f66f95a3fd8c57c0710a6R309 the existence of ` _retry: 2` switches `retry --resume`'s behaviour. How about adding tests that don't contain `_retry: 2`?

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
komamitsu: Using `retry_count` as the switch of group retry seems a bit weird to me (Sorry if we've already discussed it. I may have had a wrong idea) since I thought `retry_count` is a just retry count for unexpected failures not like affecting the graph of tasks.

digdag-tests/src/test/java/acceptance/GroupRetryIT.java
komamitsu: So `resume failed tasks` or something is clearer?

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
```diff
@@ -302,7 +302,19 @@ public Response startAttempt(RestSessionAttemptRequest request)
...
+
         List<Long> successTasks = tasks.stream()
+                // If a group error has occurred,
+                // exclude the group's child/dynamically generated tasks from successTasks
+                // even if they are in SUCCESS state
+                .filter(task -> task.getParentId().isPresent() && !groupRetryErrorTaskIds.contains(task.getParentId().get()))
+                .filter(task -> !task.getFullName().contains("^sub"))
                 .filter(task -> task.getState() == TaskStateCode.SUCCESS)
                 .map(task -> {
                     if (!task.getParentId().isPresent()) {
```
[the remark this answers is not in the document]
szyn: The above filter will only exclude child tasks with a group error at the parent, I suspect this validation cannot be removed.

digdag-docs/src/command_reference.rst
```diff
@@ -512,6 +512,8 @@ Examples:
 
 :command:`--resume`
   Retry only failed tasks. Successfully finished tasks are skipped.
+  Note that if a retry is set up at the group level (not task level), it will resume from the beginning of the group, not only failed tasks.
+  The same applies to the nested group.
```
yoyama, line 516: A bit uncertain.
I am not sure, but the following may be better.
> If __retry_ is set to each nested groups, it will resume from the top of the group.

Issue #1423, opened 2020-06-23
bq> does not support to query BigQuery table with Google Sheet external table.


wapa5pow:
Google BigQuery table can be created with Google Sheet as follows.

![image](https://user-images.githubusercontent.com/5561944/85405391-50e58980-b59b-11ea-972a-4607ce604828.png)

When querying this kind of table, bq> operator needs drive read scope: "https://www.googleapis.com/auth/drive.readonly".
(https://github.com/googleapis/google-api-java-client-services/blob/master/clients/google-api-services-sheets/v4/1.30.1/com/google/api/services/sheets/v4/SheetsScopes.java#L33)

However, digdag does not request the scope in the source code.
( https://github.com/treasure-data/digdag/blob/master/digdag-standards/src/main/java/io/digdag/standards/operator/gcp/BqClient.java#L40 )

If bq> operator query to the table with sheet, the following error occurs.

```
{
  "location" : "xxx",
  "message" : "Error while reading table: test.test, error message: Failed to read the spreadsheet. Errors: No OAuth token with Google Drive scope was found.",
  "reason" : "invalid"
}
```

To fix this, import `import com.google.api.services.sheets.v4.SheetsScopes;`, then fix scope with the following code.

```
            Set<String> scopes = new HashSet<String>();
            scopes.addAll(BigqueryScopes.all());
            scopes.add(SheetsScopes.DRIVE_READONLY);
            credential = credential.createScoped(scopes);
```

PR #1424, opened 2020-06-23
fix bq operator for table with google sheet

wapa5pow:
https://github.com/treasure-data/digdag/issues/1423

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/gcp/BqClient.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/gcp/BqClient.java
@@ -12,6 +12,7 @@ import com.google.api.services.bigquery.model.Job;
 import com.google.api.services.bigquery.model.Table;
 import com.google.api.services.bigquery.model.TableList;
 import com.google.api.services.bigquery.model.TableReference;
+import com.google.api.services.sheets.v4.SheetsScopes;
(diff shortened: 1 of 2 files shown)


yoyama:
@wapa5pow Thank you for your contribution and very sorry for late reply.
Currently we are working for major version upgrade. After then check this PR.

yoyama:
`master` branch has been changed significantly. Could you change the PR based on current master?

wapa5pow:
ok, done

digdag-standards/src/main/java/io/digdag/standards/operator/gcp/BqClient.java
```diff
@@ -37,7 +40,10 @@
     protected Bigquery client(GoogleCredential credential, HttpTransport transport, JsonFactory jsonFactory)
     {
         if (credential.createScopedRequired()) {
-            credential = credential.createScoped(BigqueryScopes.all());
+            Set<String> scopes = new HashSet<String>();
+            scopes.addAll(BigqueryScopes.all());
+            scopes.add(SheetsScopes.DRIVE_READONLY);
```
yoyama, line 45: I am not sure google credential but I have a concern about adding scope affect to existence users or not.
It means that users need to change their credential in secrets or change permission on account even though they don't need to access to Google spreadsheet.
Could you explain about it ?

PR #1425, opened 2020-06-25
Mask debug message in "evaluated config"

yoyama:
Mask "evaluated config" debug message except predefined keys.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -1,5 +1,6 @@
 package io.digdag.core.agent;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.inject.Inject;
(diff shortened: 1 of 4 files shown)


digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -388,4 +387,30 @@ public static void collectExceptionMessage(StringBuilder sb, Throwable ex, Strin
...
+    };
+
+    @VisibleForTesting
+    public Config filterConfigForLogging(Config src)
+    {
+        Config dst = cf.create();
+        for (String key : CONFIG_KEYS_FOR_LOGGING) {
+            Optional<Object> v = src.getOptional(key, Object.class);
+            if (v.isPresent()) {
+                dst.set(key, v.get());
```
komamitsu, line 411: Config.set() just shares the reference to the values without copying data and won't consume memory a lot. Looks good.

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -388,4 +387,30 @@ public static void collectExceptionMessage(StringBuilder sb, Throwable ex, Strin
             collectExceptionMessage(sb, t, used);
         }
     }
+
+    private static final String[] CONFIG_KEYS_FOR_LOGGING = {
+            "project_id",
+            "session_id",
+            "session_time",
```
szyn, line 394: How about adding `session_unixtime ` as well? I think it's good to have it.

PR #1430, opened 2020-07-03
Configurable Limits.

yoyama:
To be configurable, introduce two config parameters as follows

executor.task_max_run
executor.attempt_max_run


digdag-core/src/main/java/io/digdag/core/Limits.java
```diff
@@ -8,15 +12,21 @@
...
+    public Limits(Config systemConfig)
+    {
+        this.systemConfig = systemConfig;
+    }
 
-    public static long maxWorkflowTasks()
+    public long maxWorkflowTasks()
     {
-        return MAX_WORKFLOW_TASKS;
+        return systemConfig.get("executor.task_max_run", long.class, MAX_WORKFLOW_TASKS);
```
komamitsu, line 25: How about memoization of this value in terms of performance?

digdag-core/src/test/java/io/digdag/core/workflow/ForEachOperatorFactory.java
```diff
@@ -41,17 +45,19 @@ public String getType()
...
     }
 
     static class ForEachOperator
             implements Operator
     {
         private final TaskRequest request;
+        private final Limits limits;
 
-        public ForEachOperator(OperatorContext context)
+        public ForEachOperator(OperatorContext context, Limits limits)
```
komamitsu, line 57: Is it possible to extend OperatorContext to have `limits` field instead?

digdag-docs/src/command_reference.rst
```diff
@@ -386,7 +386,9 @@ In the config file, following parameters are available
 * log-server.local.split_size (long. default: 0  (not splitted))
 * digdag.secret-encryption-key = (base64 encoded 128-bit AES encryption key)
 * executor.task_ttl (string. default: 1d. A task is killed if it is running longer than this period.)
+* executor.task_max_run (integer. default: 1000. Max number of tasks in workflow.)
 * executor.attempt_ttl (string. default: 7d. An attempt is killed if it is running longer than this period.)
+* executor.attempt_max_run (integer. default: 100. Max number of running attempts per each site_id.)
```
komamitsu, line 391: How about adding term `at once` or `in concurrently` in the description?

digdag-core/src/test/java/io/digdag/core/workflow/ForEachOperatorFactory.java
yoyama: Good point. But OperatorContext is defined in digdag-spi and Limits is defined in digdag-core. So we cannot directly reference Limits from OperatorContext.
We may need to find good way to fix it.

digdag-core/src/main/java/io/digdag/core/Limits.java
yoyama: Good point. I am not sure how much it contribute to performance but valuable.

digdag-docs/src/command_reference.rst
yoyama: Thanks, will fix it.

digdag-core/src/main/java/io/digdag/core/Limits.java
yoyama: Done.

digdag-docs/src/command_reference.rst
yoyama: Done.

digdag-core/src/test/java/io/digdag/core/workflow/ForEachOperatorFactory.java
yoyama: I added a method `getMaxWorkflowTasks()` in `OperatorContext` and set from Limits instance.

digdag-core/src/main/java/io/digdag/core/Limits.java
```diff
@@ -8,15 +12,23 @@
...
             System.getProperty("io.digdag.limits.maxAttempts", "100"));
 
-    // TODO (dano): this should be configurable by config file etc and not just system property
+    private final long numOfMaxWorkflowTasks;
+    private final long numOfMaxAttempts;
+
+    @Inject
+    public Limits(Config systemConfig)
+    {
+        this.numOfMaxWorkflowTasks =  systemConfig.get("executor.task_max_run", long.class, MAX_WORKFLOW_TASKS);
```
komamitsu, line 21: 👍

yoyama:
@komamitsu  Just in case I changed `OperatorContext.getMaxWorkflowTasks()` to default method to avoid compile errors in test. Some tests implement OperatorContext.

PR #1432, opened 2020-07-07
Fix rerun_on description

y-ken:
follow-up for https://github.com/treasure-data/digdag/commit/dbc12e40fc86a3a0f0552a921b380c0f9dd55b72#diff-828f306a0ac2fd6cd570c28c4b2f1cd2

### DIFF
--- a/digdag-docs/src/operators/require.md
+++ b/digdag-docs/src/operators/require.md
@@ -74,7 +74,7 @@
 
   *rerun_on* control require> really kicks or not if the attempt for the dependent workflow already exists. 
   * *none* ... Not kick the workflow if the attempt already exists.
-  * *failed* ... Kick the workflow if the attempt exists and its result is not success.
+  * *failed* ... Kick the workflow if the attempt not exists or its result was not success.
   * *all* ... require> kick the workflow regardless of the result of the attempt.
 
 * **ignore_failure**: BOOLEAN


y-ken:
@LeenSun could you please take a look?

digdag-docs/src/operators/require.md
```diff
@@ -74,7 +74,7 @@
 
   *rerun_on* control require> really kicks or not if the attempt for the dependent workflow already exists. 
   * *none* ... Not kick the workflow if the attempt already exists.
-  * *failed* ... Kick the workflow if the attempt exists and its result is not success.
+  * *failed* ... Kick the workflow if the attempt not exists or its result was not success.
```
yoyama, line 77: I don't think _"attempt not exists"_ is proper because this option assumes existence of the attempt as the following description:
> *rerun_on* control require> really kicks or not if the attempt for the dependent workflow already exists.

yoyama (changes requested):
As comment, I don't think the fix is not proper.

digdag-docs/src/operators/require.md
y-ken: In the actual behavior, `rerun_on: failed` will generate a new attempt even if the attempt_id at the corresponding session_time has not been issued.  so, how about this description?
```
*failed* ... Kick the workflow if the attempt may exists and its result is not success. 
```

y-ken:
@yoyama Updated PR, Would you please take a look?

digdag-docs/src/operators/require.md
```diff
@@ -74,7 +74,7 @@
 
   *rerun_on* control require> really kicks or not if the attempt for the dependent workflow already exists. 
   * *none* ... Not kick the workflow if the attempt already exists.
-  * *failed* ... Kick the workflow if the attempt exists and its result is not success.
+  * *failed* ... Kick the workflow if the attempt may exists and its result is not success.
```
yoyama, line 77: Sorry, I can't understand the meaning of using _'may'_. in this description.

BTW, your original concern is `rerun_on: failed` with the situation when the attempt does not exist.
But rerun_on is only effective if the attempt exists as described. 
If it is confused or uncertain, adding description to rerun_on may be better as follows:

```
*rerun_on* control require> really kicks or not if the attempt for the dependent workflow already exists. 
If the attempt does not exist, the dependent workflow is always kicked.
```

y-ken: > If the attempt does not exist, the dependent workflow is always kicked.

sounds great!

y-ken:
@yoyama I have updated doc as you suggested. 🙆‍♂️

yoyama:
As result of large change in master branch, there are many conflicts happened.
Could you resolve it ?

y-ken:
@yoyama I have fixed conflicts. please take a look

PR #1444, opened 2020-08-13
`hive_engine_version` options for `td` operators

yui-knk:
We manage dig file which includes a lot of `td` operators and some jobs should be run by presto others should be hive.
We want to define engine_version of hive on only one place by using `_export:` and `td:`.
In such case `engine_version` is not enough because we need to switch engine_version value based on query engine.
Introducing `hive_engine_version` option reduces our management cost of engine_versions.

PR #1445, opened 2020-08-19
Not pass `engine_version` parameter if engine is presto

yui-knk:
We manage dig file which includes a lot of `td` operators
and some jobs should be run by presto others should be hive.
Query engine of some jobs are determined dynamically.
In such a case we need to check engine and determine engine_version
if `engine_version` is passed for both presto and hive.

But we can not control presto engine_version via API job creation endpoint.
Then it's useless effort to determine engine_version.
We will need to specify `engine_version` of only hive by this change.


yui-knk:
This another approach for https://github.com/treasure-data/digdag/pull/1444 but I decided to take original approach then close this PR.

Issue #1410, opened 2020-05-27
digdag v0.9.41 doesn't use HTTPS on RemotePluginLoader


NomadBlacky:
I'm getting the following errors in my environment.

```
+transform_report_by_athena:
  _export:
    plugin:
      repositories:
        - https://jitpack.io
      dependencies:
        - pro.civitaspo:digdag-operator-athena:0.1.5
    athena:
      region: ap-northeast-1
      database: foo
      table: bar
  +prepare_database:
    athena.query>: queries/athena_create_database.sql
    preview: false
```

```
2020-05-27 03:09:37.547 +0000 [ERROR] (2045@[0:debug_wf]+report_coordinator+reports^sub+for-0=target=0=%7B%22name%22%3A^sub+load^sub+transform_report_by_athena+prepare_database) io.digdag.core.agent.OperatorManager: Task failed with unexpected error: org.eclipse.aether.resolution.DependencyResolutionException: Could not transfer artifact pro.civitaspo:digdag-operator-athena:jar:0.1.5 from/to jcenter (http://jcenter.bintray.com/): status code: 403, reason phrase: Forbidden (403)
java.lang.RuntimeException: org.eclipse.aether.resolution.DependencyResolutionException: Could not transfer artifact pro.civitaspo:digdag-operator-athena:jar:0.1.5 from/to jcenter (http://jcenter.bintray.com/): status code: 403, reason phrase: Forbidden (403)
	at com.google.common.base.Throwables.propagate(Throwables.java:241)
	at io.digdag.core.plugin.RemotePluginLoader.resolveArtifacts(RemotePluginLoader.java:167)
	at io.digdag.core.plugin.RemotePluginLoader.load(RemotePluginLoader.java:119)
	at io.digdag.core.plugin.DynamicPluginLoader.loadCache(DynamicPluginLoader.java:66)
	at io.digdag.core.plugin.DynamicPluginLoader.lambda$load$0(DynamicPluginLoader.java:54)
	at com.google.common.cache.LocalCache$LocalManualCache$1.load(LocalCache.java:4718)
	at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3445)
	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2194)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2153)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2043)
[...]

PR #1446, opened 2020-08-20
Update filter conditions on AttemptResource

szyn:
Follow up of #1410
As pointed out [here](https://github.com/treasure-data/digdag/pull/1419/files#r468426935), the filter excluded the task containing the parent ID, but it causes an issue with validation does not work as expected.

### DIFF
--- a/digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
+++ b/digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
@@ -23,6 +23,7 @@ import io.digdag.core.session.ArchivedTask;
 import io.digdag.core.session.SessionStore;
 import io.digdag.core.session.SessionStoreManager;
 import io.digdag.core.session.StoredSessionAttemptWithSession;
+import io.digdag.core.session.StoredTask;
 import io.digdag.core.session.TaskRelation;
 import io.digdag.core.session.TaskStateCode;
 import io.digdag.core.workflow.*;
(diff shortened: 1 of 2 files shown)


digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
```diff
@@ -315,7 +315,7 @@ public Response startAttempt(RestSessionAttemptRequest request)
                 // If a group error has occurred,
                 // exclude the group's child/dynamically generated tasks from successTasks
                 // even if they are in SUCCESS state
-                .filter(task -> task.getParentId().isPresent() && !groupRetryErrorTaskIds.contains(task.getParentId().get()))
+                .filter(task -> !groupRetryErrorTaskIds.contains(task.getParentId().orNull()))
```
yoyama, line 318: Just in case, you want to add the task which has no parent in successTasks, right ?

komamitsu, line 318: > NullPointerException - if the specified element is null and this list does not permit null elements (optional)

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/List.html#contains(java.lang.Object)

Whether this method throws NullPointerException depends on the actual class's behavior and it may be a bit naive implementation. How about moving this filter right before `map` below? Like
```
                .filter(task -> !task.getFullName().contains("^sub"))
                .filter(task -> task.getState() == TaskStateCode.SUCCESS)
                .filter(task -> 
                    if (!task.getParentId().isPresent()) {
                        throw new IllegalArgumentException("Resuming successfully completed attempts is not supported");
                    }
                    !groupRetryErrorTaskIds.contains(task.getParentId().get())
                )
                .map(StoredTask::getId)
                      : 
```

szyn: @yoyama Yes, you're right.

szyn: Thank you for sharing that. I agree with you, and it sounds pretty good! 
I moved this filter based on your review: 4b25b60

komamitsu: 👍

komamitsu (approved):

PR #1447, opened 2020-08-24
Add task details as comments automatically to the query run by td operator

szyn:
Currently, it's hard to identify which the workflow runs the job because we don't have information in the query issued by a workflow.
This PR support to add task details as comments automatically to the top of the query run by td operator, the following information will be added.

- project_id
- project_name
- workflow_name
- session_id
- attempt_id
- task_name

The above information will help the admin/user to identify/debug the workflow from the job results easily.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
@@ -219,7 +219,7 @@ public class TdOperatorFactory
                     throw new ConfigException("Unknown 'engine:' option (available options are: hive and presto): "+engine);
             }
 
-            final String sql = createStmtComment(request) + stmt;
+            final String sql = createStmtComment(request).append(stmt).toString();
 
             TDJobRequest req = new TDJobRequestBuilder()
                     .setResultOutput(resultUrl.transform(t -> t.format(context.getSecrets())).orNull())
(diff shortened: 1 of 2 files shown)


digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
```diff
@@ -219,11 +219,13 @@ else if (createTable.isPresent()) {
                     throw new ConfigException("Unknown 'engine:' option (available options are: hive and presto): "+engine);
             }
 
+            final String sql = createStmtComment(request) + stmt;
```
komamitsu, line 222: Why not return `StringBuilder` from the method to avoid this String concatenation?

szyn: Good point, thank you. Updated: 9db104e

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
```diff
@@ -533,4 +535,15 @@ else if (c == DELIMITER_CHAR) {
             return escapedValue.toString();
         }
     }
+
+    @VisibleForTesting
+    static StringBuilder createStmtComment(TaskRequest request)
```
komamitsu, line 540: FYI: Another option may be a method `String wrapStmtWithComment(TaskRequest request, String sql)`

digdag-standards/src/test/java/io/digdag/standards/operator/td/TdOperatorFactoryTest.java
```diff
@@ -93,9 +110,10 @@ public void testTDJobRequestParamsWithEngineVersion()
         when(op.getDatabase()).thenReturn("testdb");
 
         TDJobRequest jobRequest = testTDJobRequestParams(projectPath, config);
+        String sql = createStmtComment(taskRequest).append("select 1").toString();
```
komamitsu, line 113: There is no test to check `createStmtComment()` works expectedly?

digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java
szyn: That's better 👍  Updated: 83f00c1

digdag-standards/src/test/java/io/digdag/standards/operator/td/TdOperatorFactoryTest.java
szyn: Added a test case for `wrapStmtWithComment()`  e55103b

komamitsu (approved):
👍

PR #1413, opened 2020-06-05
Limiting the number of tasks/loops in _parallel

szyn:
## Overview
In addition to current `_parallel` features, this PR will introduce the parallel number limitation.

In order to achieve it, this PR makes JSON objects can also be acceptable for `_parallel`, it's for `_parallel: {limit: N}` (N is an integer: 1, 2, 3, …) syntax.

### For task groups
If ``_parallel: {limit: N}`` parameter is set to a group, child tasks in the group run in parallel is limited to N. 
Note that grandchildren are not affected, same as current behavior.

### For loop operators
If ``_parallel: {limit: N}`` parameter is set, the number of loops running in parallel is limited to N. 
Note that tasks **in the loop** will be running in serial, same as current behavior.

---

## Examples

basic.dig

``_parallel: {limit: N}`` is set to a task group.

```yml
+prepare:
  _parallel:
    limit: 3
  +step1:
    sh>: sleep 2 && echo '${this.task_name}'
  +step2:
    sh>: sleep 2 && echo '${this.task_name}'
  +step3:
    sh>: sleep 2 && echo '${this.task_name}'
  +step4:
    sh>: sleep 2 && echo '${this.task_name}'
  +step5:
    sh>: sleep 2 && echo '${this.task_name}'
```

In this case, from step1, step2 and step3 run in parallel, then step4 and step5 run in parallel. +data1 and +data2 need to be successful.

<details>
<summary>Example task log</summary>
<pre>
<code>
..snip..
2020-06-05 18:12:17 +0900 [INFO] (main): Starting a new session project id=1 workflow name=basic session_time=2020-06-05T00:00:00+00:00
2020-06-05 18:12:18 +0900 [INFO] (0018@[0:default]+basic+prepare+step2): sh>: sleep 2 && echo '+basic+prepare+step2'
2020-06-05 18:12:18 +0900 [INFO] (0019@[0:default]+basic+prepare+step3): sh>: sleep 2 && echo '+basic+prepare+step3'
2020-06-05 18:12:18 +0900 [INFO] (0017@[0:default]+basic+prepare+step1): sh>: sleep 2 && echo '+basic+prepare+step1'
+basic+prepare+step1
+basic+prepare+step3
+basic+prepare+step2
2020-06-05 18:12:20 +0900 [INFO] (0019@[0:default]+basic+prepare+step4): sh>: sleep 2 && echo '+basic+prepare+step4'
[...]

Issue #950, opened 2019-01-25
[issue] digdag possibly fails silently


sonots-zozo:
I encountered a situation that the digdag failed, but no error message is shown on digdag Web UI.
It is probably because this log is with trace level.

https://github.com/treasure-data/digdag/blob/c51f2040b2eeeffad6e77a7ba435746c2e255619/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java#L1116-L1117

This article (Japanese) http://labs.septeni.co.jp/entry/2018/05/18/205544 also says that he encountered a situation that he could not find the reason without --log-level trace.

```
2018-04-27 04:41:10 +0000 [TRACE]
...
io.digdag.core.workflow.WorkflowExecutor: Task failed with error {"message":"Too many tasks. Limit: 1000, Current: 999, Adding: 4 (task limit exceeded)"
```

This log is about failure of digdag, so I feel the log level should be ERROR level rather than TRACE. 
Any reasons that the level is TRACE? Any plans to change the log level?

muga:
Let me take a look. If trace log level only can show the error message, that's not good.

muga:
closing it. #951 will mitigate this issue.

sonots:
#951 looks that it is only for task limit errors.
I doubt that my error might be not related to the task limit errors. 

How about changing this line to ERROR level?

PR #1448, opened 2020-08-28
Show error message when tasks failed with config error in WorkflowExecutor

szyn:
Follow up of #1413 
Similar to https://github.com/treasure-data/digdag/issues/950

Currently, when an error occurs due to misconfiguration at a subtask, it could silently fail without any messages because of a log with a trace level as follows.
https://github.com/treasure-data/digdag/blob/069ed09f3cac01620eeaaaf6ff931349f7dcf0f6/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java#L1249-L1252

It would be better to output the log at the error level in case of such an error. This PR supports to show error messages as well.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
+++ b/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
@@ -1330,7 +1330,11 @@ public class WorkflowExecutor
         }
         catch (ConfigException ex) {
             // Subtask config or _check task config has a problem (e.g. more than one operator).
-            return taskFailed(lockedTask, buildExceptionErrorConfig(ex).toConfig(cf));
+            Config errorConfig = buildExceptionErrorConfig(ex).toConfig(cf);
+            logger.error("Configuration error at task {}: {}",
+                    lockedTask.get().getFullName(),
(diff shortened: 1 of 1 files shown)


digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
```diff
@@ -1330,7 +1330,11 @@ private boolean taskSucceeded(TaskControl lockedTask,
         }
         catch (ConfigException ex) {
             // Subtask config or _check task config has a problem (e.g. more than one operator).
-            return taskFailed(lockedTask, buildExceptionErrorConfig(ex).toConfig(cf));
+            Config errorConfig = buildExceptionErrorConfig(ex).toConfig(cf);
+            logger.error("Configuration error at task {}: {}",
```
komamitsu, line 1334: Looks good.

But I'm also wondering if we'd better change the log level of https://github.com/treasure-data/digdag/blob/cefde0d80a6f2a573ac800454da8b4fa7f3e87ad/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java#L1251 to `warn` or `error`

komamitsu (approved):

PR #1450, opened 2020-08-28
Add a log marker "UNEXPECTED_SERVER_ERROR"

komamitsu:
digdag outputs some log messages as `error` level, but some of them are unexpected and the others are expected. But there is no marker on log messages now to distinct whether it's expected or not. This PR adds a marker on unexpected log messages so that we can handle unexpected ones differently.


szyn (approved):

komamitsu:
@szyn Thank you~

PR #1464, opened 2020-10-09
Add getTaskDefinitionByTags with functional condition

serihiro:
# What does this PR change?

This PR introduces a method to filters ECS Task definitions by `Predicate` object.
`Optional<TaskDefinition> getTaskDefinitionByTags(final Predicate<List<Tag>> condition)`

# Background

Existing `Optional<TaskDefinition> getTaskDefinitionByTags(final List<Tag> expectedTags)` filters ECS Task definitions by matching all provided Tags.

In order to filter ECS Task definitions more flexibility, we need another method which takes functional interface object as a condition.

One of the expected use case is the following

```java
final Predicate<List<Tag>> condition = tags -> tags.stream().allMatch(t -> t.getKey().startsWith("v2_"));
final Optional<TaskDefinition> v2TaskDefinitions = ecsClient.getTaskDefinitionByTags(condition);
```

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/ecs/DefaultEcsClient.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/ecs/DefaultEcsClient.java
@@ -30,6 +30,7 @@ import com.amazonaws.services.logs.model.GetLogEventsRequest;
 import com.amazonaws.services.logs.model.GetLogEventsResult;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Optional;
+import com.google.common.base.Predicate;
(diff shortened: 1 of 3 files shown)


myui (approved):
LGTM. 

Minor comment:
I personally prefer java.util.functions.Predicate and java.util.Optional over Guava since Guava library can cause dependency hell.
Since Muga used Guava for optional, it's might be historical reasons.

serihiro:
Stop using Guava library is good idea since now digdag supports only Java >= 8 officially. But since they are used in many places of digdag codebases, I am merging this PR as is for now to keep consistency.

PR #1468, opened 2020-10-20
Fix a bug that failed tasks with java.lang.Error retry infinitely

komamitsu:
Current `OperatorManager#runWithHeartbeat` doesn't catch `java.lang.Error` derived objects including `java.lang.OutOfMemoryError`. If a task always fails due to OOM (e.g. internally parse a huge JSON object), the task will be retried infinitely and the task TTL (default: 24 hours) won't work.

This PR fixes the bug by checking if the task is already requested to be canceled by `WorkflowExecutionTimeoutEnforcer` and cancels the task if needed.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -177,6 +177,19 @@ public class OperatorManager
                     }
                     callback.taskFailed(request, agentId, buildExceptionErrorConfig(ex).toConfig(cf));  // no retry
                 }
+                catch (Throwable ex) {
+                    logger.error(
+                            LogMarkers.UNEXPECTED_SERVER_ERROR,
+                            "Task failed with unexpected error: {}", ex.getMessage(), ex);
+
+                    // This block catches OutOfMemoryError and its cause can be either deterministic or non-deterministic.
+                    // So, OperatorManager can't always cancel the failed task.
+                    if (request.isCancelRequested()) {
+                        logger.warn("This task will be canceled since it's already requested to be canceled");
+                        // This call cancels a task if it's requested to be canceled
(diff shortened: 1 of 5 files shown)


digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -177,6 +177,19 @@ protected void runWithHeartbeat(TaskRequest request)
...
                     callback.taskFailed(request, agentId, buildExceptionErrorConfig(ex).toConfig(cf));  // no retry
                 }
+                catch (Throwable ex) {
+                    logger.error(
+                            LogMarkers.UNEXPECTED_SERVER_ERROR,
+                            "Task failed with unexpected error: {}", ex.getMessage(), ex);
+
+                    // This block catches OutOfMemoryError and its cause can be either deterministic or non-deterministic.
+                    // So, OperatorManager can't always cancel the failed task.
+                    if (request.isCancelRequested()) {
```
yoyama, line 187: In this implementation, only when cancel is requested, task will finish with failure.
If canceling request is not set to the task, the task will continue to run with failure until TTL.
Is my understanding correct ? If so, is there any reason to wait for TTL ?

komamitsu: This `catch` block can handle non-deterministic retryable error like OutOfMemoryError caused by temporary lack of overall JVM heap not only deterministic error like OOM caused by instantiation of too huge String. So we can't decide to cancel the failed task immediately. But if the failed task has been already running more than 24 hours, we can cancel it without any concern.

yoyama: Make sense.  👍

yoyama (approved):

szyn (approved):

komamitsu:
Thanks!

PR #1470, opened 2020-10-22
[WIP] Update key name of task creation timestamp

szyn:
This PR aligns the key name of the task creation timestamp, it should be `task_creation_timestamp`.
e.g.
https://github.com/treasure-data/digdag/blob/f20567ba7dfa2b2948bd036939fbe828bef09e25/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java#L260-L265

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
@@ -436,7 +436,7 @@ public class EcsCommandExecutor
 
     private boolean isRunningLongerThanTTL(final ObjectNode previousStatus)
     {
-        long creationTimestamp = previousStatus.get("pod_creation_timestamp").asLong();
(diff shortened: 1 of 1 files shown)


digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -436,7 +436,7 @@ private static String toLogStreamName(final ObjectNode previousStatus)
 
     private boolean isRunningLongerThanTTL(final ObjectNode previousStatus)
     {
-        long creationTimestamp = previousStatus.get("pod_creation_timestamp").asLong();
+        long creationTimestamp = previousStatus.get("task_creation_timestamp").asLong();
```
komamitsu, line 439: During migration (== restart of Digdag server application with new code), application could handle old status that contains `pod_creation_timestamp`?  If so, I think this PR needs to take care of old key name?

szyn:
We will introduce a cleanup hook (#1480) instead of this PR. Let me close this one.

PR #1473, opened 2020-10-30
Make launch type optional

serihiro:
# What does this PR change?
This PR makes `launchType` `Optional<String>` field from `String`.

# Background
In the specification of AWS ECS API, we can not specify `launchType` {`EC2`/`FARGATE`} and `capacityProviderStragety` at the same time. So `launchType` needs to be `Optional` field.


yoyama (approved):
Please add milestone

PR #1469, opened 2020-10-22
Support creating EcsClientConfig from taskConfig

serihiro:
# What does this PR change?
This PR introduces 
1. a support to create an `EcsClientConfig` instance from taskConfig
1. new properties which can be specified from `EcsClientConfig`
- a support to specify `capacityProvider` on a `RunTaskRequest`
- a support to specify `cpu` on a `ContainerOverride`
- a support to specify `memory` on a `ContainerOverride`

# Background
In the current implementation, we can not create `EcsClientConfig` instances from `taskConfig` (we can create them only from `systemConfig`). 
So it is difficult to execute Ecs tasks with different configurations for each Ecs task.

PR #1474, opened 2020-10-30
Fetch sensitive data from `systemConfig`

serihiro:
# What will this PR change?
This PR refactors `EcsClientConfig#createFromTaskConfig` so that the method fetches `access_key_id` and `secret_access_key` from `systemConfig` to build a `EcsClientConfig` object.

# Background
Basically `taskConfig` is generated from workflow definition. 
So the object should not have any sensitive data because `taskConfig` is referred as a object without sensitive data.

This PR refactors `EcsClientConfig#createFromTaskConfig` so that the method uses both of `taskConfig` and `systemConfig` to fetch parameters for `EcsClientConfig` object.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfig.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfig.java
@@ -33,28 +33,30 @@ public class EcsClientConfig
         this.assignPublicIp = builder.isAssignPublicIp();
     }
 
-    public static EcsClientConfig createFromTaskConfig(final Optional<String> clusterName, final Config config)
+    public static EcsClientConfig createFromTaskConfig(final Optional<String> clusterName, final Config taskConfig, final Config systemConfig)
     {
         final String name;
-        // `config` is assumed to have a nested config with following values
-        // at the key of `TASK_CONFIG_ECS_KEY` from `config`.
+        // `taskConfig` is assumed to have a nested taskConfig with following values
+        // at the key of `TASK_CONFIG_ECS_KEY` from `taskConfig`.
         // - launch_type (optional)
-        // - access_key_id
-        // - secret_access_key
         // - region
         // - subnets (optional)
         // - max_retries (optional)
         // - capacity_provider_name (optional)
         // - memory (optional)
         // - cpu (optional)
-        final Config ecsConfig = config.getNested(TASK_CONFIG_ECS_KEY);
+        final Config ecsConfig = taskConfig.getNested(TASK_CONFIG_ECS_KEY).deepCopy();
         if (!clusterName.isPresent()) {
-            // Throw ConfigException if 'name' doesn't exist in system config.
+            // Throw ConfigException if 'name' doesn't exist in system ecsConfig.
             name = ecsConfig.get("cluster_name", String.class);
         }
         else {
             name = clusterName.get();
         }
+
+        // This method assumes that `access_key_id` and `secret_access_key` are stored at `systemConfig`.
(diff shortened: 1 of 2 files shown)


digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfig.java
```diff
@@ -33,29 +33,31 @@ public EcsClientConfig(EcsClientConfigBuilder builder)
...
-    public static EcsClientConfig createFromTaskConfig(final Optional<String> clusterName, final Config config)
+    public static EcsClientConfig createFromTaskConfig(final Optional<String> clusterName, final Config taskConfig, final Config systemConfig)
     {
         final String name;
-        // `config` is assumed to have a nested config with following values
-        // at the key of `TASK_CONFIG_ECS_KEY` from `config`.
+        // `taskConfig` is assumed to have a nested taskConfig with following values
+        // at the key of `TASK_CONFIG_ECS_KEY` from `taskConfig`.
         // - launch_type (optional)
-        // - access_key_id
```
komamitsu, line 42: Where were these fields set in the nested config?

serihiro: I expect that these fields are set in the caller (e.g. `CommandExecutor` classes)

komamitsu: Oh, I just thought we might need to remove these system credentials (access_key_id and secret_access_key) from all task configs as well in this PR., but you didn't add any code that puts the credentials in a task config? If so, I'm not sure these system credentials are set by callers...

serihiro: Ah sorry. Maybe I wrongly understood what you meant. 

I thought that you have mentioned about **ALL** fields of them ( including `launch_type`, `region`, `subnets` and so on). But it seems that you mentioned about **ONLY** `access_key_id` and `secret_access_key`, is it correct?

With explaining the background from the beginning,

1. In the first time when `EcsCommandExecutor` was released, `access_key_id` and `secret_access_key` were expected to be put only into `systemConfig`. So `EcsClientConfig` had been built only from `systemConfig`.
1. At https://github.com/treasure-data/digdag/pull/1469, I added a support of `building EcsClientConfig from taskConfig` so that we can change the contents of `EcsClientConfig` for each `taskConfig`.
1. After merging #1469, I thought that sensitive data should not be included in `taskConfig`. So I sent this PR so that `access_key_id` and `secret_access_key` are retrieved only from `systemConfig`.

> Oh, I just thought we might need to remove these system credentials (access_key_id and secret_access_key) from all task configs as well in this PR.

Yes, you are correct. 

>  but you didn't add any code that puts the credentials in a task config?  

Implicitly it depended on users whether `access_key_id` and `secret_access_key` were put into the taskConfig not. Surely it was not good implementation. But this PR fixes it as well.

So this PR does not include any code change to remove system credentials from task configs. 

In the original implementation, `access_key_id` and `secret_access_key` were expected from `server.properties` and can be fetched with `agent.command_executor.ecs.{cluster_name}.access_key_id` and `agent.command_executor.ecs.{cluster_name}.secret_access_key` repeatedly as follows.
https://github.com/treasure-data/digdag/blob/ead8c27608b6e01a644a794c1910d6b4f50a63cb/digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfig.java#L48-L57

>  If so, I'm not sure these system credentials are set by callers...

As I explained above, they were expected to be set `systemConfig` implicitly. So originally there has not been the explicit implementations to do that.

komamitsu (approved):

PR #1475, opened 2020-11-04
Avoid potential deadlock on `tasks` table

komamitsu:
We've observed PostgreSQL used by Digdag detected deadlocks.

```
2020-10-26 23:00:55 UTC:xxxx(63248):xxxx@xxxxx:[90713]:ERROR:  deadlock detected
2020-10-26 23:00:55 UTC:xxxx(63248):xxxx@xxxxx:[90713]:DETAIL:  Process 90713 waits for ShareLock on transaction 1911558785; blocked by process 88119.
    Process 88119 waits for ShareLock on transaction 1911558683; blocked by process 90713.
    Process 90713: update tasks set updated_at = now(), retry_at = NULL, state = 1 where state in (2,3) and retry_at <= now()
    Process 88119: select id from tasks where id = $1 for update
```

It looks this happened due to the inconsistent locking orders between
multiple `select id from tasks where id = $1 for update` called by  https://github.com/treasure-data/digdag/blob/2678bf62a17695c4b67af37113a842e94ca09f31/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java#L32-L37 and `update tasks set state = ${TaskStateCode.READY_CODE} where state in (...) and retry_at <= now()` called by https://github.com/treasure-data/digdag/blob/8baa1e7db395d3a2d441ef1243c1f1da644f3c35/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java#L880.

This PR makes the 2 update operations take a lock in the same order.

A slight drawback of this change is that the new UPDATE statement to move tasks status to READY will be slowed down since the new UPDATE statement uses `UPDATE ... FROM SELECT ... ORDER BY id FOR UPDATE` and I estimate the new version query is a few times slower. But from the metrics of our production environment that executes about 770K tasks per day, the target tasks, in other words "retryable" tasks, is only 6 on average and 22 at most. So I think this doesn't matter.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
@@ -24,7 +24,7 @@ public class InProcessTaskServerApi
             return Long.parseLong(tql.getUniqueName());
         }
         catch (Throwable e) {
-            logger.warn("Failed to convert TaskQueueLock.uniqueName to integer", e);
+            logger.warn("Failed to convert TaskQueueLock.uniqueName to integer. The `uniqueName` will be handled as 0", e);
             return 0;
         }
     };
--- a/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
@@ -18,8 +18,8 @@ import java.util.stream.Collectors;
 public class InProcessTaskServerApi
     implements TaskServerApi
 {
-    private static final Logger logger = LoggerFactory.getLogger(WorkflowExecutor.class);
-    private final ToLongFunction<TaskQueueLock> CONV_FUNC_FROM_TASK_QUEUE_LOCK_TO_INT = taskQueueLock -> {
+    private static final Logger logger = LoggerFactory.getLogger(InProcessTaskServerApi.class);
+    private static final ToLongFunction<TaskQueueLock> CONV_FUNC_FROM_TASK_QUEUE_LOCK_TO_INT = taskQueueLock -> {
         try {
             return WorkflowExecutor.parseTaskIdFromEncodedQueuedTaskName(taskQueueLock.getUniqueName());
         }
--- a/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
@@ -19,12 +19,9 @@ public class InProcessTaskServerApi
     implements TaskServerApi
 {
     private static final Logger logger = LoggerFactory.getLogger(WorkflowExecutor.class);
-    private final ToLongFunction<TaskQueueLock> CONV_FUNC_FROM_TASK_QUEUE_LOCK_TO_INT = tql -> {
+    private final ToLongFunction<TaskQueueLock> CONV_FUNC_FROM_TASK_QUEUE_LOCK_TO_INT = taskQueueLock -> {
         try {
-            // Retried task's unique name has suffix `.r${retryCount}`.
-            // See io.digdag.core.workflow.WorkflowExecutor.encodeUniqueQueuedTaskName
-            String[] decodedPartsOfUniqueName = tql.getUniqueName().split("\\.");
-            return Long.parseLong(decodedPartsOfUniqueName[0]);
+            return WorkflowExecutor.parseTaskIdFromEncodedQueuedTaskName(taskQueueLock.getUniqueName());
         }
         catch (Throwable e) {
             logger.warn("Failed to convert TaskQueueLock.uniqueName to integer. The `uniqueName` will be handled as 0", e);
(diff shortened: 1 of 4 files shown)


digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
```diff
@@ -7,11 +7,28 @@
...
+
+import java.util.Comparator;
 import java.util.List;
+import java.util.function.ToLongFunction;
+import java.util.stream.Collectors;
 
 public class InProcessTaskServerApi
     implements TaskServerApi
 {
+    private static final Logger logger = LoggerFactory.getLogger(WorkflowExecutor.class);
```
yoyama, line 21: `InProcessTaskServerApi.class` is better ?

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
```diff
@@ -7,11 +7,28 @@
...
+    private final ToLongFunction<TaskQueueLock> CONV_FUNC_FROM_TASK_QUEUE_LOCK_TO_INT = taskQueueLock -> {
```
yoyama, line 22: I am curious the naming rule of the function.
Is there any reference to use capital letter with _ for the var name of function ?

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
```diff
@@ -7,11 +7,28 @@
...
+        try {
+            return WorkflowExecutor.parseTaskIdFromEncodedQueuedTaskName(taskQueueLock.getUniqueName());
+        }
+        catch (Throwable e) {
+            logger.warn("Failed to convert TaskQueueLock.uniqueName to integer. The `uniqueName` will be handled as 0", e);
+            return 0;
```
yoyama, line 28: Is there any possible case to execute this catch(){} code or just in case ?

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -1745,6 +1759,18 @@ StoredSession upsertAndLockSession(@Bind("projectId") int projectId,
...
 
+        @SqlUpdate("update tasks" +
+                " set updated_at = now(), retry_at = NULL, state = " + TaskStateCode.READY_CODE +
+                " from (" +
+                  "select id from tasks" +
+                  " where state in (" + TaskStateCode.RETRY_WAITING_CODE +"," + TaskStateCode.GROUP_RETRY_WAITING_CODE + ")" +
+                  " and retry_at \\<= now()" +
+                  " order by id for update" +
+                " ) lck" +
+                " where tasks.id = lck.id"
```
yoyama, line 1770: IIRC the purpose of the sub query is fixing the order of lock of the rows.
If so, is there any formal documentation description on it ?
I have a minor concern whether it depends on the specific Pg version or undocumented internal behavior or not.

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
komamitsu: Oh, good catch! I moved the code from `WorkflowExecutor` carelessly.
I'll fix it soon. Thanks

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
komamitsu: Oops. I should have made this `static` since this variable is a constant... I'll fix it

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
komamitsu: Basically for just-in-case. I was thinking of case that `uniqueName` unexpectedly contains non-digit string although I"m not sure it could happen.

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
komamitsu: As far as I know, the official document explicitly doesn't mention this issue and behavior. But we can find some explanations about the issue.

> UPDATE deals with rows in order they are fetched from a heap. In a common case it can be considered as unordered.

https://www.postgresql.org/message-id/CAKOSWNkb3Zy_YFQzwyRw3MRrU10LrMj04%2BHdByfQu6M1S5B7mg%40mail.gmail.com

And I think you can easily find this workaround (e.g. in the above post or https://stackoverflow.com/a/27263824/5101185) although I can't 100% guarantee this behavior won't change forever.

digdag-core/src/main/java/io/digdag/core/agent/InProcessTaskServerApi.java
komamitsu: fyi: `Constants` on https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html just in case

komamitsu:
@yoyama Addressed the feedbacks from you and put a few explanations. But if you still have a concern, maybe we can leave this PR and would revisit here when needed.

yoyama (approved):
LGTM

komamitsu:
Thanks!

PR #1476, opened 2020-11-04
Support task provider strategy

serihiro:
# What does this PR change?
This PR supports `PlacementStrategyType` and `PlacementStrategyField` at `EcsClientConfig` and `EcsCommandExecutor`

# Background
AWS ECS supports `Task Placement Strategies` option to specify a container instance at where an ECS task will run. https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-placement-strategies.html

This PR enables user to specify `TaskPlacementStrategy` and `TaskPlacementField` for each ECS task.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
@@ -536,7 +536,8 @@ public class EcsCommandExecutor
         return runTaskRequest;
     }
 
-    private void setPlacementStrategy(EcsClientConfig clientConfig, RunTaskRequest runTaskRequest) throws ConfigException
+    private void setPlacementStrategy(EcsClientConfig clientConfig, RunTaskRequest runTaskRequest)
+            throws ConfigException
     {
         if (clientConfig.getPlacementStrategyType().isPresent()) {
             final PlacementStrategyType placementStrategyType;
--- a/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
@@ -9,6 +9,8 @@ import com.amazonaws.services.ecs.model.KeyValuePair;
 import com.amazonaws.services.ecs.model.LaunchType;
 import com.amazonaws.services.ecs.model.LogConfiguration;
 import com.amazonaws.services.ecs.model.NetworkConfiguration;
+import com.amazonaws.services.ecs.model.PlacementStrategy;
+import com.amazonaws.services.ecs.model.PlacementStrategyType;
 import com.amazonaws.services.ecs.model.RunTaskRequest;
 import com.amazonaws.services.ecs.model.RunTaskResult;
 import com.amazonaws.services.ecs.model.Tag;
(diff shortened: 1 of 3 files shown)


digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -530,9 +532,31 @@ RunTaskRequest buildRunTaskRequest(
...
     }
 
+    private void setPlacementStrategy(EcsClientConfig clientConfig, RunTaskRequest runTaskRequest) throws ConfigException
+    {
+        if (clientConfig.getPlacementStrategyType().isPresent()) {
+            final PlacementStrategyType placementStrategyType;
+            try {
+                placementStrategyType = PlacementStrategyType.fromValue(clientConfig.getPlacementStrategyType().get());
+            }
+            catch (IllegalArgumentException invalidReason) {
```
komamitsu, line 546: [minor] `invalidReason` sounds a bit weird since this is a just exception here. If you extracted a pure reason-like object from an exception like with `Reason invalidReason = ex.getReason()`, it would be good.

digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfig.java
```diff
@@ -31,6 +32,12 @@ public EcsClientConfig(EcsClientConfigBuilder builder)
         this.memory = builder.getMemory();
         this.startedBy = builder.getStartedBy();
         this.assignPublicIp = builder.isAssignPublicIp();
+        this.placementStrategyType = builder.getPlacementStrategyType();
+        this.placementStrategyField = builder.getPlacementStrategyField();
+
+        if (!placementStrategyType.isPresent() && placementStrategyField.isPresent()) {
+            throw new ConfigException("PlacementStrategyField must be set with PlacementStrategyType");
+        }
```
komamitsu, line 40: Don't we need to check the other case like `placementStrategyType.isPresent() && !placementStrategyField.isPresent()`?

komamitsu:
This PR doesn't need to have some test cases?

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
serihiro: What I meant was just "This is an `IllegalArgumentException` object and the actual reason is stored in this exception object's message". But surely it is wired a bit. I will rename the exception object and leave a comment to suggest it.

digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfig.java
serihiro: It depends what kind of `placementStrategyType` users use.
It's complicated a little, some `placementStrategyType` requires `placementStrategyField` but others does not.

But at least `placementStrategyField` only is invalid usage. That's the reason why I added the validation here.

Regarding more detail validations, I decided not to implement it here because it is complicated to cover all patterns as I told above. I assume digdag users who use EcsCommandExecutor have the knowledge about ECS placement strategy option.

serihiro:
> This PR doesn't need to have some test cases?

Regarding testing I will take care in other PR because there are NO tests for `EcsCommandExecutor` so far including `EcsClientConfig`. But we do not have enough spaces to do it for some reasons so far.

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -530,9 +532,31 @@ RunTaskRequest buildRunTaskRequest(
         setEcsTaskStartedBy(clientConfig, runTaskRequest);
         setEcsNetworkConfiguration(clientConfig, runTaskRequest);
         setCapacityProviderStrategy(clientConfig, runTaskRequest);
+        setPlacementStrategy(clientConfig, runTaskRequest);
         return runTaskRequest;
     }
 
+    private void setPlacementStrategy(EcsClientConfig clientConfig, RunTaskRequest runTaskRequest) throws ConfigException
```
komamitsu, line 539: `throws` should be at next line

komamitsu:
Oh, I didn't suggest that you write all test cases of `EcsCommandExecutor`, but just for the changes this PR includes. I have no ideas what `for some reasons` meant, though...

serihiro:
@komamitsu 
I fixed this PR according to your comments. Could you check this again?

komamitsu:
LGTM

> Regarding testing I will take care it at other PR

Please handle this later

komamitsu (approved):

PR #1477, opened 2020-11-06
Enhance py> operator error handling

yoyama:
Some of errors does not handling in runner.py of py>. And it cause no generation of output.json and users cannot get information of errors.
To mitigate it, change error handling in runner.py as follows:

- catch exception in whole main code
- handle SystemExit exception because it only show exit code and not kind to users.

### DIFF
--- a/digdag-standards/src/main/resources/digdag/standards/py/runner.py
+++ b/digdag-standards/src/main/resources/digdag/standards/py/runner.py
@@ -160,8 +160,8 @@ try:
         args = digdag_inspect_arguments(callable_type, False, params)
         result = callable_type(**args)
 except SystemExit as e:
-    # SystemExit only show exit code and it is not kind to user. So create specific error message.
-    # This error will happen if called python module name and method name are equal to standard module. (e.g tokenize.main)
+    # SystemExit only shows an exit code and it is not kind to users. So this block creates a specific error message.
+    # This error will happen if called python module name and method name are equal to those of the standard library module. (e.g. tokenize.main)
     error = Exception("Failed to call python command with code:%d" % e.code, "Possible cause: Ivalid python module call, duplicae module name with standard library")
     error_type, error_value, _tb = sys.exc_info()
     error_message = "%s %s" % (error.args[0], error.args[1])
--- a/digdag-standards/src/main/resources/digdag/standards/py/runner.py
+++ b/digdag-standards/src/main/resources/digdag/standards/py/runner.py
@@ -140,32 +140,37 @@ def digdag_inspect_arguments(callable_type, exclude_self, params):
     else:
         return args
 
-callable_type, method_name = digdag_inspect_command(command)
 error = None
+error_message = None
 error_value = None
 error_traceback = None
+callable_type = None
+method_name = None
 
-if method_name:
-    init_args = digdag_inspect_arguments(callable_type.__init__, True, params)
-    instance = callable_type(**init_args)
+try:
+    callable_type, method_name = digdag_inspect_command(command)
+    if method_name:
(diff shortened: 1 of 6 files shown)


digdag-standards/src/main/resources/digdag/standards/py/runner.py
```diff
@@ -140,32 +140,37 @@ def digdag_inspect_arguments(callable_type, exclude_self, params):
...
+    else:
+        args = digdag_inspect_arguments(callable_type, False, params)
         result = callable_type(**args)
-    except Exception as e:
-        error = e
-        error_type, error_value, _tb = sys.exc_info()
-        error_traceback = traceback.format_exception(error_type, error_value, _tb)
+except SystemExit as e:
+    # SystemExit only show exit code and it is not kind to user. So create specific error message.
+    # This error will happen if called python module name and method name are equal to standard module. (e.g tokenize.main)
```
serihiro, line 164:
```suggestion
    # This error will happen if called python module name and method name are equal to those of the standard library module. (e.g. tokenize.main)
```

serihiro (approved):
I left miner comments but the design and implementation are LGTM.

yoyama:
@serihiro Thanks for your review. I will fix the comments.

digdag-standards/src/main/resources/digdag/standards/py/runner.py
```diff
@@ -140,32 +140,37 @@ def digdag_inspect_arguments(callable_type, exclude_self, params):
...
+        args = digdag_inspect_arguments(callable_type, False, params)
         result = callable_type(**args)
-    except Exception as e:
-        error = e
-        error_type, error_value, _tb = sys.exc_info()
-        error_traceback = traceback.format_exception(error_type, error_value, _tb)
+except SystemExit as e:
+    # SystemExit only shows an exit code and it is not kind to users. So this block creates a specific error message.
+    # This error will happen if called python module name and method name are equal to those of the standard library module. (e.g. tokenize.main)
+    error = Exception("Failed to call python command with code:%d" % e.code, "Possible cause: Ivalid python module call, duplicae module name with standard library")
```
chezou, line 165: As of Python 3.6, f-string enables writing string format shorter, but if we need to support Python 2.7 (while it is EOL), it'd be common to use `format` rather than `%`. https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals


```suggestion
    error = Exception("Failed to call python command with code:{}".format(e.code), "Possible cause: Ivalid python module call, duplicae module name with standard library")
```

digdag-standards/src/main/resources/digdag/standards/py/runner.py
```diff
@@ -140,32 +140,37 @@ def digdag_inspect_arguments(callable_type, exclude_self, params):
...
+    error_type, error_value, _tb = sys.exc_info()
+    error_message = "%s %s" % (error.args[0], error.args[1])
```
chezou, line 167:
```suggestion
    error_message = "{} {}".format(error.args[0], error.args[1])
```

chezou (commented):
Left minor comments around string formatting.

yoyama:
@chezou Thank you so much your review. I will confirm the policy on the supported versions of Python and then fix if needed.

yoyama:
@chezo As rough policy, I would like to drop Python 2.7 support.
This is v0_10 branch and it will be v0.10.0 in the future. So it is good timing to clarify the drop of Python2 support.
I am not sure 3.5 should be supported or not, but current implementation works in 3.5 and it may be better.
So I would like to keep the current implementation.
Anyway thank you so much your review !

PR #1480, opened 2020-11-10
Update handling of a task with CANCEL_REQUESTED flag

szyn:
## Overview

Currently, there is a problem with post-processing when a task is canceled.

This PR adds support for handling cleanup when a task is canceled, with the following changes.

### WorkflowExecutor
- Don't `setToCanceled` on enqueueTask and let the OperatorManager do the canceling (= canceled by `taskFailed`).
  - If a task has a CANCEL_REQUESTED flag during/after execution, it is necessary to perform the cleanup process.
  - And task transitions will be changed as follows:
    - As-Is: RETRY_WAITING > READY > CANCELED
    - To-Be: RETRY_WAITING > READY > RUNNING > CANCELED https://github.com/treasure-data/digdag/blob/520baa783235545def0492c98455cad1d56ffd5f/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java#L72-L162
  - ^ This allows cleanup to be performed while the task is running/finished

### OperatorManager
- Changes to execute cleanup when the task is CANCEL_REQUESTED

### CommnadExecutor/Operator
- Adds cleanup functions to the Operator and CommnadExecutor interfaces and implements them in PyOperatorFactory and EcsCommandExecutor.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -160,8 +160,13 @@ public class OperatorManager
                         callback.retryTask(request, agentId, ex.getRetryInterval().get(), ex.getStateParams(cf).get(), ex.getError(cf));
                     }
                     else {
-                        logger.error("Task {} failed.\n{}", request.getTaskName(), formatExceptionMessage(ex));
-                        logger.debug("", ex);
+                        if (request.isCancelRequested()) {
+                            logger.warn("Task {} canceled.", request.getTaskName());
+                        }
+                        else {
+                            logger.error("Task {} failed.\n{}", request.getTaskName(), formatExceptionMessage(ex));
+                            logger.debug("", ex);
+                        }
                         // TODO use debug to log stacktrace here
                         callback.taskFailed(request, agentId, ex.getError(cf).get());  // TODO is error set?
                     }
@@ -345,6 +350,10 @@ public class OperatorManager
                 projectPath, mergedRequest, secretProvider, privilegedVariables, limits);
 
         Operator operator = factory.newOperator(context);
+
+        if (mergedRequest.isCancelRequested()) {
+            return operator.cleanup(mergedRequest);
+        }
 
         return operator.run();
     }
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -352,7 +352,7 @@ public class OperatorManager
         Operator operator = factory.newOperator(context);
 
         if (mergedRequest.isCancelRequested()) {
-            return operator.cleanup(mergedRequest);
+            operator.cleanup(mergedRequest);
         }
 
         return operator.run();
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -50,6 +50,7 @@ import java.util.stream.Collectors;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static io.digdag.spi.TaskExecutionException.buildExceptionErrorConfig;
+import static java.util.Locale.ENGLISH;
 
 public class OperatorManager
 {
@@ -352,7 +353,13 @@ public class OperatorManager
         Operator operator = factory.newOperator(context);
 
         if (mergedRequest.isCancelRequested()) {
+            // NOTE: In the case of failure to perform cleanup,
+            // target resources continue to remain
             operator.cleanup(mergedRequest);
+            // Throw exception to stop the task
+            throw new TaskExecutionException(String.format(ENGLISH,
+                    "Got a cancel-requested: attempt_id=%d, task_id=%d",
+                    mergedRequest.getAttemptId(), mergedRequest.getTaskId()));
         }
 
         return operator.run();
--- a/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
@@ -161,7 +161,7 @@ public class OperatorManager
                     }
                     else {
                         if (request.isCancelRequested()) {
-                            logger.warn("Task {} canceled.", request.getTaskName());
+                            logger.warn("Task {} is canceled.", request.getTaskName());
                         }
                         else {
                             logger.error("Task {} failed.\n{}", request.getTaskName(), formatExceptionMessage(ex));
--- a/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
+++ b/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
@@ -86,8 +86,6 @@ import static java.util.Locale.ENGLISH;
  * READY:
  *   enqueueReadyTasks:
  *     enqueueTask:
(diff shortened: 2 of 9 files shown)


digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java
```diff
@@ -104,6 +104,8 @@
  *       : ERROR with error
  *
  *   taskSucceeded:
+ *     (if CANCEL_REQUESTED flag is set) lockedTask.setToCanceled:
```
komamitsu, line 107: This is outside the scope of this PR, but I'm wondering a successfully finished task should be handled as SUCCESS not as CANCELED even if it's received a cancel request...

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -289,6 +282,42 @@ public CommandStatus poll(
...
+            try {
+                task = client.getTask(clusterName, taskArn);
+            }
+            catch (TaskSetNotFoundException e) {
+                final String message = s("Cannot get the ECS task status. attemptId=%d, taskId=%d", attemptId, taskId);
+                logger.warn(message);
+                // Throw exception to stop the task
+                throw new TaskExecutionException(message);
+            }
+            final String message = s("Command task execution cancel requested: attemptId=%d, taskId=%d", attemptId, taskId);
```
komamitsu, line 312: The responsibility of this method is only cleaning up a task and cancel requests are not related to this method, I think. This method can be used by non-cancel operation in the future. So maybe this message shouldn't mention cancel requests.

digdag-standards/src/main/java/io/digdag/standards/operator/PyOperatorFactory.java
```diff
@@ -124,6 +125,25 @@ public TaskResult runTask()
...
+            Config state = TaskState.of(request).params();
+            if (state.has("commandStatus")) {
+                logger.debug(String.format("Starting cleanup: attemptId=%d, taskId=%d",  attemptId, taskId));
+                try {
+                    exec.cleanup(commandContext, state);
+                } catch (IOException ex) {
+                    throw Throwables.propagate(ex);
+                }
+            }
+            return TaskResult.empty(request);
```
komamitsu, line 144: When is this line executed?

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
```diff
@@ -160,8 +160,13 @@ protected void runWithHeartbeat(TaskRequest request)
...
-                        logger.debug("", ex);
+                        if (request.isCancelRequested()) {
+                            logger.warn("Task {} canceled.", request.getTaskName());
+                        }
+                        else {
+                            logger.error("Task {} failed.\n{}", request.getTaskName(), formatExceptionMessage(ex));
+                            logger.debug("", ex);
+                        }
                         // TODO use debug to log stacktrace here
                         callback.taskFailed(request, agentId, ex.getError(cf).get());  // TODO is error set?
```
komamitsu, line 171: What if `runWithWorkspace` throws `TaskExecutionException` not from `operator.cleanup(mergedRequest)`? It looks the clean up won't be called.

digdag-standards/src/main/java/io/digdag/standards/operator/PyOperatorFactory.java
szyn: This will never be executed. Since it expects to throw Exception, about to change the cleanup method to void.

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
szyn: Good point, I agree with you, I will change this message to a more general one.

szyn:
- [WorkflowExecutionTimeoutEnforcer#enforceAttemptTTLs()](https://github.com/treasure-data/digdag/blob/v0_10/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java#L149-L150)
- [WorkflowExecutionTimeoutEnforcer#enforceTaskTTLs()](https://github.com/treasure-data/digdag/blob/v0_10/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java#L179-L180)

Note that those two functions will change the status to CANCEL_REQUESTED if the task/attempt TTLs are over. 
As a result, there are no differences because those are called the same method.

digdag-standards/src/main/java/io/digdag/standards/operator/PyOperatorFactory.java
szyn: This line is executed when the state does not have "commandStatus".  e.g. Not a polling-based task. But now I feel it'd better change to void as return type.

digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java
szyn: If `runWithWorkspace` throws `TaskExecutionException`, it depends on what type of `TaskExecutionException`, but it will retry the task if it is called by `TaskExecutionException#ofNextPolling` which is used by a polling-based task. And then `cleanup` be executed at that time if `mergedResest.isCancelRequested()` is true. 
On the other hand, if cleanup failed to perform, target components will remain, I added a note about that and throw `TaskExecutionException` after cleanup to stop the task. https://github.com/treasure-data/digdag/pull/1480/commits/afb261043d8778f5c77f8d05736e73578b6a47c8

digdag-standards/src/main/java/io/digdag/standards/operator/PyOperatorFactory.java
szyn: Updated. https://github.com/treasure-data/digdag/pull/1480/commits/36655d268db2c1cd567d31400573f8bd8429421c

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
szyn: Updated: https://github.com/treasure-data/digdag/pull/1480/commits/be276de9707ed5181d516f60fd09b1dba76e02c0

digdag-core/src/test/java/io/digdag/core/agent/OperatorManagerTest.java
```diff
@@ -88,6 +90,21 @@ public void testRunWithHeartbeatWithSuccessTask()
...
+
+        OperatorManager om = spy(operatorManager);
+        Operator op = mock(Operator.class);
+        OperatorFactory of = mock(OperatorFactory.class);
+        doReturn(of).when(registry).get(any(), any());
+        doReturn(op).when(of).newOperator(any());
+        om.runWithHeartbeat(taskRequest);
+        verify(op, times(0)).run();
+        verify(op, times(1)).cleanup(any(TaskRequest.class));
+    }
```
komamitsu, line 106: Don't need to check if `taskFailed` is called?

komamitsu (approved):
👍

digdag-core/src/test/java/io/digdag/core/agent/OperatorManagerTest.java
szyn: I think it would be better to check, I'll add it. Thank you!

szyn: I also added other callback cases, just in case.

PR #1484, opened 2020-11-13
Enable to fall back to default configuration to support the common configuration for multiple ECS cluster

serihiro:
# What does this PR change?
Support falling back to the default configuration prefix for `EcsClientConfig` so that users can use the common configuration items for several ECS cluster names.

# Background

Currently, `access_key_id` and `secret_access_key` are expected to be stored into `systemConfig` with the following keys respectively.

- `agent.command_executor.ecs.#{cluster_name}.access_key_id`
- `agent.command_executor.ecs.#{cluster_name}.secret_access_key`

But in some use cases (e.g. using multiple clusters), we need to specify same `access_key_id` and `secret_access_key` for multiple ECS cluster names.

This PR enables it by falling back to default prefix if specified `cluster_name` does not match to those on `systemConfig`.

e.g. When there are the following keys in `systemConfig` and digdag  user specifies `cluster2` as a cluster name,

- `agent.command_executor.ecs.cluster1.access_key_id`
- `agent.command_executor.ecs.cluster1.secret_access_key`

the following items will be fetched instead.

- `agent.command_executor.ecs.__default_config__.access_key_id`
- `agent.command_executor.ecs.__default_config__.secret_access_key`

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfig.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfig.java
@@ -58,8 +58,10 @@ public class EcsClientConfig
         // - capacity_provider_name (optional/String)
         // - container_memory (optional/Integer)
         // - container_cpu (optional/Integer)
-        // - placementStrategyType (optional/String)
-        // - placementStrategyField (optional/String)
+        // - placement_strategy_type(optional/String)
+        // - placement_strategy_field (optional/String)
+        // - assign_public_ip (optional/boolean)
+        // - started_by (optional/String)
         // - task_cpu (optional/String) e.g. `1 vcpu` or `1024` (CPU unit)
         // - task_memory (optional/String) e.g. `1 GB` or `1024` (MiB)
         // For more detail of the value format of `task_cpu` and `task_memory`, please see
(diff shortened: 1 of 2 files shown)


digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfig.java
```diff
@@ -109,7 +120,7 @@ private static EcsClientConfig buildEcsClientConfig(String clusterName, Config e
                 .withCapacityProviderName(ecsConfig.getOptional("capacity_provider_name", String.class))
                 .withContainerCpu(ecsConfig.getOptional("container_cpu", Integer.class))
                 .withContainerMemory(ecsConfig.getOptional("container_memory", Integer.class))
-                .withStartedBy(ecsConfig.getOptional("startedBy", String.class))
+                .withStartedBy(ecsConfig.getOptional("started_by", String.class))
```
yoyama, line 123: Could you explain this change? May be for keeping naming rule.
Is there any compatibility issue (in our side)?

serihiro: Just typo

yoyama: Do you mean that this option is not used yet and no impact of this change ?

serihiro: Yes. This option is not yet used in our environment.

yoyama (approved):
LGTM

serihiro:
Thanks

PR #1489, opened 2020-12-01
Make CommandExecutor pluggable

szyn:
Follow-up #1315, make CommandExecutor pluggable using the same approach with #1325.
Note that if this design is OK, I will add documentation for this.

See also: https://github.com/treasure-data/digdag/pull/1315#issuecomment-578309601


szyn:
I will recreate this PR since it's difficult to rebase...

PR #1493, opened 2020-12-08
[Master to v0_10] Release note update

yoyama:
Add release notes and change version.
It is not 0.10.0 yet. It will be updated later.


komamitsu (approved):
👍

PR #1501, opened 2020-12-15
[Master to v0_10] digdag-ui 1499

yoyama:
Merge #1499 to v0_10


serihiro (approved):
I suppose this change will not affect to the behavior of digdag-ui because `ini` package is a `ini file` parser literally.

PR #1503, opened 2020-12-16
Set target_session_id and target_attempt_id when require op is called

hkdnet:
I want to know which attempt and which session is required by a `require` operator to understand my attempt dependency. Currently, digdag doesn't have such information. What I can is guessing it by using time.

With this patch, two values, `target_session_id` and `target_attempt_id`, are stored to state param of `require` operator. Then I can use them to find the session and the attempt that were required.

### DIFF
--- a/digdag-tests/src/test/java/acceptance/RequireIT.java
+++ b/digdag-tests/src/test/java/acceptance/RequireIT.java
@@ -451,7 +451,7 @@ public class RequireIT
     }
 
     @Test
-    public void testRequireHasTargetSessionIdEvenWhenJustWaiting()
+    public void testRequireHasTargetSessionIdEvenWhenChildIsNotKicked()
             throws Exception
     {
         // Create new project
--- a/digdag-tests/src/test/java/acceptance/RequireIT.java
+++ b/digdag-tests/src/test/java/acceptance/RequireIT.java
@@ -103,7 +103,7 @@ public class RequireIT
 
         // Wait for the attempt to complete
         boolean success = false;
-        for (int i = 0; i < 30; i++) {
+        for (int i = 0; i < 60; i++) {
             CommandStatus attemptsStatus = main("attempts",
                     "-c", config.toString(),
                     "-e", server.endpoint(),
@@ -503,7 +503,7 @@ public class RequireIT
 
         // Wait for the attempt to complete
         boolean success = false;
-        for (int i = 0; i < 30; i++) {
+        for (int i = 0; i < 60; i++) {
             CommandStatus attemptsStatus = main("attempts",
                     "-c", config.toString(),
                     "-e", server.endpoint(),
--- a/digdag-tests/src/test/java/acceptance/RequireIT.java
+++ b/digdag-tests/src/test/java/acceptance/RequireIT.java
@@ -3,6 +3,7 @@ package acceptance;
 import io.digdag.client.DigdagClient;
 import io.digdag.client.api.Id;
 import io.digdag.client.api.RestSessionAttemptCollection;
+import io.digdag.client.config.Config;
 
 import com.google.common.io.ByteStreams;
 import com.google.common.io.Resources;
@@ -30,6 +31,8 @@ import static java.util.Arrays.asList;
 import static org.junit.Assert.fail;
 import static utils.TestUtils.copyResource;
 import static utils.TestUtils.getAttemptId;
+import static utils.TestUtils.getSessionId;
+import static utils.TestUtils.getStateParams;
 import static utils.TestUtils.main;
 
 import static org.hamcrest.Matchers.containsString;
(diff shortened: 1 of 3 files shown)


yoyama:
@hkdnet Sorry for late reply. I would like to ask you to move the branch to `v0_11` if you need the fix for TD workflow.

hkdnet:
@yoyama Thanks for your release operation!
I switched the base branch to `v0_11` 👍

yoyama:
@hkdnet Thank you for your corporation.
I guess the implementation looks good.
But we need tests for it. Could you add some tests to confirm the parameters can be retrieved as expected.
We have integration test sub module in `digdaga-tests` and `RequireIT` is for require>.
You can add your test in it.

hkdnet:
@yoyama Thank you for your support. I didn't know the integration test.
I added test cases 👍

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -115,6 +118,27 @@ public void testRequire()
 
         // Verify that the file created by the child workflow is there
         assertThat(Files.exists(childOutFile), is(true));
+
+        CommandStatus attemptTasks = main("tasks",
+                "-c", config.toString(),
+                "-e", server.endpoint(),
+                String.valueOf(attemptId));
```
yoyama, line 125: May be `attemptId.toString()` or `attemptId.get()` is enough ?

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -115,6 +118,27 @@ public void testRequire()
...
+
+        // It has 2 tasks, +parent and +parent+require.
+        assertThat(stateParams.size(), is(2));
+        Config requireOperatorStateParams = stateParams.get(1);
+
+        Id targetAttemptId = requireOperatorStateParams.get("target_attempt_id", Id.class);
+        CommandStatus targetAttemptStatus = main("attempts",
+                "-c", config.toString(),
+                "-e", server.endpoint(),
+                String.valueOf(targetAttemptId));
```
yoyama, line 137: ditto

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -425,4 +449,85 @@ private void prepareForChildWF(Path childOutFile)
...
+                    "-e", server.endpoint(),
+                    "require", "parent",
+                    "--session", session);
+            assertThat(startStatus.code(), is(0));
+            attemptId = getAttemptId(startStatus);
+        }
+
+        // Wait for the attempt to complete
+        boolean success = false;
+        for (int i = 0; i < 30; i++) {
```
yoyama, line 506: Increase the maximum wait to 60 may be better. We sometimes faces timeout because of CI performance.

digdag-tests/src/test/java/acceptance/RequireIT.java
hkdnet: Agreed. This code follows the existing style.
https://github.com/treasure-data/digdag/blob/c9787276b8d3332da9ed3a87aaf330a222d44df9/digdag-tests/src/test/java/acceptance/RequireIT.java#L107

I'm going to fix my code.

digdag-tests/src/test/java/acceptance/RequireIT.java
hkdnet: Thanks. Updated the limits. https://github.com/hkdnet/digdag/commit/f00f7002c6049bb42f6cac2a64dfec59a0f4665d

digdag-tests/src/test/java/acceptance/RequireIT.java
hkdnet: Updated https://github.com/hkdnet/digdag/commit/19de2e155de25ddd34f1e45e75c40457567bac36

digdag-tests/src/test/java/acceptance/RequireIT.java
```diff
@@ -425,4 +449,85 @@ private void prepareForChildWF(Path childOutFile)
             Files.write(projectDir.resolve("child.dig"), child.getBytes("UTF-8"));
         }
     }
+
+    @Test
+    public void testRequireHasTargetSessionIdEvenWhenJustWaiting()
```
yoyama, line 454: I guess the purpose of this test is that we can get correct session_id and attempt_id when the session of the child already exists and require> will not kick by itself. So change the test name may be better.

In addition, this implementation does not assure parent will wait for child or not. Because the child workflow is very short and it is not certain whether the status of child is running or stopped when the parent session starts.
So 'waiting' is not correct in my understand.

hkdnet: > I guess the purpose of this test is that we can get correct session_id and attempt_id when the session of the child already exists and require> will not kick by itself

Yes 👍 

> 'waiting' is not correct in my understand

I misunderstood the behavior, sorry. The parent **doesn't wait** for the child unless the parent actually kicks the child.
I will rename the test name and delete assertions that are based on the assumption the child finished successfully.

hkdnet: Sorry, I misread your comment... 🙈 
The parent always waits for the child. It's independent of whether the parent kicks the child or not.
Will just rename the test case.

hkdnet: done: https://github.com/hkdnet/digdag/commit/1a136ba7c377fb64bbfd57333cda7216aa287894

yoyama (approved):
LGTM.

hkdnet:
@yoyama Thank you for your review!
Then, what is the next step? (not sure when to merge and who do it)

PR #1510, opened 2020-12-22
Change version to '0.10.0-SNAPSHOT'

yoyama:
Fix the version to `0.10.0-SNAPSHOT`


szyn (approved):

szyn:
Do we need to update this line as well? https://github.com/treasure-data/digdag/blob/v0_10/digdag-docs/src/conf.py#L28

yoyama:
@szyn Thank you for your pointing out. I will fix it the merge.

PR #1512, opened 2020-12-23
KubernetesCommandExecutor throw an exception if pod exceeds expire of TemporalConfigStorage

TrsNium:

reopened https://github.com/treasure-data/digdag/pull/1359


TrsNium:
Failed to follow the remote master, so I'll rebuild it with another PR.

PR #1513, opened 2020-12-23
Throw an error depending on the expiration date of TemporalConfigStorage

TrsNium:

reopened #1359


digdag-spi/src/main/java/io/digdag/spi/Storage.java
```diff
@@ -35,4 +35,8 @@ String put(String key, long contentLength,
     {
         return Optional.absent();
     }
+
+    Long getDirectDownloadExpiration();
+
+    Long getDirectUploadExpiration();
```
yoyama, line 41: To keep compatibility, it is better to change to default method.
In addition, I guess it is better that the return value is Optional<Long> like getDirectDownloadHandle(), getDirectUploadHandle().

digdag-standards/src/main/java/io/digdag/standards/command/KubernetesCommandExecutor.java
```diff
@@ -256,6 +258,21 @@ CommandStatus getCommandStatusFromKubernetes(final CommandContext context,
             log(logMessage, clog);
             nextExecutorState.set("log_offset", FACTORY.numberNode(offset + logMessage.length())); // update log_offset
         }
+        else if(isLaunchingLongerThanInConfigStorageExpiration(previousStatusJson)){
```
yoyama, line 261: spaces `if (..) {`

digdag-spi/src/main/java/io/digdag/spi/Storage.java
yoyama: In addition add comments on unit of return value is good. (seconds? or milliseconds?)

digdag-storage-gcs/src/main/java/io/digdag/storage/gcs/GCSStorage.java
```diff
@@ -165,6 +165,18 @@ public void list(String objectPrefix, FileListing callback)
         return Optional.of(DirectUploadHandle.of(url));
     }
 
+    @Override
+    public Long getDirectDownloadExpiration()
+    {
+        return config.get("direct_download_expiration", Long.class, 10L*60);
```
yoyama, line 171: Could you explain about default value `10L*60`?

digdag-storage-gcs/src/main/java/io/digdag/storage/gcs/GCSStorage.java
```diff
@@ -165,6 +165,18 @@ public void list(String objectPrefix, FileListing callback)
...
+    }
+
+    @Override
+    public Long getDirectUploadExpiration()
+    {
+        return config.get("direct_upload_expiration", Long.class, 10L*60);
```
yoyama, line 177: ditto

digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3Storage.java
```diff
@@ -224,6 +225,18 @@ public void list(String keyPrefix, FileListing callback)
         return Optional.of(DirectUploadHandle.of(url));
     }
 
+    @Override
+    public Long getDirectDownloadExpiration()
+    {
+        return config.get("direct_download_expiration", Long.class, 10L*60);
```
yoyama, line 231: Could you explain about default value `10L*60`?

digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3Storage.java
```diff
@@ -224,6 +225,18 @@ public void list(String keyPrefix, FileListing callback)
...
+    }
+
+    @Override
+    public Long getDirectUploadExpiration()
+    {
+        return config.get("direct_upload_expiration", Long.class, 10L*60);
```
yoyama, line 237: Could you explain about default value `10L*60`?

digdag-storage-gcs/src/main/java/io/digdag/storage/gcs/GCSStorage.java
TrsNium: https://github.com/treasure-data/digdag/blob/432517d0141959a7bc906790ce99233bc39544ce/digdag-storage-gcs/src/main/java/io/digdag/storage/gcs/GCSStorage.java#L159
@yoyama  Because it is just original implementation use `10L*60` as default value

digdag-spi/src/main/java/io/digdag/spi/Storage.java
TrsNium: thanks for your review.
Updat to use default method and use Optinal type.

yoyama:
@szyn Could you review this PR?
LGTM for me.

yoyama (approved):

digdag-standards/src/main/java/io/digdag/standards/command/KubernetesCommandExecutor.java
```diff
@@ -276,7 +293,7 @@ CommandStatus getCommandStatusFromKubernetes(final CommandContext context,
             final InputStream in = outConfigStorage.getContentInputStream(outputArchiveKey);
             ProjectArchives.extractTarArchive(context.getLocalProjectPath(), in); // runtime exception
         }
-        else if (defaultPodTTL.isPresent() && isRunningLongerThanTTL(previousStatusJson)) {
+        else if (isRunningLongerThanOutConfigStorageExpiration(previousStatusJson) || (defaultPodTTL.isPresent() && isRunningLongerThanTTL(previousStatusJson))) {
```
szyn, line 296: I guess you are using a non-blocking operator such as `py>`. In this case, IIRC, `(defaultPodTTL.isPresent() && isRunningLongerThanTTL(previousStatusJson))` will never become true. In addtion, this block will not execute when user/[WorkflowExecutionTimeoutEnforcer](https://github.com/treasure-data/digdag/blob/953036db19567ac24b1fc38e9fb802dd866c952e/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java#L179-L180) cancel-requested to the attempt. Is it OK/acceptable for you?  
We also had the same kind of issue. Thus we introduce the `cleanup` function to the Operator and CommnadExecutor interfaces and implements them in PyOperatorFactory and EcsCommandExecutor. Please see #1480, and this function helps to perform the cleanup process if the attempt received cancel-requested. I feel it is a good idea to handle it by using the same way to avoid such an issue. 
How about adding it to `KubernetesCommandExecutor` as well? How do you think about that?

PR #1517, opened 2021-01-05
Support adding tags on RunTaskRequest

serihiro:
# What does this PR change?
This PR enables users to specify `Tag` for `RunTaskRequest` on `EcsCommandExecutor`.

# Background
Amazon ECS supports `tag` for ecs tasks as arbitrary meta data of them. It is useful to manage/find ECS tasks on AWS console.

Expected use cases are
- Specifying project name
- Specifying notes

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
@@ -565,6 +565,7 @@ public class EcsCommandExecutor
         setEcsTaskOverride(commandContext, commandRequest, td, runTaskRequest, clientConfig); // RuntimeException,ConfigException
         setEcsTaskLaunchType(clientConfig, runTaskRequest);
         setEcsTaskStartedBy(clientConfig, runTaskRequest);
+        setEcsTaskTags(clientConfig, runTaskRequest);
         setEcsNetworkConfiguration(clientConfig, runTaskRequest);
         setCapacityProviderStrategy(clientConfig, runTaskRequest);
         setPlacementStrategy(clientConfig, runTaskRequest);
@@ -789,6 +790,13 @@ public class EcsCommandExecutor
     {
         if (clientConfig.getStartedBy().isPresent()) {
             runTaskRequest.setStartedBy(clientConfig.getStartedBy().get());
+        }
+    }
+
+    protected void setEcsTaskTags(EcsClientConfig clientConfig, RunTaskRequest runTaskRequest)
+    {
+        if (clientConfig.getTags().isPresent()) {
+            runTaskRequest.setTags(clientConfig.getTags().get());
         }
     }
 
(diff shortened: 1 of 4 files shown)


digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfigBuilder.java
```diff
@@ -130,6 +134,26 @@ public EcsClientConfigBuilder withTaskMemory(Optional<String> taskMemory)
         return this;
     }
 
+    public EcsClientConfigBuilder withTags(Optional<Config> tags)
+    {
+        if (tags.isPresent()) {
+            final Config tagConfig = tags.get();
+            final List<Tag> tagList = new ArrayList<>();
```
komamitsu, line 141: [minor] You can also use Stream API like
```
    this.tags = Optional.of(
        tagConfig.getKeys().stream()
            .map(key -> new Tag().withKey(key).withValue(tagConfig.get(key, String.class)))
            .collect(Collectors.toList()));
```

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -566,6 +565,7 @@ RunTaskRequest buildRunTaskRequest(
         setEcsTaskOverride(commandContext, commandRequest, td, runTaskRequest, clientConfig); // RuntimeException,ConfigException
         setEcsTaskLaunchType(clientConfig, runTaskRequest);
         setEcsTaskStartedBy(clientConfig, runTaskRequest);
+        setEcsTaskTags(clientConfig, runTaskRequest);
```
komamitsu, line 568: We'd better write a test to see if `RunTaskRequest.tags` is expectedly set?

digdag-standards/src/main/java/io/digdag/standards/command/ecs/EcsClientConfigBuilder.java
serihiro: I refactored here 90a1fff

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
serihiro: @komamitsu 
Actually it is difficult to test `setEcsXXX` methods themselves because they are protected methods and `runTaskRequest` is a just local variable. So it is difficult to evaluate them.

It is the reason why I tested the setup configuration logic at `EcsClientConfigTest`. https://github.com/treasure-data/digdag/pull/1517/files#diff-c0efef3822af1f3097d11b328b9c7f1044e8bf176b73791a02e819c19979fcdeR126-R161

If you have a good idea about this testing, could you tell me it?

komamitsu: I think adding a test in `io.digdag.standards.command.EcsCommandExecutorTest` using some mocks is an option. We may be able to add some tests for this change later.

komamitsu (approved):

serihiro:
Thanks!

PR #1518, opened 2021-01-08
Set `started_at` in group tasks as well

komamitsu:
We sometimes want to know the stats of task execution delays and durations in Digdag. `tasks.started_at` column is important to know a delay between previous task and target task, but group task has no value now.

```
=> select sum(case when started_at is not null then 1 else 0 end) as count_started_at, sum(case when started_at is null then 1 else 0 end) as count_null_started_at from tasks where task_type = 1;
 count_started_at | count_null_started_at 
------------------+-----------------------
                0 |                  3415
(1 row)

=> select sum(case when started_at is not null then 1 else 0 end) as count_started_at, sum(case when started_at is null then 1 else 0 end) as count_null_started_at from tasks where task_type = 1;
 count_started_at | count_null_started_at 
------------------+-----------------------
                0 |                  3412
(1 row)

  :

=> select sum(case when started_at is not null then 1 else 0 end) as count_started_at, sum(case when started_at is null then 1 else 0 end) as count_null_started_at from tasks where task_type = 1;
 count_started_at | count_null_started_at 
------------------+-----------------------
                0 |                  3414
(1 row)
```

So the field should be set even in group tasks as well.

```
$ git grep started_at
digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java:                    " and started_at < :startedBefore" +
digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java:                " set started_at = coalesce(started_at, now()), updated_at = now(), state = :newState" +
digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java:                .startedAt(getOptionalTimestampInstant(r, "started_at"))
digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java:                .startedAt(getOptionalTimestampInstant(r, "started_at"))
digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20161110112233_AddStartedAtColumnAndIndexToTasks.java:                    " add column started_at timestamp with time zone");
digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20161110112233_AddStartedAtColumnAndIndexToTasks.java:                    " add column started_at timestamp");
digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20161110112233_AddStartedAtColumnAndIndexToTasks.java:            handle.update("create index tasks_on_state_and_started_at on tasks (state, started_at, id asc) where started_at is not null");
digdag-core/src/main/java/io/digdag/core/database/migrate/Migration_20161110112233_AddStartedAtColumnAndIndexToTasks.java:            handle.update("create index tasks_on_state_and_started_at on tasks (state, started_at, id asc)");
```

`tasks.started_at` is checked only by io.digdag.core.database.DatabaseSessionStoreManager#findTasksStartedBeforeWithState to collect expired tasks https://github.com/treasure-data/digdag/blob/97cea9ff6526ea1ec3d88c5a25039ae8c12e8656/digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java#L167-L173.

So just setting `tasks.started_at` in group tasks makes `WorkflowExecutionTimeoutEnforcer.findTasksStartedBeforeWithState` collect long running root and group tasks as well. As a result, WorkflowExecutionTimeoutEnforcer will unexpectedly stop an attempt even with 30 sequential 1 hour tasks (this expire shouldn't happen since each task doesn't exceed 24 hours TTL). So, we also need to make WorkflowExecutionTimeoutEnforcer.findTasksStartedBeforeWithState see only non-group tasks.

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -1186,7 +1186,7 @@ public class DatabaseSessionStoreManager
                     " else " + TaskStateCode.READY_CODE +
                     " end, " +
                     " started_at = case" +
-                    // Update `started_at` only when it's a group task and the column is null.
+                    // Update `started_at` only when it's a group task and the column is null, in other words "not started group task."
                     // Non group tasks' `started_at` are updated by TaskControlStore.setStartedState.
                     " when task_type = " + TaskType.GROUPING_ONLY + " then coalesce(started_at, now()) else started_at end" +
                     " where state = " + TaskStateCode.BLOCKED_CODE +
--- a/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
+++ b/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
@@ -1184,7 +1184,11 @@ public class DatabaseSessionStoreManager
                     " when task_type = " + TaskType.GROUPING_ONLY + " then " + TaskStateCode.PLANNED_CODE +
                     " when " + bitAnd("state_flags", Integer.toString(TaskStateFlags.CANCEL_REQUESTED)) + " != 0 then " + TaskStateCode.CANCELED_CODE +
                     " else " + TaskStateCode.READY_CODE +
-                    " end" +
+                    " end, " +
+                    " started_at = case" +
+                    // Update `started_at` only when it's a group task and the column is null.
+                    // Non group tasks' `started_at` are updated by TaskControlStore.setStartedState.
+                    " when task_type = " + TaskType.GROUPING_ONLY + " then coalesce(started_at, now()) else started_at end" +
                     " where state = " + TaskStateCode.BLOCKED_CODE +
                     " and parent_id = :parentId" +
                     " and exists (" +
@@ -1550,6 +1554,9 @@ public class DatabaseSessionStoreManager
         public <T> T insertRootTask(long attemptId, Task task, SessionBuilderAction<T> func)
         {
             long taskId = dao.insertTask(attemptId, task.getParentId().orNull(), task.getTaskType().get(), task.getState().get(), task.getStateFlags().get());  // tasks table don't have unique index
+            // Root task's `started_at` isn't updated by trySetChildrenBlockedToReadyOrShortCircuitPlannedOrCanceled().
+            // So this line is just for setting `started_at` of root task although there may be room for optimization.
+            dao.setStartedState(taskId, task.getState().get(), task.getState().get());
             dao.insertTaskDetails(taskId, task.getFullName(), task.getConfig().getLocal(), task.getConfig().getExport());
             dao.insertEmptyTaskStateDetails(taskId);
             return func.call(new DatabaseTaskControlStore(handle), taskId);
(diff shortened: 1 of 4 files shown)


digdag-server/src/main/java/io/digdag/server/WorkflowExecutionTimeoutEnforcer.java
```diff
@@ -46,11 +46,14 @@
...
-    // READY is excluded the workflow itself is working correctly and number of threads is insufficient.
+    //
+    // - BLOCKED and PLANNED are excluded because there're other running tasks that should be enforced instead.
+    // - READY is excluded the workflow itself is working correctly and number of threads is insufficient.
+    // - GROUP_RETRY_WAITING is excluded because this status is supposed to be immediately changed to PLANNED
+    //   and long running group tasks with this state could happen only due to WorkflowExecutor trouble
+    //   not due to workflow definitions.
     private static final TaskStateCode[] TASK_TTL_ENFORCED_STATE_CODES = new TaskStateCode[] {
         TaskStateCode.RETRY_WAITING,
-        TaskStateCode.GROUP_RETRY_WAITING,
```
komamitsu, line 53: GROUP_RETRY_WAITING should be excluded in this change since if a group task has sequential 30 children tasks which takes 1 hour and the group task is in GROUP_RETRY_WAITING state, the group task will be regarded as an expired task.

Excluding this state should be okay because in current implementation group tasks don't have `started_at` and they're ignored by `WorkflowExecutionTimeoutEnforcer`. In a sense this state in this list hasn't made sense.

Also, non-group task won't be this state.

https://github.com/treasure-data/digdag/blob/432517d0141959a7bc906790ce99233bc39544ce/digdag-core/src/main/java/io/digdag/core/workflow/WorkflowExecutor.java#L712-L722

https://github.com/treasure-data/digdag/blob/432517d0141959a7bc906790ce99233bc39544ce/digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java#L1036-L1045

digdag-core/src/test/java/io/digdag/core/workflow/WorkflowExecutorTest.java
```diff
@@ -434,6 +438,61 @@ public void randomEnqueue()
...
+        Instant rootTaskStartedAt;
+        {
+            // Root task
+            StoredTask task = tasks.get(0);
+            assertEquals("+nested", task.getFullName());
+            assertTrue(task.getTaskType().isGroupingOnly());
+            assertEquals(TaskStateCode.SUCCESS, task.getState());
+            assertThat(task.getUpdatedAt(), is(greaterThanOrEqualTo(startTime)));
+            assertThat(task.getUpdatedAt(), is(lessThanOrEqualTo(finishTime)));
+            assertThat(task.getStartedAt().get(), is(greaterThanOrEqualTo(startTime)));
```
serihiro, line 462: [IMO] IIUC, the purpose of this assertion is to confirm that a group taks' `started_at` was updated (sorry if I mis-understood about it). So I think it would be better to leave the comment to describe the purpose for better clarification.

digdag-core/src/main/java/io/digdag/core/database/DatabaseSessionStoreManager.java
```diff
@@ -1184,7 +1184,11 @@ public int trySetChildrenBlockedToReadyOrShortCircuitPlannedOrCanceled(long task
                     " when task_type = " + TaskType.GROUPING_ONLY + " then " + TaskStateCode.PLANNED_CODE +
                     " when " + bitAnd("state_flags", Integer.toString(TaskStateFlags.CANCEL_REQUESTED)) + " != 0 then " + TaskStateCode.CANCELED_CODE +
                     " else " + TaskStateCode.READY_CODE +
-                    " end" +
+                    " end, " +
+                    " started_at = case" +
+                    // Update `started_at` only when it's a group task and the column is null.
+                    // Non group tasks' `started_at` are updated by TaskControlStore.setStartedState.
+                    " when task_type = " + TaskType.GROUPING_ONLY + " then coalesce(started_at, now()) else started_at end" +
```
serihiro, line 1191: [IMO] It would be better if you left a note about the cases that `started_at` is NULL and not NULL in this comment.

serihiro (approved):
I left minor comments but this change itself looks good for me.

PR #1520, opened 2021-01-12
Add `wait` operator by porting https://github.com/yoyama/digdag-wait-op

komamitsu:
The developers and users of Digdag often want to stop a workflow for a specific duration to emulate a long running task. `sh>: sleep 5` may be a workaround, but it's a blocking operator and doesn't fit when we want to emulate long running non-blocking operator task.

This pull request ports https://github.com/yoyama/digdag-wait-op with some minor changes to address the above pain point.


komamitsu:
Let me add `poll_interval` option for development use case.

digdag-standards/src/main/java/io/digdag/standards/operator/WaitOperatorFactory.java
```diff
@@ -0,0 +1,155 @@
...
+
+        private Optional<Duration> pollInterval(Config config)
+        {
+            Duration pollInterval;
+            try {
+                Optional<String> pollIntervalStr = config.getOptional("poll_interval", String.class);
+                if (!pollIntervalStr.isPresent()) {
+                    return Optional.absent();
+                }
+                pollInterval = Durations.parseDuration(pollIntervalStr.get());
```
yoyama, line 92: I guess we may make the codes simple by using Optional.transform().
But not strong opinion.

```
       try {
            Optional<Duration> pollInterval = config
                    .getOptional("poll_interval", String.class)
                    .transform(Durations::parseDuration);
            if (pollInterval.isPresent()) {
                logger.debug("wait poll_interval: {}", pollInterval.get());
            }
            return pollInterval;
        }
        catch (RuntimeException re) {
            throw new ConfigException("Invalid configuration", re);
        }
```

digdag-tests/src/test/java/acceptance/WaitIT.java
```diff
@@ -0,0 +1,120 @@
...
+        {
+            copyResource(targetResourcePath, root().resolve("wait.dig"));
+            ExecResult result = runAndMonitorDuration(() ->
+                    main("run", "-o", root().toString(), "--project", root().toString(), "wait.dig"));
+            CommandStatus status = result.commandStatus;
+            assertThat(status.errUtf8(), status.code(), is(0));
+            assertThat(result.duration, greaterThan(baselineDuration));
+            assertThat(result.duration, lessThan(
+                    // Actual wait duration can be longer than the specified 10 seconds for some reason
+                    baselineDuration.plusSeconds((long) (expectedDuration * 1.5))));
```
yoyama, line 75: I have a bit concern in this condition.
When CI performance is degraded, it may fail unexpectedly.

yoyama (approved):
I left some comments.
But LGTM. 👍

digdag-tests/src/test/java/acceptance/WaitIT.java
komamitsu: Agreed. Just tweaked the value since I don't think there is an easy robust way to handle the concern.

komamitsu:
Thanks!

PR #1522, opened 2021-01-20
Fix documentation title.

yoyama:
Currently documentation title is generated from a latest tag and it is wrong if we set tags for other purpose (backup and so on).
Current title is `... Digdag 0_10_before_merge_v0_9_38 documentation`
As a workaround, I set fixed title in `conf.py` .
It require additional work in the release, but the release does not happen so often.

### DIFF
--- a/digdag-docs/src/conf.py
+++ b/digdag-docs/src/conf.py
@@ -69,3 +69,5 @@ html_extra_path = ['_extra']
 html_context = {
   'extra_css_files': ['_static/custom.css']
 }
+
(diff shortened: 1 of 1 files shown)


szyn (approved):
As a workaround, I think this change is fine. It might be a good idea to leave a TODO comment that fixes this titile properly when we release this version officially so that we don't forget.

yoyama:
I found the annotation tag is only used for document title by `git describe`.
So we can avoid this issue by restricting to use annotation tag only for release.
So I close this.

PR #1531, opened 2021-02-02
Use Application Default Credential if gcp.credential is absent

rhase:
# What this PR changes

Use [Application Default Credential (ADC)](https://cloud.google.com/docs/authentication/production#automatically) if gcp.credential secret is not defined.  
Applied to bq, bq_ddl, bq_load, bq_extract and gcs_wait.

# Motivation

It's very convenient to use ADC when you run digdag on gce or gke because you don't have to manage service account credentials by your self (Google will manage them for you).  
You can also test workflow on your machine with your google account by `gcloud auth application-default login`. You don't need service account and gcp.credential secret.


yoyama:
Thank you for your contribution.
It seems good. But could you create unit test for the class ?

rhase:
@yoyama 
Added tests.
Could you review it?

szyn:
@rhase Thank you for adding test cases. Looks good to me 👍 BTW, could you update the target branch to `master`?  We want to merge into master as for this PR. 
Please let me know if you would like me to do it.

rhase:
I tried to change target branch by "Edit" button of this PR, but that seems to have caused unexpected effect...
I guess I should recreate PR to master branch.

rhase:
recreated as #1585

PR #1532, opened 2021-02-09
Add some timeout configurations to `mail` operators

komamitsu:
Current `mail` operator's timeout configurations are hard-coded and the values seems a bit too short in production environment.

This PR does
- make it possible to configure `mail` operator's connection timeout and socket timeout
- increase the default timeout values

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/MailOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/MailOperatorFactory.java
@@ -46,8 +46,8 @@ public class MailOperatorFactory
 
     private static final String CONFIG_KEY_CONNECT_TIMEOUT = "connect_timeout";
     private static final String CONFIG_KEY_SOCKET_TIMEOUT = "socket_timeout";
-    private static final Duration DEFAULT_CONNECT_TIMEOUT = Duration.ofSeconds(60);
-    private static final Duration DEFAULT_SOCKET_TIMEOUT = Duration.ofSeconds(180);
+    private static final DurationParam DEFAULT_CONNECT_TIMEOUT = DurationParam.of(Duration.ofSeconds(60));
+    private static final DurationParam DEFAULT_SOCKET_TIMEOUT = DurationParam.of(Duration.ofSeconds(180));
 
     private final TemplateEngine templateEngine;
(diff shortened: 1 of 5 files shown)


digdag-standards/src/main/java/io/digdag/standards/operator/MailOperatorFactory.java
```diff
@@ -288,12 +311,16 @@ private InternetAddress newAddress(String str)
...
+                // TODO: This code expects `params` has `port` field even if `secrets` has the field.
+                //       Maybe we need to revisit here later to see whether this behavior is reasonable or not.
                 .port(secrets.getSecretOptional("port").transform(Integer::parseInt).or(params.get("port", int.class)))
                 .startTls(secrets.getSecretOptional("tls").transform(Boolean::parseBoolean).or(params.get("tls", boolean.class, true)))
                 .ssl(secrets.getSecretOptional("ssl").transform(Boolean::parseBoolean).or(params.get("ssl", boolean.class, false)))
                 .debug(params.get("debug", boolean.class, false))
                 .username(secrets.getSecretOptional("username").or(params.getOptional("username", String.class)))
                 .password(secrets.getSecretOptional("password"))
+                .connectTimeout(params.get(CONFIG_KEY_CONNECT_TIMEOUT, DurationParam.class, DEFAULT_CONNECT_TIMEOUT))
+                .socketTimeout(params.get(CONFIG_KEY_SOCKET_TIMEOUT, DurationParam.class, DEFAULT_SOCKET_TIMEOUT))
```
yoyama, line 323: It seems mail> support most of parameters for smtp connection in secrets.
So it may be better to follow the manner.
But I guess we don't need to store the parameters excepts `password` in secrets. 
So the current implementation is reasonable.

yoyama (approved):
👍

serihiro (approved):

digdag-standards/src/main/java/io/digdag/standards/operator/MailOperatorFactory.java
komamitsu: Yeah, it's what I thought as well and I wanted to go with a simpler way to reduce the complexity.

komamitsu:
Thanks!

PR #1533, opened 2021-02-15
Add `digdag profiler` subcommand

komamitsu:
Digdag users and developers sometimes want to know how much the start time of tasks delayed from the preceding task. This PR introduces `digdag profile` cli subcommand that analyzes historical `task_archives` table for that.

This subcommand does the followings:
- Analyze multiple attempts' tasks
- Collect and summarize some metrics including `task start delay` that is the duration between when the task was ready and when the task actually started 
- Skip task start delay analysis for some cases
  - See https://github.com/treasure-data/digdag/pull/1533/files#diff-b31c79802ceb314691a6c73bf0a541807295ea5d3b5f924c73c0d796f1b4734dR32-R59

The implementation in this PR just outputs overall stats. But we have the following plans:
- Task stats per site
- Attempt stats (overall / per site)


komamitsu:
Noticed `failure-alert` and `error` tasks are dynamically inserted into the root task and causes meaningless very long delays. We need to take care of this case…

digdag-docs/src/profiler.md
```diff
@@ -0,0 +1,89 @@
...
+Current profiler requires Digdag source code as follows
+
+```
+$ git clone https://github.com/treasure-data/digdag.git
+$ cd digdag
+$ ./gradlew digdag-profiler:installDist
+$ digdag-profiler/build/install/digdag-profiler/bin/digdag-profiler -c $HOME/.config/digdag/digdag.properties --from 2020-12-30T00:00:00
+{"attempts":2048,"totalTasks":8192,"totalRunTasks":8190,"totalSuccessTasks":8189,"totalErrorTasks":1,"mostDelayedTask":{"subtaskConfig":{},"exportParams":{},"resetStoreParams":[],"storeParams":{},"report":{"inputs":[],"outputs":[]},"error":{},"resumingTaskId":null,"id":369,"attemptId":1024,"upstreams":[],"updatedAt":"2021-02-16T08:56:01Z","retryAt":null,"startedAt":"2021-02-16T08:56:00Z","stateParams":{},"retryCount":0,"parentId":1000,"fullName":"+echo^failure-alert","config":{"local":{"_type":"notify","_command":"Workflow session attempt failed"},"export":{}},"taskType":0,"state":"success","stateFlags":0},"startDelayMillis":{"min":0,"max":2000,"average":67,"stddev":286},"execDurationMills":{"min":0,"max":11000,"average":1115,"stddev":1546}}
+```
+
```
yoyama, line 17: Is thre any reason, you don't integrate `profiler` to `digdag` execute binary ?
If we can use the profiler as `digdag profiler <option>`, it may be convenient and natural for digdag manner.

digdag-profiler/src/main/java/io/digdag/profiler/ConfigElementLoader.java
```diff
@@ -0,0 +1,62 @@
...
+import io.digdag.core.config.PropertyUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Properties;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+
+public class ConfigElementLoader
```
yoyama, line 16: Not strong opinion, but I guess this class can be available not only profiler but also other module.
So move to digdag-client may be better.

digdag-docs/src/profiler.md
komamitsu: The reasons why I created this tool separately are:
1. this tool is not for normal Digdag users, but for kind of Digdag developers
2. it's a bit uncommon since it accesses the database directly not thought REST API
3. I'm sure the existing all-in-one monolithic executable cli is the best way

But, on second thought, probably #2 doesn't matter since `digdag migrate` already exists. And #1 and #3 may be a matter of taste. So I think we can move this tool into the existing executable cil. Do you think we'd better merge this functionality to the existing cli?

komamitsu:
> Noticed failure-alert and error tasks are dynamically inserted into the root task and causes meaningless very long delays. We need to take care of this case…

Similar issue can happen in group retry. I think skipping attempts that has any failed task would be an option for the first step.

digdag-docs/src/profiler.md
komamitsu: As we talked later, I agree with your idea. I'll update this PR. Thanks!

digdag-profiler/src/main/java/io/digdag/profiler/ConfigElementLoader.java
komamitsu: ConfigElementLoader was created only to call `io.digdag.cli.Command#loadSystemProperties` and 
we'll move digdag-profiler to digdag-cli as `Profile` class which inherits `Command` class, so we don't need this class anymore.

komamitsu:
@yoyama I updated this PR. Can I ask you to take a look at it again? Thanks

digdag-cli/src/main/java/io/digdag/cli/Profile.java
```diff
@@ -0,0 +1,127 @@
...
+import static io.digdag.client.DigdagClient.objectMapper;
+
+
+public class Profile
+    extends Command
+{
+    @Parameter(
+            names = {"--from"},
+            required = true,
+            converter = TimestampConverter.class
```
yoyama, line 33: It might be helpful if `description` is defined with timestamp format.

digdag-cli/src/main/java/io/digdag/cli/Profile.java
```diff
@@ -0,0 +1,127 @@
...
+    )
+    Instant timeRangeFrom;
+
+    @Parameter(
+            names = {"--to"},
+            converter = TimestampConverter.class
```
yoyama, line 39: ditto.

digdag-cli/src/main/java/io/digdag/cli/profile/TaskAnalyzer.java
```diff
@@ -0,0 +1,97 @@
...
+        AtomicLong lastId = new AtomicLong();
+        TasksSummary.Builder tasksSummaryBuilder = new TasksSummary.Builder();
+        logger.debug("Task analysis started");
+
+        while (true) {
+            logger.debug("Collecting attempts");
+
+            // Paginate to avoid too long transaction
+            AtomicLong counter = new AtomicLong();
+            Map<Long, List<StoredSessionAttemptWithSession>> partitionedAttemptIds = tm.begin(() -> {
```
yoyama, line 57: I guess the context of this logic is
- Need finished attempts for some duration (createFrom and createTo)
- sm.findFinishedAttemptsWithSessions() return list of attempts
- partitioning the list and set wait to avoid too much stress to DB

IIUC, if sm.findFinishedAttemptsWithSessions return not List but **ResultIterator**, we don't need to do partitioning? (But Digdag not use the interface and I am not sure it is good)
http://jdbi.org/jdbi2/apidocs/org/skife/jdbi/v2/ResultIterator.html

And while running iteration, we can set small wait to avoid stress to DB.
In this case, we can expect to reduce memory consumption of JVM.
In addition, turning parameter 'partition' is a bit difficult to choose proper value and if we don't need it, it may be better.

But I don't have strong opinion, just my thought. 🙂

digdag-cli/src/main/java/io/digdag/cli/Profile.java
komamitsu: Ah, unfortunately digdag-cli doesn't use this field at all and outputs the descriptions of options like https://github.com/treasure-data/digdag/pull/1533/files#diff-cee923b77be1e7ada3d52d17f3d385a391e14e6870f9328a1d235b9f5d44ba9eR120-R124.

I noticed other 3 options have the meaningless field and I'll remove them.

digdag-cli/src/main/java/io/digdag/cli/profile/TaskAnalyzer.java
komamitsu: The option sounds cool and simple!

But the description of the class sounds that RDBMS transaction is kept open until the last entry is fetched to me. If so, the long transaction acquires shared locks and it potentially blocks write operation long time. The tuning parameters are actually not easy to understand, but I think few users are interested in them 😉

komamitsu: Anyway, the feedback that the implementation of partitioning seems too much is really important. Thank you.

I just updated it https://github.com/treasure-data/digdag/pull/1533/commits/f1f735f35a15d4e1427180e29a76a82bbdf30401

digdag-cli/src/main/java/io/digdag/cli/profile/TaskAnalyzer.java
```diff
@@ -0,0 +1,103 @@
...
+                break;
+            }
+
+            // Process partitioned attemptIds list from the beginning
+            for (List<AttemptIdWithSiteId> attemptIds : attemptIdsList) {
+                logger.debug("Processing {} attempts", attemptIds.size());
+                tm.begin(() -> {
+                    for (AttemptIdWithSiteId attemptIdWithSiteId : attemptIds) {
+                        List<ArchivedTask> tasks = sm.getSessionStore(attemptIdWithSiteId.siteId).getTasksOfAttempt(attemptIdWithSiteId.attemptId);
+                        TasksSummary.updateBuilderWithTasks(tasksSummaryBuilder, tasks);
```
yoyama, line 90: Is there any reason that `updateBuilderWithTasks` is static method of `TaskSummary` ?
I guess this is  a job for `TaskSummaryBuilder`. So  `TaskSummaryBuilder.updateWithTasks(tasks)` may be better.
But not strong opinion.

komamitsu: Good point! Fixed it

yoyama (approved):
LGTM. 👍

komamitsu:
Thanks!

PR #1548, opened 2021-03-16
Publish local

yoyama:
Add gradle tasks `releaseLocal`, `releaseLocalSnapshot` to publish to maven local repository.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -462,12 +462,10 @@ task release() {
 
 // Publish to ~/.m2
 task releaseLocal() {
-    // This is necessary to upload the executable jar to bintray as well.
-    // It makes to run _bintrayRecordingCopy task after cli task.
     release.mustRunAfter cli
     dependsOn subprojects.publishToMavenLocal
     doLast {
-        println "Released ${project.version}."
+        println "Released ${project.version} to local repository."
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -460,6 +460,18 @@ task release() {
     }
 }
 
+// Publish to ~/.m2
+task releaseLocal() {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -460,6 +460,18 @@ task release() {
...
 }
 
+// Publish to ~/.m2
+task releaseLocal() {
+    // This is necessary to upload the executable jar to bintray as well.
+    // It makes to run _bintrayRecordingCopy task after cli task.
+    release.mustRunAfter cli
+    dependsOn subprojects.publishToMavenLocal
+    doLast {
+        println "Released ${project.version}."
```
komamitsu, line 470: How about adding `to local repository` or something?

build.gradle
```diff
@@ -460,6 +460,18 @@ task release() {
     }
 }
 
+// Publish to ~/.m2
+task releaseLocal() {
+    // This is necessary to upload the executable jar to bintray as well.
+    // It makes to run _bintrayRecordingCopy task after cli task.
```
komamitsu, line 466: What happens if `_bintrayRecordingCopy` isn't called?

komamitsu (approved):
👍

build.gradle
yoyama: Good point. I will update it.

build.gradle
yoyama: `mustRunAfter` forces to run `cli` before the task run.
This is required when upload all-in-one jar to Bintray. 
But `releaseLocal` does not upload it to Bintray.
So I will remove the comment to avoid misleading.
We don't need cli binary itself for our purpose and no copy to local repository.
But keep consistency with `release`, I will keep `mustRunAfter`.
Thanks!

PR #1558, opened 2021-04-13
Change log level

myui:
Change logging level for problem investigation.

Note: fallback to local docker command executor should not generally be happened.


yoyama:
I think debug message is logged and log level change has no impact.

PR #1557, opened 2021-04-12
Fix another broken test

komamitsu:
We noticed `acceptance.td.TdDdlIT#testRetries` was still broken after merging the previous fix. This PR does
- Fix the broken test
- Test a topic branch on CircleCI to detect CircleCI test failure in advance

PR #1559, opened 2021-04-16
Restore skipped `td` operator integration tests

komamitsu:
We've commented out some verifications in `td` operator's integration tests that accessed actual Treasure Data endpoint and recently started to fail due to HTTP access decommission https://github.com/treasure-data/digdag/pull/1553 https://github.com/treasure-data/digdag/pull/1557. Those verification internally depended on LittleProxy to see what TD APIs are called, but they didn't work via HTTPS since LittleProxy doesn't support request capture with HTTPS.

This PR introduces `utils.TestUtils.RecordableWorkflow` that inject a custom TDClient recording all API calls instead of depending on LittleProxy and replaces the commented out verifications with it.

### DIFF
--- a/digdag-cli/src/main/java/io/digdag/cli/Run.java
+++ b/digdag-cli/src/main/java/io/digdag/cli/Run.java
@@ -222,6 +222,25 @@ public class Run
 
     private static final List<Long> USE_ALL = null;
 
+    protected DigdagEmbed.Bootstrap setupBootstrap(Properties systemProps)
+    {
+        return new DigdagEmbed.Bootstrap()
+                .setEnvironment(env)
+                .setSystemConfig(PropertyUtils.toConfigElement(systemProps))
+                .setSystemPlugins(loadSystemPlugins(systemProps))
+                .addModules(binder -> {
+                    Multibinder.newSetBinder(binder, SecretStore.class);
+                    binder.bind(SecretStoreManager.class).to(LocalSecretStoreManager.class).in(Scopes.SINGLETON);
+                    binder.bind(ResumeStateManager.class).in(Scopes.SINGLETON);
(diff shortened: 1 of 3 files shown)


digdag-cli/src/main/java/io/digdag/cli/Run.java
```diff
@@ -222,21 +222,9 @@ public SystemExitException usage(String error)
 
     private static final List<Long> USE_ALL = null;
 
-    public void run(String workflowNameArg, String matchPattern)
```
komamitsu, line 225: This change is just extracting some part of `run` method to a method so that we can customize the bootstrap from test code.

digdag-tests/src/test/java/utils/TestUtils.java
```diff
@@ -171,12 +185,30 @@ public static CommandStatus main(LocalVersion localVersion, InputStream in, Coll
...
+            InputStream in,
+            RecordableWorkflow.CustomMain.CommandAdder commandAdder)
     {
         try (
                 PrintStream outp = new PrintStream(out, true, "UTF-8");
                 PrintStream errp = new PrintStream(err, true, "UTF-8");
         ) {
-            Main main = new Main(localVersion.getVersion(), env, outp, errp, in);
+            Main main;
+            if (commandAdder != null) {
```
yoyama, line 206: Trivial thing, but it it is better to use Optional. (No strong opinion)

yoyama (approved):
GJ 👍

digdag-tests/src/test/java/utils/TestUtils.java
komamitsu: Ah, I'm used to Kotlin recently and tend to use just a nullable value even in Java. Also, Optional type value in Java isn't null safe, so I may sometimes simply use just nullable value 😄

komamitsu:
Thanks!

PR #1561, opened 2021-04-19
`pg` operator should check if the status table exists in advance before trying to create the table

komamitsu:
`pg` operator records the status of workflow tasks in the target PostgreSQL database (a.k.a. `status table`) so that it achieves exactly-once semantics. But this way requires some privileges including CREATE TABLE. For the case DBA doesn't want to give those permissions to the workflow, `pg` operator also provides an option to prepare `status table`  in advance.

We recently noticed it doesn't work expectedly since `pg` operator still issues `CREATE TABLE IF NOT EXISTS` that requires CREATE TABLE privilege and fails even when `status table` is already created in advance.

This PR makes `pg` operator check if `status table` exists in advance before trying to create the table to avoid this issue.

### DIFF
--- a/digdag-standards/src/test/java/io/digdag/standards/operator/pg/PgConnectionTest.java
+++ b/digdag-standards/src/test/java/io/digdag/standards/operator/pg/PgConnectionTest.java
@@ -26,10 +26,10 @@ import static org.junit.Assert.*;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Mockito.doNothing;
-import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.times;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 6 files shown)


digdag-tests/src/test/java/acceptance/td/RedshiftIT.java
```diff
@@ -58,8 +58,9 @@
     private static final String REDSHIFT_CONFIG = System.getenv("REDSHIFT_IT_CONFIG");
     private static final String SRC_TABLE = "src_tbl";
     private static final String DEST_TABLE = "dest_tbl";
-    private static final String DATA_SECHEMA = "data_schema";
-    private static final String STATUS_TABLE_SECHEMA = "status_table_schema";
+    private static final String DATA_SCHEMA = "data_schema";
```
komamitsu, line 61: All the changes in this file are just for refactoring and keeping consistency with PgIT

digdag-standards/src/test/java/io/digdag/standards/operator/pg/PgConnectionTest.java
```diff
@@ -111,9 +116,10 @@ public void executeUpdate()
 
     @Test
     public void txHelperPrepare()
-            throws SQLException
+            throws SQLException, NotReadOnlyException
     {
         TransactionHelper txHelper = pgConnection.getStrictTransactionHelper(null, "__digdag_status", Duration.ofDays(1));
+        doThrow(DatabaseException.class).when(pgConnection).executeReadOnlyQuery(eq("SELECT count(*) FROM \"__digdag_status\""), any());
```
serihiro, line 122: It would be better if you add an explanation of this mocking. IIUC, this mocking is to simulate the case when `__digdag_status` table does not exist. I think that case is a bit high context, so an explanation would be nice.

serihiro (approved):
I left a minor comment but this change looks good to me.

digdag-standards/src/test/java/io/digdag/standards/operator/pg/PgConnectionTest.java
komamitsu: Good point! And I'll add another case that status table doesn't exist in this test class as well as PgIT.

komamitsu:
Thanks!

PR #1564, opened 2021-04-22
Collect tasks profile per site as well

komamitsu:
We've added `digdag profile` cli command https://github.com/treasure-data/digdag/pull/1533. But it collects and outputs only the whole summary of all the sites' tasks not grouping per site. When using Digdag with multiple sites, profiled metrics per site would be very useful information to see the performance and stability. This PR introduces the feature.


digdag-cli/src/main/java/io/digdag/cli/profile/TasksStats.java
```diff
@@ -0,0 +1,104 @@
+package io.digdag.cli.profile;
```
komamitsu, line 1: Moved from TaskSummary

digdag-cli/src/main/java/io/digdag/cli/profile/TasksSummary.java
```diff
@@ -101,35 +66,23 @@ public TasksSummary(
         this.execDurationMillis = execDurationMillis;
     }
 
-    static class Builder
+    interface Builder
```
komamitsu, line 69: This is abstracted for PropagatableTasksSummaryBuilder

yoyama (approved):
LGTM.

komamitsu:
Thanks!

PR #1566, opened 2021-04-30
Filter out dynamically generated`^sla` task in `digdag profile` command

komamitsu:
`digdag profile` command filters out `^error`, `^check` and `^failure-alert` tasks that are generated dynamically since their tasks' `started_at` don't show actual start delay. But I noticed `^sla` task also needs to be filtered out.


szyn (approved):
👍

PR #1572, opened 2021-05-13
Add note about the polling cycle

akito19:
[`s3_wait>` article](https://docs.digdag.io/operators/s3_wait.html) doesn't mention the polling cycle yet.
So, I'd like to mention it in our doc.

https://github.com/treasure-data/digdag/blob/42ea7cb71270957ea6424c2df88dd7c65d248789/digdag-standards/src/main/java/io/digdag/standards/operator/aws/S3WaitOperatorFactory.java#L47


szyn (approved):
Looks good to me 👍

PR #1575, opened 2021-05-18
Change publish destination from Bintray to OSS Sonatype

yoyama:
Because of sunset of Bintray, we move to OSS Sonatype.
As the internal discussion, we stop support of `releaseSnapshot` tasks due to snapshot repository behavior is different between Bintray and sonatype and fix to support will affect to our internal deployment.
In addition, we have never used `releaseSnapshot` for Digdag OSS release so far.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,6 +28,11 @@ allprojects {
     apply plugin: 'java'
     apply plugin: 'jacoco'
     apply plugin: 'net.ltgt.apt-idea'
+    // 'maven' plugin is deprecated and replaced by 'maven-publish'.
+    // But it is required for task 'pom' which generates pom.xml in root and subproject root.
+    // ci/validate.sh call the task and run some mvn commands like 'enforcer'.
+    // But there is no enforcer rule definition. So it might be better to remove the task and the plugin.
+    apply plugin: 'maven'
 
     repositories {
         mavenCentral()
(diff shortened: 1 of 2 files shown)


komamitsu:
When I tried `./gradlew publishToMavenLocal -PwithoutUi`, it failed with this error. We'd better address this error?

```
> Task :digdag-cli:signMavenJavaPublication FAILED

FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':digdag-cli:signMavenJavaPublication' (type 'Sign').
> No value has been specified for property 'signatory.keyId'.
```

build.gradle
```diff
@@ -298,20 +255,38 @@ subprojects {
...
+                        url "https://oss.sonatype.org/service/local/staging/deploy/maven2"
+                    }
+                    credentials {
+                        username = project.hasProperty("ossrhUsername") ? ossrhUsername : ""
+                        password = project.hasProperty("ossrhPassword") ? ossrhPassword : ""
+                    }
+                }
+            }
+        }
+        signing {
```
komamitsu, line 287: If you want `gradle publishToMavenLocal` to work without `signing.gnupg` properties, this would be a workaround
```
    if (project.hasProperty('signing.gnupg.keyName')) {
        signing {
            :     
```

yoyama: Good point! We need to publish in local and it is useful fix.

yoyama: I fixed it.

komamitsu (approved):
👍

PR #1580, opened 2021-05-24
Add pom info

yoyama:
Add some information in pom file to publish packages to OSS sonatype.
Note: As developer I listed original developer and current commiters.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -266,6 +266,48 @@ subprojects {
                         artifact javadocJar
                         pom.withXml(pomXmlModifier)
                     }
+                    pom {  // https://central.sonatype.org/pages/requirements.html
+                        name = "${project.name}"
+                        afterEvaluate { project ->
+                            description = "Digdag"
+                        }
+                        url = "https://www.digdag.io/"
+                        licenses {
+                            license {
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -266,6 +266,47 @@ subprojects {
                         artifact javadocJar
                         pom.withXml(pomXmlModifier)
                     }
+                    pom {  // https://central.sonatype.org/pages/requirements.html
+                        name = "${project.name}"
+                        afterEvaluate { project ->
+                            description = "Digdag"
```
komamitsu, line 272: Does this need to be in `afterEvaluate`? Simple configuration looks to work (e.g. https://github.com/komamitsu/fluency/blob/master/build.gradle#L107)

komamitsu (approved):

build.gradle
yoyama: I refer embulk's build.gradle.
In embulk, description is described in each subproject and need to evaluate later.
Of course it is not required in current description "Digdag", anyway no impact to publish.
So I keep it for a while.
https://github.com/embulk/embulk/blob/master/build.gradle#L159-L162

PR #1585, opened 2021-05-27
Use Application Default Credential if gcp.credential is absent

rhase:
Chaged target branch of #1531 to master.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -167,6 +167,9 @@ subprojects {
             destinationFile = file("$buildDir/jacoco/${name}.exec")
             excludes = ['**/Immutable*', '*Immutable*', 'generated/**']
         }
+
+        // for BqDdlOperatorFactoryTest
+        environment "GOOGLE_APPLICATION_CREDENTIALS", "/tmp/dummy-service-account-key.json"
     }
 
     javadoc {
--- a/build.gradle
+++ b/build.gradle
@@ -167,9 +167,6 @@ subprojects {
             destinationFile = file("$buildDir/jacoco/${name}.exec")
             excludes = ['**/Immutable*', '*Immutable*', 'generated/**']
         }
-
-        // for BqDdlOperatorFactoryTest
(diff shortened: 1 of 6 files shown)


rhase:
@szyn 
I've recreated #1531 .

build.gradle
```diff
@@ -167,6 +167,9 @@ subprojects {
             destinationFile = file("$buildDir/jacoco/${name}.exec")
             excludes = ['**/Immutable*', '*Immutable*', 'generated/**']
         }
+
+        // for BqDdlOperatorFactoryTest
+        environment "GOOGLE_APPLICATION_CREDENTIALS", System.getProperty("java.io.tmpdir") + "/dummy-service-account-key.json"
```
szyn, line 172: Is there any reason why you are passing this via an environment variable? If not, how about putting it in GcpCredentialProviderTest.java?

szyn, line 172: Is there any reason why you are passing this via an environment variable? If not, how about putting this variable in GcpCredentialProviderTest.java?

rhase: I have to use environment variable to cofigure the path of the credential in [ADC](https://cloud.google.com/docs/authentication/production#automatically) and I don't know how to overwrite environment variable of the java process from the java code itself.
Do you have any ideas how to do that?

szyn: Thank you for sharing it, I understand that case. I noticed that this's surely the simplest way to do it, so there's no problem. As you describe in the comment it's only for `BqDdlOperatorFactoryTest, it meas all subprojects don't have to set this environment variable. How about moving this variable to `build.gradle` under `digdag-standard` project?

rhase: Moved under digdag-standard !

szyn (approved):
👍 Thank you for your contribution!

szyn:
Note that we'd better have integration tests for this as well. We will work on it as a follow-up.

PR #1590, opened 2021-05-28
Detect stuck non-blocking operator tasks

komamitsu:
I observed non-blocking operator `td` was stuck in the middle of config evaluation for hours according to JVM thread dump. It looked the first time of this kind of issue, but we'd like do see how often the issue happens. This pull request introduces a mechanism to detect stuck non-blocking operators and logs it as the first step.

This feature is disabled on these operators:
- `http`
- `mail`
- `notify`
- `py` / `rb` / `sh` with DockerCommandExecutor or SimpleCommandExecutor
- `wait` in blocking mode
- `pg` / `redshift`

TODO:
- [x] The issue I experienced as written above happened at https://github.com/treasure-data/digdag/blob/afb261043d8778f5c77f8d05736e73578b6a47c8/digdag-core/src/main/java/io/digdag/core/agent/OperatorManager.java#L239 before `operator.run()`. It means current implementation of this PR can't detect the issue. So this PR also needs to have a similar mechanism for the call of `io.digdag.core.agent.OperatorManager#evalConfig`

- [x] In current implementation of this PR, the default impl of Operator#isBlocking returns `true` and the above non-blocking operators' `isBlocking` are overridden to return `false`. But on second thought, we'd better set the default value to `false` and override some blocking operators (e.g. pg, redshift, mail, http) `isBlocking()` to return `true`

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/agent/AgentConfig.java
+++ b/digdag-core/src/main/java/io/digdag/core/agent/AgentConfig.java
@@ -10,13 +10,16 @@ import org.immutables.value.Value;
 @JsonDeserialize(as = ImmutableAgentConfig.class)
 public interface AgentConfig
 {
-    static final int DEFAULT_HEARTBEAT_INTERVAL = 60;
-    static final int DEFAULT_LOCK_RETENTION_TIME = 300;
-    static final int DEFAULT_MAX_TASK_THREADS = 0;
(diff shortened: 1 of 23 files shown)


szyn (approved):
LGTM 👍

komamitsu:
Thanks!

PR #1598, opened 2021-06-06
Add digdagignore

seiyab:
Issue: #1092

Reading prior work #1175 , I
- wrote change in `ProjectArchive`.
- use eclipse JGit for implementation.

I still have some worries.
- I don't understand "In your current implementation, `ignoreFileList` method is called every tar entry copying. " in https://github.com/treasure-data/digdag/pull/1175#issuecomment-511006334 yet.
- whether `listFiles` method is good place to parse ignorefile or not.
- where should I write doc about `.digdagignore`

### DIFF
--- a/digdag-core/src/main/java/io/digdag/core/archive/ProjectArchive.java
+++ b/digdag-core/src/main/java/io/digdag/core/archive/ProjectArchive.java
@@ -5,6 +5,7 @@ import java.io.IOException;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.util.HashSet;
+import java.util.Optional;
 import java.util.Set;
 import java.nio.file.Path;
 import static java.util.Locale.ENGLISH;
@@ -59,9 +60,9 @@ public class ProjectArchive
     static void listFiles(Path projectPath, PathConsumer consumer)
         throws IOException
     {
+        Optional<DigdagIgnore> digdagIgnore = DigdagIgnore.ofProject(projectPath);
         DirectoryStream.Filter<Path> filter = (target) -> (
-                rejectDotFiles(target) &&
-                DigdagIgnore.ofProject(projectPath)
+                rejectDotFiles(target) && digdagIgnore
                         .map((ignore) -> ignore.accept(target))
                         .orElse(true)
         );
--- a/digdag-core/src/main/java/io/digdag/core/archive/ProjectArchive.java
+++ b/digdag-core/src/main/java/io/digdag/core/archive/ProjectArchive.java
@@ -60,7 +60,9 @@ public class ProjectArchive
     static void listFiles(Path projectPath, PathConsumer consumer)
         throws IOException
     {
+        // parse .digdagignore
         Optional<DigdagIgnore> digdagIgnore = DigdagIgnore.ofProject(projectPath);
+        // filter rejects files / directories match .digdagignore and files / directories start with dot (.)
         DirectoryStream.Filter<Path> filter = (target) -> (
                 rejectDotFiles(target) && digdagIgnore
                         .map((ignore) -> ignore.accept(target))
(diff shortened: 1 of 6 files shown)


seiyab:
IgnoreNode in JGit doesn't seem to work with `Path::toString` on Windows.
I'll search for a solution.

seiyab:
I resolved the problem 479493f .
Though it might be good solution, it should work with almost all cases.
- Windows
    - According [this post](https://softwareengineering.stackexchange.com/a/250800), it is safe.
        - I'm not sure if it is really true.
- Not Windows
    - `replace` method replaces `/` by `/`. Nothing happens.

digdag-core/src/main/java/io/digdag/core/archive/ProjectArchive.java
```diff
@@ -63,7 +60,13 @@ public void listFiles(PathConsumer consumer)
     static void listFiles(Path projectPath, PathConsumer consumer)
         throws IOException
     {
-        listFilesRecursively(projectPath, projectPath, consumer, new HashSet<>());
+        Optional<DigdagIgnore> digdagIgnore = DigdagIgnore.ofProject(projectPath);
```
szyn, line 63: How about moving these lines (L63-L68) to `listFilesRecursively`? I feel there is no need to add a new argument `filter` to `listFilesRecursively`. Also, it would better add comments for parsing `.digdagignore`, and files/directories that start with a dot will ignore.

digdag-core/src/test/java/io/digdag/core/archive/ProjectArchiveTest.java
```diff
@@ -128,6 +119,46 @@ public void listFilesExcludeDotFiles()
         assertThat(files, is(expected.build()));
     }
 
+    @Test
```
szyn, line 122: 👍

szyn (commented):
Thank you for creating this PR!

> I don't understand "In your current implementation, ignoreFileList method is called every tar entry copying. " in #1175 (comment) yet.

Honestly, I don't fully understand it either. However, I assume it is inefficient because the implementation you mentioned creates a list of files to ignore and then loops through the files included in the project to evaluate whether the target file is included each time.

> whether listFiles method is good place to parse ignorefile or not.

The current implementation evaluates the files/directories to ignore in `listFilesRecursively`, and it's good to follow that. I left a minor comment. Please see my comment.

> where should I write doc about .digdagignore

How about adding it to the end of the `push` command reference? Although this is a client mode documentation, I think this is the best place to add the `.digdagignore` description since the instructions for pushing the project are here. It would also be helpful to add a note that files/directories that start with a dot will ignore.
https://docs.digdag.io/command_reference.html?highlight=push#push
https://github.com/treasure-data/digdag/blob/master/digdag-docs/src/command_reference.rst#push

digdag-core/src/main/java/io/digdag/core/archive/ProjectArchive.java
seiyab: @szyn 
Thank you for your review.
> How about moving these lines (L63-L68) to `listFilesRecursively` ?

IMO, it is unnatural just a bit to parse `.digdagignore` and instanciate `DigdagIgnore` every recursive call of `listFilesRecursively`.
I would move it if you want.

szyn: Thanks, you're right. No need to move it.

szyn (approved):
Looks good to me, thank you for contributing 👍

FredericoCoelhoNunes:
Hey! Any estimate when this functionality will be available? In my case, I need it so I can have test .dig files that are used by my tests suite but that I don't want to push to my production server. The way my code is organized, they need to reside in the root of my directory.

minhhoang1023:
Same here! This feature would be HUGELY beneficial to me. Know when this would be released allows me to decide whether to pursue for other solutions in the mean time

szyn:
Thank you for your feedback. I'm going to prepare the next version soon. I think we can release the next one within this week, and I'd appreciate your patience.

PR #1601, opened 2021-06-07
Stop skipping IT by assumeThat

yoyama:
This PR replace `assumeThat` with `assertThat` to avoid unexpected skip of tests by misconfiguration.
Integration tests(IT) in CI should be executed regardless of setting.
Some of assumeThat codes are still kept because it is used to absorb environmental difference (e.g. Unix vs Windows, H2 vs Postgres) and reasonable.

I updated Github Workflow ci.yml to skip IT explicitly because we don't have environmental variables to run many of tests in IT.


szyn (approved):
👍

PR #1604, opened 2021-06-10
Guava Throwables deprecation warning

yoyama:
Currently there are many deprecation warning on `Throwables` of Guava.
This PR reduces warning message in build and reduce dependency to Guava.

- Upgrade Gradle to 6.8.3 as normal refactoring work.
- Upgrade JUnit to 4.13.2 to use some useful methods.
- Change Guava version to follow the result of `gradlew deps`.
- Create sub project `digdag-commons` which is lightweight libraries and all other sub project can depend on it.
- Create `ThrowablesUtil` class and provide a few methods which are compatible with original methods.
- Change all code as using `ThrowablesUtil` instead of Throwable.
- Change `assertThat()` from JUnit to Hamcrest to fix deprecation warning.
   `assertThat` of JUnit call Hamcrest's one internally. So we can expect the compatibility between them.
   https://github.com/junit-team/junit4/blob/05fe2a64f59127c02135be22f416e91260d6ede6/src/main/java/org/junit/Assert.java#L964


yoyama:
I found upgrade JUnit  to 4.13.2 causes too many deprecation warning of assertThat. I will fix it.

build.gradle
```diff
@@ -83,7 +83,7 @@ subprojects {
         jacksonVersion = "2.9.10"
         jacksonDatabindVersion = "2.9.10"
         awsJavaSdkVersion = "1.11.686"
-        guavaVersion = "19.0"
+        guavaVersion = "30.0-android"
```
komamitsu, line 86: This version is for Android project? `-jre` should be good.

digdag-cli/src/main/java/io/digdag/cli/GraphvizWorkflowVisualizer.java
```diff
@@ -13,7 +13,7 @@
 import java.io.IOException;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
-import com.google.common.base.Throwables;
+import io.digdag.commons.guava.ThrowablesUtil;
```
komamitsu, line 16: How about just `io.digdag.commons.ThrowablesUtil` since we're hiding the explicit dependency to Guava ?

yoyama: I don't have strong opinion about the package name. 
My thought is simply we may have some different functionalities under `io.digdag.commons`  in the future.
But currently only this Class exists. 
So I agree with you and change the package name to `io.digdag.commons';

build.gradle
yoyama: The reason of the version is simply following the implicit version upgrade as following.
```
|    |    |    +--- com.fasterxml.jackson.datatype:jackson-datatype-guava:2.9.10
|    |    |    |    +--- com.google.guava:guava:18.0 -> 30.0-android
```
But yes. there is no reason to choose `-android`. So I will check and try to set `-jre`.

komamitsu:
```
 |    |    |    |    +--- com.google.guava:guava:18.0 -> 30.0-android
```

How did you get the result? I checked v0_11 dependencies and got the following one. But I didn't get 30-android.
```
|    \--- com.google.inject.extensions:guice-multibindings:4.2.2
|         \--- com.google.inject:guice:4.2.2
|              +--- javax.inject:javax.inject:1
|              +--- aopalliance:aopalliance:1.0
|              \--- com.google.guava:guava:25.1-android (*)
```

BTW `com.google.inject:guice:4.2.2` explicitly specifies `25.1-android`
https://github.com/google/guice/blob/86a88f0728ca477b4e33e3c1cb042e0ec02f2feb/pom.xml#L144-L148 and I found a related issue https://github.com/google/guice/issues/1207.

Latest Guice specifies `30.1-jre`
https://github.com/google/guice/blob/master/pom.xml#L144-L148, so `-jre` seems okay.

My concern is `-jre` and `-android` have some differences and it may cause a problem in the future.

yoyama: @komamitsu I guess recent change in `master` branch may affect the guava version.
And agree with `-jre`.
 I will update to the latest `30.1.1-jre`.

https://mvnrepository.com/artifact/com.google.inject/guice/4.2.2

digdag-cli/src/main/java/io/digdag/cli/GraphvizWorkflowVisualizer.java
yoyama: Fixed.

build.gradle
yoyama: Updated the guava version.

komamitsu: I checked the dependency of latest `master` branch again, but didn't get `30.0-android`. I want to just avoid hidden unexpected dependency. My operation to check the dependency was like this
```
$ ./gradlew -p digdag-core dependencies | less
(search for "com.google.guava:guava")
  :
|    \--- com.google.inject.extensions:guice-multibindings:4.2.2
|         \--- com.google.inject:guice:4.2.2
|              +--- javax.inject:javax.inject:1
|              +--- aopalliance:aopalliance:1.0
|              \--- com.google.guava:guava:25.1-android (*)
```

Is it different from yours?

yoyama: I guess `digdag-standards` will update guava version.

komamitsu: Ah, bingo...

```
$ ./gradlew -p digdag-standards dependencies | less
  :
+--- com.google.cloud:google-cloud-core:1.94.0
|    +--- com.google.guava:guava:30.0-android (*)
|    +--- com.google.api:gax:1.60.0
|    |    +--- com.google.guava:guava:29.0-android -> 30.0-android (*)
```

Good to know that. Thanks!

komamitsu (approved):

PR #1618, opened 2021-07-26
Add an extension point to check pushed project content

komamitsu:
This PR adds an extension point to check pushed project content.

TODO: Write some unit tests


yoyama (approved):
👍

PR #1620, opened 2021-08-01
Fix polling (ECS) command executors fail to run tasks on retry

builtinnya:
This PR fixes a bug that ECS command executor fails to submit ECS tasks on retry. 

## Reproducible steps

digdag version: v0.10.1 (and some earlier versions probably)

Assuming that ECS command executor is set up correctly:

1. Add and run the following workflow

    ```
    +ecs-retry-test:
      sh>: |
        echo "Task has been executed!"
        exit 1
      _retry: 1
    ```

2. See task logs to confirm that the command executor actually prints `Task has been executed!` once.

### Expected behavior

In the above steps, the command executor should print `Task has been executed!` twice.

## Cause

Each operator runs a command (= submits an ECS task) only when `"commandStatus"` doesn't exist on state params.
However, ECS command executor's polling mechanism persists `"commandStatus"` even on retry, which tries to poll an ECS task that has already exited.

## Approach in this PR

This PR takes minimal change approach and simply removes `"commandStatus"` from state params on retry.

## Considerations

- Maybe we should also use `TaskExecutionException` for commands' failure to propagate state params.
- Currently, each operator tries to remove `"commandStatus"` from state params on failure (e.g. [`sh>` op](https://github.com/treasure-data/digdag/blob/99c38452cfe1dd62042e6e454ba553a26c48c6db/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java#L120)) but it has no effect on the state params on retry because it just throws `RuntimeException` right after that. Should we remove these confusing lines?

### DIFF
--- a/digdag-plugin-utils/src/main/java/io/digdag/util/BaseOperator.java
+++ b/digdag-plugin-utils/src/main/java/io/digdag/util/BaseOperator.java
@@ -51,9 +51,11 @@ public abstract class BaseOperator
 
             boolean doRetry = retry.evaluate();
             if (doRetry) {
+                // Remove command status so that pollable command executors can run tasks on retry
+                Config nextStateParams = retry.getNextRetryStateParams().remove("commandStatus");
                 throw TaskExecutionException.ofNextPollingWithCause(ex,
                         retry.getNextRetryInterval(),
-                        ConfigElement.copyOf(retry.getNextRetryStateParams()));
+                        ConfigElement.copyOf(nextStateParams));
             }
             else {
                 throw ex;
--- a/digdag-plugin-utils/src/test/java/io/digdag/util/BaseOperatorTest.java
+++ b/digdag-plugin-utils/src/test/java/io/digdag/util/BaseOperatorTest.java
@@ -1,6 +1,8 @@
 package io.digdag.util;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
 import io.digdag.client.config.Config;
 import io.digdag.client.config.ConfigElement;
 import io.digdag.client.config.ConfigFactory;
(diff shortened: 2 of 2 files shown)


szyn:
Thank you for creating this PR! 
First of all, we haven't been able to reproduce this issue in our environment yet. It seems that task retry works appropriately. Thus let me confirm, is this issue happening for another executor type such as Simple/DockerCommandExecutor as well?
I understand this approach is quite simple, but  I'm worried that this change for BaseOperator affects all command executors. If this problem is happening only with EcsCommandExecutor, we would better take a different approach to fix it.

builtinnya:
@szyn 
This only applies for EcsCommandExecutor. Retry works correctly for other command executors because it doesn't rely on `"commandStatus"` and can't be affected by the previous execution anyway.

>I'm worried that this change for BaseOperator affects all command executors. If this problem is happening only with EcsCommandExecutor, we would better take a different approach to fix it.

I understand your concern but we probably need to change codes other than EcsCommandExecutor and could affect all other command executors anyway because EcsCommandExecutor's polling behavior itself is achieved by coordination of BaseOperator, ShOperatorFactory, and so on.

At least the current executors can't be affected by this change as it can't rely on `"commandStatus"` presence.

https://github.com/treasure-data/digdag/blob/0c6e58afa78ba774c968efc078c43a15fa8a7857/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java#L105-L113

Operators run commands only if `"commandStatus"` doesn't exist in state params and the `else` branch is never executed (`poll()` is currently implemented only on EcsCommandExecutor.)

To summarize my points:

- The current EcsCommandExecutor's polling behavior is already relying on the implementation of BaseOperator, ShOperatorFactory, PyOperatorFactory, and so on. So changing some of these codes would be inevitable.
- Command executors other than EcsCommandExecutor aren't affected by this PR's changes.

Please let me know if some of my changes aren't clear to you and any suggestion for better changes.

myui (approved):
I could reproduce it with ECS command executor.
Since commandStatus is not used by other command executors, no drawbacks in this PR. LGTM.

While sh and py operator tried to remove commandStatus for non-zero exits, I guess it's different objects and no effect for retrying.
https://github.com/treasure-data/digdag/blob/master/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java#L120

myui:
Sho agreed to merge this PR. Thank you for contributing!

szyn:
Yes, finally, I could reproduce this issue... and this approach looks good to me 👍  Thank you for contributing!

PR #1622, opened 2021-08-03
Fix integration test skip

yoyama:
Currently acceptance.* (integration tests) in `digdag-tests` does not run either of CircleCI and GitHub action.
This PR fixes them.

- Some of integration tests ran on GitHub Action in past. But now stoped.
- Should run all integration test (digdag-tests) in CircleCI
- Some of them (S3WaitIT, S3StorageIT, DockerIT) fail. So exclude temporarily. I will fix it later.
- Suspect of Pg connection leak causes failure in some tests. As workaround, stop second container for Pg and run Pg in `digdag-build` with max connection increase
- td.PyIT is still unstable. Add wait before getting log

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,7 @@ jobs:
 
       # Set up Digdag database
       - run:
+          name: setup PostgreSqL
           command: |
             sudo ls /etc/postgresql/
             sudo sed -i 's/max_connections = 1000/max_connections = 2000/' /etc/postgresql/9.5/main/postgresql.conf
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,11 +11,6 @@ jobs:
         environment:
           TERM: dumb
           TZ: 'UTC'
-      - image: circleci/postgres:9.5-alpine
-        environment:
-          POSTGRES_USER: digdag_test
-          POSTGRES_DB: digdag_test
-
     steps:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -11,17 +11,16 @@ jobs:
         environment:
           TERM: dumb
           TZ: 'UTC'
-      - image: circleci/postgres:9.5-alpine
```
yoyama, line 14: stop using secondary container. Instead run Pg in digdag-build image.

digdag-tests/src/test/java/acceptance/td/PyIT.java
```diff
@@ -149,7 +149,7 @@ public void testRunOnEcs()
             }
             assertThat(attempt.getSuccess(), is(true));
         }
-
+        Thread.sleep(60 * 1000); // Log will be delayed
```
komamitsu, line 152: I think I saw the impl of EcsCommandExecutor waits for the end marker from CWL. So, missing log means a bug of `EcsCommandExecutor`?

yoyama: I am not sure the root cause, but the failure frequently happened and difficult to pass CI.
After adding the step, I don't face the issue. So log will be delayed.

komamitsu (approved):

PR #1627, opened 2021-08-10
Modify wrong sample

kumo203:
I think the sample code lines for td> operator are somehow oppositely placed.
query: option is for expanding string and data: option is for non-expanding string, I think .
Please check🙇

### DIFF
--- a/digdag-docs/src/operators/td.md
+++ b/digdag-docs/src/operators/td.md
@@ -11,11 +11,11 @@
 
     +simple_query_expanded:
       td>:
-        data: "SELECT '${session_id}' FROM nasdaq"
+      query: "SELECT '${session_id}' FROM nasdaq"
 
     +simple_query_nonexpanded:
       td>:
-      query: "SELECT * FROM nasdaq"
+        data: "SELECT * FROM nasdaq"
 
     +create_new_table_using_result_of_select:
       td>: queries/select_sql.sql


hiroyuki-sato:
Hello, @kumo203. Thank you for creating this PR.

Have you ever compared both configuration results?
It seems that `query` is just an alias to `data:`.
It seems that both options work the same.

I'm not a Treasure Data user. so, I can't use `td>` operator.

https://github.com/treasure-data/digdag/blob/master/digdag-standards/src/main/java/io/digdag/standards/operator/td/TdOperatorFactory.java#L118

szyn:
I guess you'd say the order of task names (`+simple_query_expanded` and `+simple_query_nonexpanded`) are wrong 👀

szyn (approved):

kumo203:
Hello @hiroyuki-sato. I'm very sorry that I was overlooking your message and not able to reply for a long time.

I am a Treasure Data user and in my Treasure Data environment, the behaviors of `query:` and `data:` are different.
When I write like following, `query:` line works fine but `data:` line becomes error.

###sample .dig start###
timezone: UTC

_export:
  td:
    _temp_table: ${"temp_test_table_"+moment(session_time).format("YYYYMMDD")}
    database: test_db
    
+check1:
  echo>: ${td._temp_table}

+creation:
  td_ddl>:
  create_tables: ["${td._temp_table}"]

+query:
  +task1:
    td>:
    query: "insert into ${td._temp_table} select 123 as test1"
  +task2:
    td>:
      data: "insert into ${td._temp_table} select 345 as test2"
###sample .dig end###

The error message is like following.
>Query 20210822_171625_44729_s5r5d failed: line 7:13: mismatched input '$'. Expecting: <identifier>

I didn't check in Digdag, so I didn't know that the `td>` behavior difference between Digdag and Treasure Data.
I'm sorry about it.


Hello @szyn. Yes, I meant so or in Japanese, サンプルコードの`query:`と`data:`の例が反対でした。
Sorry for not able to write clear English...
And thank you for merging my PR.

hiroyuki-sato:
Hello, @kumo203. Thank you for your reply!

PR #1629, opened 2021-08-18
Enlarged retry timeout to fetch log events for non-zero status

myui:
Cloudwatch logs eventually became available.

When stdout/stderr output is too large, 60 wait is not enough to wait EoL marker.
So, enlarged retry timeout to fetch log events for non-zero status from 60 sec to 300 sec.


digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -361,6 +364,12 @@ CommandStatus createNextCommandStatus(
             }
             while (Instant.now().getEpochSecond() < timeout);
 
+            if (previousExecutorStatus.get("logging_finished_at") == null) {// case for timeout
+                logger.warn(s("Cannot find logging_finished_at marker in CloudWatch logs.\n"
```
komamitsu, line 368: IIRC, `warn` level is visible to customer. My concern is some customer might be worried too much when seeing this warning. What do you think?

komamitsu (approved):

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
myui: don't have strong opinion. changed to debug level.

PR #1630, opened 2021-09-01
Improved error handling of ECS execution when container task failed to start

myui:
Currently, digdag is trying to fetch Cloudwatch logstream even if container task failed to start (then, no log stream could be happen).

So, improved error handling of ECS execution when container task failed to start.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
@@ -18,6 +18,7 @@ import com.amazonaws.services.ecs.model.Task;
 import com.amazonaws.services.ecs.model.TaskDefinition;
 import com.amazonaws.services.ecs.model.TaskOverride;
 import com.amazonaws.services.ecs.model.TaskSetNotFoundException;
+import com.amazonaws.services.ecs.model.TaskStopCode;
 import com.amazonaws.services.logs.model.GetLogEventsResult;
 import com.amazonaws.services.logs.model.OutputLogEvent;
 import com.fasterxml.jackson.databind.node.JsonNodeFactory;
@@ -401,15 +402,14 @@ public class EcsCommandExecutor
         final Task task;
         try {
             task = client.getTask(cluster, taskArn);
+            if (logger.isDebugEnabled()) {
+                logger.debug("Get task: " + task);
+            }
         }
         catch (TaskSetNotFoundException e) {
             // if task is not present, an operator will throw TaskExecutionException to retry polling the status.
             logger.info(s("Cannot get the Ecs task status where cluster=%s, taskArn=%s. Will be retried.", cluster, taskArn));
             return EcsCommandStatus.of(false, previousStatus.deepCopy());
-        }
-
-        if (logger.isDebugEnabled()) {
-            logger.debug("Get task: " + task);
         }
 
         final EcsTaskStatus taskStatus = EcsTaskStatus.of(task.getLastStatus());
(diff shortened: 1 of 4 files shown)


digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -489,6 +504,10 @@ ObjectNode fetchLogEvents(final EcsClient client,
         final Optional<String> previousToken = !previousExecutorStatus.has("next_token") ?
                 Optional.absent() : Optional.of(previousExecutorStatus.get("next_token").asText());
         final GetLogEventsResult result = client.getLog(toLogGroupName(previousStatus), toLogStreamName(previousStatus), previousToken);
+        if (result == null) {
+            return previousExecutorStatus.deepCopy();
```
myui, line 508: `fetchLogEvents()` changed to not throw exception when LogStream is not found.

If container tasks exit non-zero, LogStream may not be created. Then, fall through to https://github.com/treasure-data/digdag/blob/v0_11/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java#L394 without showing application stdout/stderr (no entry in Cloudwatch logs).

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -439,7 +439,22 @@ CommandStatus createNextCommandStatus(
...
+            nextStatus.put("status_code", containerExitCode);
+
+            String stopCode = task.getStopCode();
+            nextStatus.put("ecs_stop_code", stopCode);
+
+            // see com.amazonaws.services.ecs.model.TaskStopCode
+            if(TaskStopCode.TaskFailedToStart.toString().equals(stopCode)) {
+                String stoppedReason = task.getStoppedReason();
+                nextStatus.put("ecs_stopped_reason", stoppedReason);
+                throw new RuntimeException(s(
```
myui, line 452: If container failed to launch, no LogStream is created. So, avoid retrying to fetch Cloudwatch logs by throwing fail-fast error.

digdag-standards/src/main/java/io/digdag/standards/command/ecs/DefaultEcsClient.java
```diff
@@ -333,31 +348,44 @@ public void close()
...
                     waitWithRandomJitter(baseWaitSecs, maxJitterSecs);
                 }
                 else {
                     throw ex;
                 }
             }
         }
-        logger.error("Failed to call EcsClient method after Retried {} times", maxRetry);
-        throw new RuntimeException("Failed to call EcsClient method");
+        String errMsg = String.format(Locale.ENGLISH, "Failed to fetch Cloudwatch log stream while system retried %d times. %s",
```
myui, line 385: Error message should be more descriptive.

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
```diff
@@ -439,7 +439,22 @@ CommandStatus createNextCommandStatus(
             // finish this poll once and wait finish marker in head of this method in next poll, considering risk of crushing in this poll.
             nextStatus.put("task_finished_at", Instant.now().getEpochSecond());
             // Set exit code of container finished to nextStatus
-            nextStatus.put("status_code", task.getContainers().get(0).getExitCode());
+            int containerExitCode = task.getContainers().get(0).getExitCode().intValue();
```
yoyama, line 442: Just in case, can we expect `getExitCode()` always return non-null value ?

yoyama:
IIUC, this PR will not fix the task failure caused by AWS infrastructure but inform the cause of error. Is this right ?

digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
myui: I assume so because at least current code assumes non-null value in 
https://github.com/treasure-data/digdag/blob/v0_11/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java#L347

myui:
> IIUC, this PR will not fix the task failure caused by AWS infrastructure but inform the cause of error. Is this right ?

True.

This PR introduces 

1.  better error message
2. avoid long retrying when ECS failed to launch and throw an exception with detailed error message
3. if container python task exit non-zero, WF task itself can be failed even if LogStream is missing. Current LogStream missing exception becomes python task exit non-zero failure.

yoyama:
Make sense to me. Thank you for your explanation.

yoyama (approved):

PR #1643, opened 2021-09-16
Add a hook for processing AuthenticatedUser

myui:
We would like to store WF invoked user information in TD.  So, add hook.

Example usage:

```java
    @Overrided
    protected void putAdditionalParams(@Nonnull Config params)
    {
        AuthenticatedUser user = getAuthenticatedUser();
        final Config info = user.getUserInfo().getNested("td").getNested("user");
        if (info.has("id")) {
            params.set("user_id", info.get("id", Integer.class));
        }
        if (info.has("email")) {
            params.set("user_email", info.get("email", String.class));
        }

        final Config context = user.getUserContext().getNested("td").getNested("user");
        if (context.has("ip_address")) {
            params.set("ip_address", context.get("ip_address", String.class));
        } else if (context.has("ip")) {
            params.set("ip_address", context.get("ip", String.class));
        }
    }
```

### DIFF
--- a/digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
+++ b/digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
@@ -261,7 +261,7 @@ public class AttemptResource
                     .or(ImmutableList.of());
 
             Config params = request.getParams();
-            processAuthenticatedUserInfo(authedUser, params);
+            putAdditionalParams(params);
 
             // use the HTTP request time as the runTime
             AttemptRequest ar = attemptBuilder.buildFromStoredWorkflow(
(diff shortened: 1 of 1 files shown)


digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
```diff
@@ -256,10 +258,13 @@ public Response startAttempt(RestSessionAttemptRequest request)
                     .transform(r -> collectResumingTasks(r))
                     .or(ImmutableList.of());
 
+            Config params = request.getParams();
+            putAdditionalParams(params);
```
yoyama, line 262: IIUC `request.params` store user defined parameters.
So mix of system and user defined may be harmful.

yoyama: Ah scheduler may write some parameters to it. Let me check it.

yoyama: Ok, I confirmed both user define and system defined params are stored.

yoyama (approved):

digdag-server/src/main/java/io/digdag/server/rs/AttemptResource.java
myui: so, then `params.set("__td__.user_id", info.get("id", Integer.class));` or some.

PR #1649, opened 2021-09-29
[digdag-ui] Proposal: folding task timeline on click

seiyab:
# What is this PR?
See the following screen shot.
Sep-29-2021 22-59-37

# Why?
It was difficult to read timelines if workflows are complicated.

# Other comments
- I added test framework without discussion. I would drop it if I shouldn't add it.
- I'm not sure if it is useful for other users. I would close this PR if it just increase complexity and isn't useful.
- I missed to make CI to run the tests. I would fix it.


szyn:
Thank you for creating this PR. I believe this is a helpful feature because the timeline becomes complicated when there are many tasks. 

Adding tests is also very welcome, but could you please make the test implementation become a separate PR from a review perspective? 
As you said, you also need to fix the CI. I believe it would be better to review that PR first and merge it into the master.

seiyab:
@szyn 
Thank you for your comment 👍 🏄 
I separated PR. #1653

seiyab:
@szyn 
CI on master branch fails since 6cde583572a25760697 .
Is this PR related? If so, I would help.

szyn:
@seiyab Thank you for following up. It seems this PR is not related to the failure 👍 
The tests for that commits were failed due to flaky tests in the integration testing, not the UI testing. (acceptance.td.PyIT > testRunOnEcs FAILED)
After that, the tests started to fail due to the changes in the CLI. Please see also https://github.com/treasure-data/digdag/pull/1679
No worries, but let me ping you if I need your help.

Issue #1663, opened 2021-10-19
SELinux with docker needs :Z volume mount option


ehaupt:
When using SELinux with docker, volumes should be mounted with the `:Z` [mount option](https://docs.docker.com/storage/bind-mounts/#configure-the-selinux-label).

This is set here:

https://github.com/treasure-data/digdag/blob/07c11785b6f5c8c2469a94ec764e1d1a448be667/digdag-standards/src/main/java/io/digdag/standards/command/DockerCommandExecutor.java#L115-L116

When using SELinux, this should be:

```java
            command.add("-v").add(String.format(ENGLISH,
                        "%s:%s:rw,Z", projectPath, projectPath));  // use projectPath to keep pb.directory() valid
```

Maybe a config option could be introduced:

```yaml
_export:
  docker:
    image: "python:3.8-slim"
    pull_always: true
    selinux: true
```

PR #1676, opened 2021-12-06
Set selinux label when mounting docker volumes

ehaupt:
When using SELinux, docker volumes need to be mounted with :Z in order to work properly.

This implements a `selinux` boolean config option in the docker context to enable or disable SELinux support.

Also document new configuration option.

Issue: https://github.com/treasure-data/digdag/issues/1663


szyn (approved):
👍  Thank you for your contribution!

PR #1680, opened 2021-12-20
Improve logging.

yoyama:
This is back port from Treasure Workflow.
Improve logging.
- Change some repeated output while running tasks from `INFO` to `DEBUG` level.
- Add fields to SPI `TaskRequest`
- Add `__last_job_id` parameter to state_params in `td_wait>` and `td_wait_table>` operator


szyn (approved):
👍

PR #1681, opened 2021-12-20
Make CommandExecutor pluggable

### LINKED ISSUE #901: [feature-request] Use Custom CommandExecutors

szyn:
Follow-up #1315 and make CommandExecutor pluggable using the same approach as #1325. Close #901
We're going to add the system property `agent.command_executor.type`, which allows that users can specify command executor type. 

See also: #1315 (comment)


digdag-tests/src/test/resources/META-INF/services/io.digdag.spi.Plugin
```diff
@@ -1 +1,2 @@
 acceptance.TestingAuthenticatorPlugin
+acceptance.TestingCommandExecutorPlugin
```
yoyama, line 2: Adding new line will be better.

szyn: Thank you, I added a new line. 8042e55

yoyama (approved):
👍

szyn:
Thank you!

PR #1682, opened 2021-12-22
Logging config

yoyama:
This PR adds cli option `--logback-config` to specify customized logback xml file.
Digdag set a logback file from predefined logback files (`logback-color.xml`, `logback-console.xml`, `logback-file.xml`) in jar .
So it is difficult to customize logging by ourselves. This PR support it and will be helpful for especially developers.
In addition, exclude noisy debug message of HikariCP as netty.


szyn (approved):
`--logback-config` option is very useful. Thank you! 👍

PR #1683, opened 2021-12-22
Update Guice to 5, JDBC for PostgreSQL

yoyama:
This PR update Google Guice from 4 to 5. And JDBC library for Pg to the latest.
This will reduce warning "WARNING: An illegal reflective access operation has occurred" with Java11.


szyn (approved):
👍

PR #1688, opened 2022-01-12
[WIP] Upgrade Jdbi2 to 3

yoyama:
The last release of Jdbi2 is on 2017. It is 5 years ago.
We should migrate to Jdbi3.
This PR try to migrate to Jdbi3 experimentally.
Unfortunately Jdbi3 is not fully compatible with Jdbi2 and need to fix many codes.
In addition Jdbi3 uses caffeine ver 3.x but it does not support Java8.
So force caffeine version to 2.x.


szyn (commented):
Could you resolve conflicts?

PR #1696, opened 2022-01-19
Fixed to remove k8s pod on finished

hnarimiya:
When I executed KubernetesCommandExecutor, there was a case where the pod remained.
So I modified it to explicitly delete the pod after the archive file acquisition was complete.


yoyama (approved):
LGTM 👍

PR #1708, opened 2022-02-14
Fix the s3_wait empty check example

akito19:
`s3_wait` operator shows an example for `continue_on_timeout`. However, it doesn't work when `${s3.last_object}` is found.

```
+task1:
  s3_wait>: bucket/object
  timeout: 60s
  continue_on_timeout: true
+task2:
  if>: ${s3.last_object}
  _do:
    echo>: "No timeout"
 ```
 
 So, update the example in order to pass the `task2` even if the operator isn't timed out.


yoyama (approved):

szyn (approved):
Thank you!

Issue #1709, opened 2022-02-17
Got errors when I try to show logs on the web client


mattp0123:
I got a bunch of network errors when I click on the "REFRESH LOGS" in the Digdag web client.
These errors only occur on task using the "for each loop" and I'm not able to see the logs.
Please help.

![Screen Shot 2022-02-17 at 10 05 57 AM](https://user-images.githubusercontent.com/68014351/154391280-0b8e8449-859e-498c-b0ff-3b453b6fc3e0.png)

mattp0123:
@hiroyuki-sato Thanks for your reply!
### What Digdag version are you using?
I'm using 0.10.4.

### Have you ever tried `digdag log` command?
`digdag log <attemp_id>` works correctly.

### Could you provide reproduced example?
I'm using the following .dig file:
```digdag=
+task:
  for_each>:
    item: [hello, world, foo, bar]
  _parallel: true
  _do:
    sh>: ts-node /path/to/my/task.ts ${item}
```
And my task.ts file:
```javascript=
const [,, item] = process.argv
console.log(item)
process.exit(1)
```

### Could you provide the error log (a stacktrace for example.)
Where can I find Digdag web server logs?

### Does **network errors** mean corrupted the error log file?
No. Log files (.log.gz) are not corrupt.

hiroyuki-sato:
You need to specify a task log path. Launch digdag server with `-O dir` option.

mattp0123:
@hiroyuki-sato

I did specify the task log path when I launch the Digdag server (I'm using supervisord).
There is only one line of log showed up when I click the "REFRESH LOGS" button. (The one http request got 200 OK response instead of 400 Bad Request)

<img width="1314" alt="Screen Shot 2022-02-18 at 11 02 09 AM" src="https://user-images.githubusercontent.com/68014351/154609587-9813a2ee-ef48-4bb1-83f0-e36b0f714700.png">

hiroyuki-sato:
It seems that digdag works correctly in my environment.

```
digdag server -m -O /tmp/log
```

setup digdag client.

```
mkdir foo
cd foo
```

create `bar.dig`

```yaml
timezone: UTC

+task:
  for_each>:
    item: [hello, world, foo, bar]
  _parallel: true
  _do:
    sh>: echo "${item}"
```

```
digdag push foo
digdag start foo bar --session now
```


* digdag server 0.10.4
* OS: macOS 12.2.1
* Java8
* Safari: 15.3

[...]

PR #1711, opened 2022-02-22
[Doc] Add missing server options.

hiroyuki-sato:
This PR adds missing server options. I don't add `-H` and `-X` options.

Reported by [mattp0123](https://github.com/treasure-data/digdag/issues/1709#issuecomment-1044832740)


yoyama:
IIUC, admin function is deprecated. So it might be better to add (deprecated).

PR #1712, opened 2022-02-24
Support start_date and end_date for schedule

yui-knk:
I sometimes want to run workflows only limited period.
This PR includes:

* Add `start_date` and `end_date` columns into `schedules` table
* Both columns are nullable, so
  * When both of them are null, it behaves as is
  * When start_date is set, it runs once start_date comes
  * When end_date is set, it runs until end_date comes
  * When both of them are set, it runs only when `start_date < now() < end_date`
* `start_date` and `end_date` can be specified only on dig file like below, acceptable datetime format is arguable, but I think it's better to not allow to set timezone (timezone can be configurable only `timezone` confg)

```
timezone: Asia/Tokyo

schedule:
  minutes_interval>: 1
  start_date: 2022-02-23T14:01:00
  end_date: 2022-02-23T14:03:00

_export:

+say_hello:
  echo>: Hello world!
```


yoyama:
We are planning to implement similar functions in TD and back port to Digdag.
This PR will conflict to it. So I would like to decline this PR. 
Cc @szyn

yoyama:
#1750 has been merged. So close.

PR #1716, opened 2022-03-02
Implement new Quartz cron scheduler

ehaupt:
**Background**:

Digdag provides different [scheduling options](http://docs.digdag.io/scheduling_workflow.html).

For our projects we need more fine grained options to define schedules. For instance a schedule should run at 10:15am on the last Friday of every month.

**Solution**:

Implement a new [Quartz](http://www.quartz-scheduler.org/) cron scheduler provided by [cron-utils](https://github.com/jmrozanec/cron-utils).

This makes it possible to use more advanced scheduling. [Here](http://www.quartz-scheduler.org/documentation/quartz-2.3.0/tutorials/crontrigger.html) is a nice overview of possible options (see Examples).

**Implementation**:

This patch implements a new `quartz_cron` scheduler.

Usage:

```yaml
timezone: "Europe/Zurich"

schedule:
  quartz_cron>: 0 15 10 ? * 6L

+PrintMessage:
  echo>: I run at 10:15am on the last Friday of every month
```

Co-authored-by: @holec


yoyama (changes requested):
@ehaupt Thank your your contribution.

I have a few comments on this PR.

- We are planning to replace very old cron4j to others. And cron-utils is one of the candidate and good choice 👍 
- We don't want to keep both cron4j and cron-utils.
- If we use cron-utils, it supports some type of cron. So it might be better to support others as well
- Implement cron types as different scheduler like `quartz_cron` will be complicated.
- If we replace cron4j, committers need to check the compatibility strictly.

Compatibility check must be done by committers.
So I would like to decline the PR unfortunately. 
I filed an issue to keep these discussion. #1717.
If we prioritize the issue, your PR will be referred and helpful.

Anyway I appreciate your contribution very much.

ehaupt:
@yoyama Thank you for your swift and elaborate response.

Ideally, I would have liked to provide a solution with one drop-in replacement for `cron>:` but since the `Quartz` introduces a different syntax (more fields), I opted for `quartz_cron>:`.

Maybe the future implementation you suggested will need an additional configuration option for specifying the cron type:

```yaml
schedule:
  cron:
    cron_type: quartz
  cron>: 0 15 10 ? * 6L
```

yoyama:
@ehaupt Yes. We will need a parameter to set cron dialect

yoyama:
I will close this PR based on discussion.

PR #1723, opened 2022-03-10
Retry ECS RunTask on AGENT error which can happen as its normal operation

builtinnya:
## Problem

As described in [Amazon ECS events - Amazon Elastic Container Service](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_cwe_events.html#ecs_container_instance_events), the ECS container agent disconnects and reconnects several times per hour as a part of its normal operation.

>The Amazon ECS container agent disconnects and reconnects several times per hour as a part of its normal operation, so agent connection events should be expected. These events are not an indication that there is an issue with the container agent or your container instance.

So,  [ECS RunTask API can fail with `AGENT` reason](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/api_failures_messages.html) when it's requested in the middle of the reconnection. ECS RunTask API callers should handle this situation.

Currently, EcsCommandExecutor, which relies on ECS RunTask API, doesn't consider the `AGENT` error and simply fails with a RuntimeException `"Submitted task could not be found"`.
https://github.com/treasure-data/digdag/blob/a1d842cbc21eb2e076ea46c6a4e4ae37a7dd5d7e/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java#L135-L148
https://github.com/treasure-data/digdag/blob/a1d842cbc21eb2e076ea46c6a4e4ae37a7dd5d7e/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java#L539-L547

## What this PR does

This PR tries to handle the `AGENT` error by simply retrying a RunTask API call after a few seconds.
This approach has been confirmed to be effective to handle the random agent reconnections in our production environment.

## On reproduction

It's difficult to establish stable steps to reproduce this issue due to its randomness but it should be easy to see what happens when RunTask failed with `AGENT` error and why it should be handled.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/ecs/DefaultEcsClient.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/ecs/DefaultEcsClient.java
@@ -10,6 +10,7 @@ import com.amazonaws.services.ecs.model.DescribeTaskDefinitionRequest;
 import com.amazonaws.services.ecs.model.DescribeTaskDefinitionResult;
 import com.amazonaws.services.ecs.model.DescribeTasksRequest;
 import com.amazonaws.services.ecs.model.DescribeTasksResult;
+import com.amazonaws.services.ecs.model.Failure;
 import com.amazonaws.services.ecs.model.InvalidParameterException;
 import com.amazonaws.services.ecs.model.ListTagsForResourceRequest;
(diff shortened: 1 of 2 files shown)


szyn (approved):
Thank you for creating this PR. This sounds reasonable and looks good to me.

PR #1724, opened 2022-03-10
EcsCommandExecutor: return failure status when exit code is empty

builtinnya:
## Problem

When a Docker container can't be started for some reason, [Amazon ECS agent](https://github.com/aws/amazon-ecs-agent) emits [`DockerTimeoutError: Could not transition to started; timed out after waiting 3m0s` and the ECS task has `STOPPED` status](https://docs.aws.amazon.com/AmazonECS/latest/userguide/stopped-task-error-codes.html) and **no `exitCode`.**

Currently, EcsCommandExecutor treats the stopped containers without `exitCode` as successful (`exit code = 0`) and workflows continue.

https://github.com/treasure-data/digdag/blob/a1d842cbc21eb2e076ea46c6a4e4ae37a7dd5d7e/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java#L399

https://github.com/treasure-data/digdag/blob/a1d842cbc21eb2e076ea46c6a4e4ae37a7dd5d7e/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java#L514-L518

([`JsonNode#intValue()` returns 0 for non-number nodes](https://fasterxml.github.io/jackson-databind/javadoc/2.7/com/fasterxml/jackson/databind/JsonNode.html#intValue()) such as `null`.)

Containers that are not even started should not be treated as sucessful.

## What this PR does

This PR tries to fix the issue by treating the exit status of containers without `exitCode` as `1` (error) to make it fail.

## On reproduction

It's difficult to establish stable steps to reproduce this issue because it's hard to make ECS agents fail with `DockerTimeoutError` appropriately but it should be easy to see what happens when ECS tasks have no `exitCode`  and why it should be treated as error.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/EcsCommandExecutor.java
@@ -395,8 +395,13 @@ public class EcsCommandExecutor
             // To fetch log until all logs is written in CloudWatch,
             // finish this poll once and wait finish marker in head of this method in next poll, considering risk of crushing in this poll.
             nextStatus.put("task_finished_at", Instant.now().getEpochSecond());
-            // Set exit code of container finished to nextStatus
-            nextStatus.put("status_code", task.getContainers().get(0).getExitCode());
+            // Set exit code of container finished to nextStatus.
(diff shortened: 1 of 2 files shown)


szyn (approved):
Thank you for creating this PR, too. This also sounds reasonable and looks good to me.

PR #1725, opened 2022-03-17
Fixed to be able to overwrite the parameter in the retry command

hnarimiya:
I added a parameter option to the Retry command.

There are cases where I want to change a part of the parameter and execute it when retry the task that failed due to the parameter.


yoyama (approved):
LGTM. Thank you for your contribution.

PR #1726, opened 2022-03-26
Drop Python2

chezou:
Since support of Python 2 was dropped few years ago, it's a good time to drop Python 2 support for digdag as well.

According to this https://endoflife.date/python, the latest supported Python is 3.7, but I'm assuming to run 3.6+ since CI runs on Python 3.6. Note that the docker image which is running on CircleCI is still using Python 3.6, but I gave up to bump it since it failed to build the latest Docker image.


yoyama:
@chezou Thank you for your PR.
I agree with drop of Python2.
The failure on build of Docker image may be fixed in #1740 .
Anyway we may need to confirm the timing to merge this and release.
Cc @szyn

PR #1734, opened 2022-04-17
Add language api class init examples

KentFujii:
I added examples for class initialization with Ruby/Python language API.


yoyama:
@KentFujii Thank you for your contribution.
I think it might be better to describe what you want to explain in this example.
I think adding some comments to examples is one of the way for it.

PR #1744, opened 2022-06-01
add skip_delayed_by description

kohki1234:
I add skip_delayed_by description as it works not only backfill situation but normal scheduling situation as well


szyn (approved):
👍

PR #1746, opened 2022-06-08
Add explanation for arguments passing

magicpieh28:
I added explanation about how to run the class or function that needs arguments when using `py>`.

Currently, there is no explanation in the digdag documents, but actually we can pass arguments using export. So I thought the explanation needed.

### DIFF
--- a/digdag-docs/src/operators/py.md
+++ b/digdag-docs/src/operators/py.md
@@ -76,6 +76,40 @@ See [Python API documents](../python_api.html) for details including variable ma
     print("simple execution")
   ```
 
+  You can pass arguments to class for initialization using `_export` as the following:
+
+  ```yaml
+  # sample.dig
+  _export:
+    arg: awsome execution
+  py>: tasks.MyWorkflow.my_task
+  ```
+
+  ```python
+  # tasks.py
+  class MyWorkflow(object):
+      def __init__(self, arg: str):
+          self.arg = arg
+      
+      def my_task(self):
+          print(self.arg)
+  ```
+
+  Or, you can pass arguments to function as the following:
+
+  ```yaml
+  # simple_sample.dig
+  _export:
+    arg: simple execution
+  py>: simple_tasks.my_func
+  ```
+
+  ```python
+  # simple_tasks.py
+  def my_func(arg: str):
+    print(arg)
+  ```
+
 * **python**: PATH STRING or COMMAND ARGUMENTS LIST
 
   The python defaults to `python`. If an alternate python and options are desired, use the `python` option.


yoyama:
@magicpieh28 Thank you for your contribution.

Based on the example https://github.com/treasure-data/digdag/blob/master/examples/python_args.dig
It describe another way to pass argument.
So if we add the description about the way to pass argument in doc, it might be better to describe both way.

yoyama:
@magicpieh28 Thank you for your update.
I hope add description about the way to pass argument as task parameters.

digdag-docs/src/operators/py.md
```diff
@@ -76,6 +76,40 @@ See [Python API documents](../python_api.html) for details including variable ma
     print("simple execution")
   ```
 
+  You can pass arguments to class for initialization using `_export` as the following:
```
yoyama, line 79: Only describe the way to use `_export` is not enough.
Could you add describe the way to pass argument as task parameters.
```
+a:
  py>: tasks.python_args.required_arguments
  required1: "1"
  required2: 2
```

magicpieh28: Sorry for the late confirmation. Thank you for your review and detailed instructions. Please check my modification.
- modify example to describe what data type the passed arguments will actually have inside Python
- add description the way to pass arguments as task parameters
- add example about how to pass arguments for class and method simultaneously

digdag-docs/src/operators/py.md
```diff
@@ -76,6 +76,112 @@ See [Python API documents](../python_api.html) for details including variable ma
...
+      required1_1: awesome execution
+      required1_2: "awesome execution"
+      required2: {a: "a"}
+      required3: 1
+      required4: 1.0
+      required5: [a, 1, 1.0, "a"]
+    py>: tasks.MyWorkflow.my_task
+  ```
+
+  Also, you can do the same thing by defining arguments under the `py>:` operation as the following:
```
yoyama, line 94: Could you exchange the order of description ?
I think pass by argument of `py>` is default way. `_export` is another way.
So this block should be described at first. Then describe L79.

magicpieh28: Thank you for your review. I changed the order, please check the modification.

digdag-docs/src/operators/py.md
```diff
@@ -76,6 +76,111 @@ See [Python API documents](../python_api.html) for details including variable ma
...
+    py>: simple_tasks.my_func
+  ```
+
+  ```python
+  # simple_tasks.py
+  def my_func(required1: str, required2: dict[str, str]):
+    print(f"{required1}: {required2}")
+  ```
+
+  Finally, you can pass combination of class and mehtod arguments to Python script as the following:
```
yoyama, line 154: This behavior is a bit strange. 
IIUC, we can't use same argument name between class initializer and method argument for it.
It might be better to add a warning about it.

magicpieh28: Sorry for my poor English. Just to be clear, what I tried to explain is in digdag, indeed we can pass both(class and method) arguments at once.
So, on-premise that the above behavior is correct what I understood is I just need to add a warning like ‘But you can not pass combination with the same name.’ instead of modifying the example.
Is this correct? @yoyama

yoyama: @magicpieh28 Sorry for my comment is not clear.
Yes, your description and example are correct. But I feel a bit strange on this specification on `py>`.
Especially, if both initializer of Class and the method have same parameter name, we can't set them separately. (If I understand correctly).
So I propose add description to warn about it as you describe.
Anyway thank you for your contribution.

magicpieh28:
@yoyama 
Please review my modification. Thank you.

digdag-docs/src/operators/py.md
```diff
@@ -76,6 +76,111 @@ See [Python API documents](../python_api.html) for details including variable ma
...
+      required2: {a: "a"}
+      required3: 1
+      required4: 1.0
+      required5: [a, 1, 1.0, "a"]
+    py>: tasks.MyWorkflow.my_task
+  ```
+
+  This example assume following Python script:
+
+  ```python
```
yoyama, line 108: This example does not work by copy and paste it.
Don't we need to  import `Union` as follows ?
```
from typing import Union

class MyWorkflow(object):
...
```

yoyama:
@magicpieh28 Sorry for my late reply. I left a comment. Could you take a look it ?

Issue #1748, opened 2022-06-15
node-sass is deprecated


seiyab:
Moving onto dart-sass is recommended.
https://sass-lang.com/blog/libsass-is-deprecated

Major drawbacks of node-sass:
- No new features will be added.
- `npm install` fails on macOS. https://medium.com/macoclock/fix-broken-node-gyp-issue-for-mac-os-596e4a8bcffd

seiyab:
Oh I found node-sass might be no longer used.

seiyab:
`node-sass` is indirectly used.
`bootstrap-loader` uses `sass-loader`.

seiyab:
https://github.com/treasure-data/digdag/pull/1757 Done.

PR #1757, opened 2022-07-13
LibSass -> Dart Sass

seiyab:
Issue: #1748

- node-sass -> sass
- remove webpack-loader

I followed this doc: https://getbootstrap.com/docs/4.0/getting-started/webpack/

I checked behavior of workflows page, projects page, session page and so on with example workflows.
But I'm not sure enough whether it fully works or not yet.


yoyama (approved):
LGTM 👍 
I also build and confirm the behavior of UI and no found any issues.

PR #1759, opened 2022-07-21
Add multiple options to make a connection with ECS Cluster.

rariyama:
## What I did
I added more options to access to ECS Cluster.
There is currently only one option to make a connection with ECS Cluster which is the way to connect with AWS AccessKeyId and SecretAccessKey.
I made a change that [DefaultAWSCredentialsProviderChain](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html) is called if agent.command_executor.ecs.\<name\>.access_key_id is not defined, and then the other types of credentials such as environment variables or Instance Profile are used.
I referred to [the process of authorizing S3 Bucket](https://github.com/treasure-data/digdag/blob/master/digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3StorageFactory.java#L52-L79) and made the implementation closer to it as possible.

## What I confirmed
I confirmed the items below.
- `./gradlew check` is succeeded
- I can use access_key_id and secret_access_key if both are defined on properties.
- If access_key_id and secret_access_key are not written, the other credential is used and task is succeeded.

This is the first PR I create on this repository so I'm sorry in advance if there is any inconvenience.

### DIFF
--- a/digdag-docs/src/command_executor.md
+++ b/digdag-docs/src/command_executor.md
@@ -39,8 +39,8 @@ Each sub keys of `agent.command_executor` are as follows:
 | key                                |  description                                     |
 | :--------------------------------- | :----------------------------------------------- |
 | ecs.name                           | ECS Cluster name. The value &lt;name&gt; is used as the key of following configuration |
-| ecs.&lt;name&gt;.access_key_id     | AWS access key for ECS. The key needs permissions for ECS and CloudWatch  |
-| ecs.&lt;name&gt;.secret_access_key | AWS secret key                                   |
+| ecs.&lt;name&gt;.access_key_id     | (Optional)AWS access key for ECS. The key needs permissions for ECS and CloudWatch. If it is not specified, other credentials are used for authorization.  |
+| ecs.&lt;name&gt;.secret_access_key | (Optional)AWS secret key                                   |
 | ecs.&lt;name&gt;.launch_type       | The launch type of container. `FARGATE` or `EC2` |
 | ecs.&lt;name&gt;.region            | AWS region                                       |
 | ecs.&lt;name&gt;.subnets           | AWS subnet                                       |
@@ -55,6 +55,10 @@ Following keys are for configuration of temporal storage with AWS S3.
 | ecs.temporal_storage.s3.endpoint                      | The end point URL for S3         |
 | ecs.temporal_storage.s3.credentials.access-key-id     | AWS access key for the bucket    |
 | ecs.temporal_storage.s3.credentials.secret-access-key | AWS secret key                   |
+
+#### The ways of authorizing to ECS cluster and tasks.
+[DefaultAWSCredentialsProviderChain](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html) besides AWS access key and secret can be used as a credential for connecting with ECS on version 0.10.5 or above . As a result of that, if `ecs.<name>.access_key_id` is not specified, digdag server looks for one of the credentials described in the document.
+
 
 ### How to use from workflow
 
--- a/digdag-standards/src/main/java/io/digdag/standards/command/ecs/DefaultEcsClientFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/command/ecs/DefaultEcsClientFactory.java
@@ -2,8 +2,10 @@ package io.digdag.standards.command.ecs;
 
 import com.amazonaws.ClientConfiguration;
 import com.amazonaws.Protocol;
+import com.amazonaws.auth.AWSCredentialsProvider;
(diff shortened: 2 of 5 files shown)


hiroyuki-sato:
This PR author writes an article about this change.
https://tech.wealthnavi.com/entry/20221125/1669353895 written in Japanese.

@yoyama, Could you look at this PR when you get a chance?

I think this PR does not change current behavior by default.

yoyama:
Basically it looks good.
But I hope adding tests for https://github.com/treasure-data/digdag/pull/1759/files#diff-4094a97e13055fa1f84ffe3297bb4c703cea9506ad50ae581ae9bd80a0f3504bR31

rariyama:
Thanks for your reply.
I'm sorry if I misunderstood but your comment means that [`createClient`](https://github.com/treasure-data/digdag/pull/1759/files#diff-4094a97e13055fa1f84ffe3297bb4c703cea9506ad50ae581ae9bd80a0f3504bR18-R33) is needed to be tested? `DefaultAWSCredentialsProviderChain` is a method of AWS SDK so it should have been tested by others and it doesn't need to be tested in digdag.

szyn:
Since this PR change will change the behavior when system properties are not set, it would be a good idea to add a test to see if the expected authentication method (`AWSStaticCredentialsProvider`/`DefaultAW SCredentialsProviderChain`) is used depending on the existence of the system property.

IMO, I understand that there is basically no problem since the behavior is the same as before in cases where system property (`agent.command_executor.ecs.<name>.access_key_id`) is set. Maybe writing an integration test as a follow-up PR would be a good idea.

I have one request. I would appreciate it if you leave a note at the bottom of this page describing the new behavior, as in the PR, and please mention that it is supported in v0.10.5 and above. https://github.com/treasure-data/digdag/blob/011d848689ae3d31a8624d59bc8ece5c38e3ae5e/digdag-docs/src/command_executor.md?plain=1#L58

rariyama:
@szyn 
Thank you for your comments.

> Thank you for creating this PR. Since this PR change will change the behavior when system properties are not set, it would be a good idea to add a test to see if the expected authentication method (AWSStaticCredentialsProvider/DefaultAW SCredentialsProviderChain) is used depending on the existence of the system property.
> 
> IMO, I understand that there is basically no problem since the behavior is the same as before in cases where system property (agent.command_executor.ecs.<name>.access_key_id) is set. Maybe writing an integration test as a follow-up PR would be a good idea.

I agreed with the idea to add integration test for confirming that digdag server can connect with ECS with the expected way. If possible, If possible, I'll try to add the test. 

> I have one request. I would appreciate it if you leave a note on this page describing the new behavior, as in the PR, and please mention that it is supported in v0.10.5 and above.

I added the explanation on this commit https://github.com/treasure-data/digdag/pull/1759/commits/e47c6fa885f051578760b01d87cf3f1594a09059, so please review this if you are available.

By the way, We can choose credentials provided by [DefaultAWSCredentialsProviderChain](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html) other than AWS access key when connecting with S3 temporal storage because [similar method](https://github.com/treasure-data/digdag/blob/master/digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3StorageFactory.java#L52-L78) has already been implemented in S3StorageFactory and [TemporalProjectArchiveStorage uses the function](https://github.com/treasure-data/digdag/blob/master/digdag-standards/src/main/java/io/digdag/standards/command/ecs/TemporalProjectArchiveStorage.java#L19-L21). Thus, the description I added on this PR can be applied to S3 access key and secret, so I think it'll also be better to mention about that. How do you think about my idea?

szyn:
Sounds good, thank you.

> By the way, We can choose credentials provided by [DefaultAWSCredentialsProviderChain](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/DefaultAWSCredentialsProviderChain.html) other than AWS access key when connecting with S3 temporal storage because [similar method](https://github.com/treasure-data/digdag/blob/master/digdag-storage-s3/src/main/java/io/digdag/storage/s3/S3StorageFactory.java#L52-L78) has already been implemented in S3StorageFactory and [TemporalProjectArchiveStorage uses the function](https://github.com/treasure-data/digdag/blob/master/digdag-standards/src/main/java/io/digdag/standards/command/ecs/TemporalProjectArchiveStorage.java#L19-L21). Thus, the description I added on this PR can be applied to S3 access key and secret, so I think it'll also be better to mention about that. How do you think about my idea?

Thank you for sharing your opinion. That's a good point 👍 I agree with mentioning it. 
May I ask you to add a note to the document on this point? Of course, this can be done as a separate PR.

szyn (approved):

PR #1768, opened 2022-09-20
add td_result_export.md to index

mickeey2525:
## Purpose
To make td_result_export more visible by adding it to the index of the Treasure Data operator.

https://docs.digdag.io/operators/treasure_data.html

This is the current page index. td_result_export is not listed here, so it is hard to find the operator if they don't know the operator.

Screen Shot 2022-09-20 at 15 27 53


yoyama (approved):
LGTM. Thank you for your contribution.

PR #1782, opened 2023-01-11
Fix CircleCI failures and IT failures

yoyama:
This PR contains multiple fix

- Fix CircleCI failure due to change of credential handling
- Fix IT failures which are caused by #1778


rune-chan (approved):
LGTM

PR #1787, opened 2023-02-11
Add explanation to ecs command executor

rariyama:
This PR is related to https://github.com/treasure-data/digdag/pull/1759

- I added an additional explanation about S3 tempral storage. When digdag tries to connect with S3 temporal storage, it can use some types of credential besides `aws_access_key`. Thus, I described about that on README.
- I fixed the description of `use_environment_file` by removing an unnecessary statement.


szyn (approved):
Thank you for following up!

PR #1802, opened 2023-03-28
[Backport] Add support for project metadata

szyn:
Backport from Treasure Workflows.

Enhance API to support project metadata. 
Note that digdag does not support the addition of metadata at this moment, but we will introduce it as a foundation.


yoyama (approved):
LGTM.

Issue #1560, opened 2021-04-18
sh> operator of the Digdag v0.10.0 on Windows platform


hiroyuki-sato:
Relates to #1540

The `sh>` operator on the Digdag 0.10.0 doesn't work correctly on a Windows Platform.

[@hundyhundy](https://www.capybara-engineer.com/entry/2021/04/16/135450) investigated this issue. This article was written in Japanese 
In summary, It needs to change script name like `runner.ps1` instead of `runner.sh` when Digdag uses PowerShell.
https://github.com/treasure-data/digdag/blob/42ea7cb71270957ea6424c2df88dd7c65d248789/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java#L136

Workaround: Windows user uses 0.9.42 until fix this issue.

hiroyuki-sato:
This patch fixes this issue. Could you tell me what do you think this change?
I believe the current behavior doesn't change anything.

If this patch seems good, what branch should I use for PR? (v0.11? or master)



```diff
diff --git a/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java b/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
index e6cf6f3608..8ff5540fde 100644
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
@@ -133,14 +133,15 @@ public class ShOperatorFactory
         {
             final Path tempDir = workspace.createTempDir(String.format("digdag-sh-%d-", request.getTaskId()));
             final Path workingDirectory = workspace.getPath(); // absolute
-            final Path runnerPath = tempDir.resolve("runner.sh"); // absolute
+            final Boolean winOS = isWindowsPlatform();
+            final Path runnerPath = tempDir.resolve(winOS ? "runner.ps1" : "runner.sh"); // absolute
 
             final List<String> shell;
             if (params.has("shell")) {
                 shell = params.getListOrEmpty("shell", String.class);
             }
             else {
-                shell = ImmutableList.of("/bin/sh");
+                shell = ImmutableList.of(winOS ? "PowerShell.exe" : "/bin/sh");
             }
 
             final ImmutableList.Builder<String> cmdline = ImmutableList.builder();
@@ -223,5 +224,14 @@ public class ShOperatorFactory
[...]

PR #1812, opened 2023-07-17
Fix for windows

kono:
# sh> operator of v0.10.0 on Windows platform and compatibility

When specifying "powershell" for shell in a .dig file, the "powershell.exe" was 
specified as ["powershell.exe","-"] until digdag ver. 0.9.x. 
Since ver. 0.10 the "-" is no longer necessary. 

However, for compatibility, the specification ["powershell.exe","-"] 
should be read as ["powershell.exe"].

[[WIP] Fix sh> operator issue on Windows #1654](https://github.com/treasure-data/digdag/issues/1560) is also included.

### DIFF
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
@@ -242,10 +242,8 @@ public class ShOperatorFactory
         {
             final String os = System.getProperty("os.name").toLowerCase();
             if ( os.startsWith("windows")) {
-                System.out.println("windows!!");
                 return true;
             }
-            System.out.println("not windows");
             return false;
         }
     }
--- a/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
+++ b/digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
@@ -158,12 +158,9 @@ public class ShOperatorFactory
             final List<String> shell = getShell(params, winOS);
 
             final ImmutableList.Builder<String> cmdline = ImmutableList.builder();
-            if (params.has("shell")) {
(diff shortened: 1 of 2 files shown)


yoyama:
I am wondering  whether rewriting the task definiton of `sh>` operator is good way or not.
In addition the logic to check condition to rewrite is not solid.

digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
```diff
@@ -128,28 +128,41 @@ private void runCode(final Config state)
...
 
+        private List<String> getShell(final Config params, final Boolean winOS){
+            // Until digdag ver. 0.9, it was necessary to write ["powershell.exe","-"] to specify shell, 
+            // but since ver. 0.10, jobs are not executed unless ["powershell.exe"] is written.
+            // To resolve this incompatibility, change the shell specification ["powershell.exe","-"]
+            // to ["powershell.exe"].            
+            List<String> temp_shell;
+            if (params.has("shell")) {
+                temp_shell = params.getListOrEmpty("shell", String.class);
+                if (temp_shell.get(0).toLowerCase().contains("powershell") && temp_shell.size() == 2){
```
yoyama, line 139: `contains("powershell")` will match to strings like 'powershell-batch.bat`. Is it intentional ?

digdag-standards/src/main/java/io/digdag/standards/operator/ShOperatorFactory.java
```diff
@@ -128,28 +128,41 @@ private void runCode(final Config state)
...
+                    if (temp_shell.get(1).contains("-")){
```
yoyama, line 140: ditto. It will match like '-option` or 'abc-def.txt`.

kono:
@yoyama , thank you for your review.
Changed to check more solidly.

PR #1819, opened 2023-11-08
UI: Disable "RETRY *" buttons until the session is updated to prevent multiple executions

exoego:
Screen-Recording-2023-11-08-at-22 49 02


y-ken:
This is what we needed to avoid duplicated attempt execution 👍🏻

aamine (approved):

